PR #5, opened 2020-10-04
Fix haiku_example.py

asmith26:
I seemed to be getting:

```python
TypeError: mean_square_loss() takes 1 positional argument but 2 were given
```

So I think this fixes it.


hbq1 (approved):

hbq1:
Hey, thank you for the fix!

PR #12, opened 2020-10-18
Implement Adabelief

### LINKED ISSUE #6: Implement scaling by AdaBelief
See paper:

https://arxiv.org/abs/2010.07468

This should be a trivial modification of the existing `scale_by_adam` transform.

joaogui1:
Closes #6 

1. Is the implementation ok? (will fix the docstring together with tests)
2. How exactly is testing done? Just add it to the `@parameterized.named_parameters` arguments with default learning rate?

### DIFF
--- a/optax/_src/transform.py
+++ b/optax/_src/transform.py
@@ -397,8 +397,7 @@ def scale_by_belief(
     nu = _update_moment(prediction_error, state.nu, b2, 2)
     count_inc = _safe_int32_increment(state.count)
     mu_hat = _bias_correction(mu, b1, count_inc)
-    nu_nonzero = jax.tree_map(lambda v: v + eps, nu)
-    nu_hat = _bias_correction(nu_nonzero, b2, count_inc)
+    nu_hat = _bias_correction(nu, b2, count_inc)
     updates = jax.tree_multimap(
         lambda m, v: m / (jnp.sqrt(v + eps_root) + eps), mu_hat, nu_hat)
     return updates, ScaleByAdamState(count=count_inc, mu=mu, nu=nu)
--- a/optax/_src/transform.py
+++ b/optax/_src/transform.py
@@ -298,6 +298,10 @@ def _safe_int32_increment(count):
   one = jnp.array(1, dtype=jnp.int32)
   return jnp.where(count < max_int32_value, count + one, max_int32_value)
 
+def _bias_correction(moment, decay, count):
+  """Perform bias correction. This becomes a no-op as count goes to infinity."""
+  bias_correction = 1 - decay**count
+  return jax.tree_map(lambda t: t / bias_correction.astype(t.dtype), moment)
 
 def scale_by_adam(
     b1: float = 0.9, b2: float = 0.999,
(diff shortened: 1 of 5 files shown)


optax/_src/transform.py
```diff
@@ -364,6 +366,44 @@ def update_fn(updates, state, params=None):
...
+
+  def init_fn(params):
+    mu = jax.tree_map(jnp.zeros_like, params)  # First moment
+    s = jax.tree_map(jnp.zeros_like, params)  # Second Central moment
+    return ScaleByAdamState(count=jnp.zeros([], jnp.int32), mu=mu, nu=s)
+
+  def update_fn(updates, state, params=None):
+    del params
+    mu = _update_moment(updates, state.mu, b1, 1)
+    nu = _update_moment(updates - state.mu, state.nu, b2, 2)
```
mtthss, line 396: `updates` and `state.mu` might be arbitrary data structure,
therefore you cannot subtract them directly

Please use tree_multimap as follows:

```
prediction_error = jax.tree_multimap(lambda g, m: g-m, updates, mu)
nu = _update_moment(prediction_error, state.nu, b2, 2)
```

optax/_src/transform.py
```diff
@@ -364,6 +366,44 @@ def update_fn(updates, state, params=None):
   return GradientTransformation(init_fn, update_fn)
 
 
+def scale_by_belief(
+    b1: float = 0.9, b2: float = 0.999,
+    eps: float = 1e-8, eps_root: float = 0.0) -> GradientTransformation:
+  """Rescale updates according to the Adam algorithm.
+
+  References:
+    [Kingma et al, 2014](https://arxiv.org/abs/1412.6980)
```
mtthss, line 375: Please link the Adabelief paper instead here

mtthss:
Could you also expose the new function in the top level `__init__.py`?
(importing it there, and adding it to the public API defined by `__all__`).

mtthss:
For testing the `adabelief` optimiser could you add it to the list of optimisers in `test_parabel` and `test_rosenbrock`?

optax/_src/transform.py
```diff
@@ -393,10 +393,12 @@ def init_fn(params):
   def update_fn(updates, state, params=None):
     del params
     mu = _update_moment(updates, state.mu, b1, 1)
-    nu = _update_moment(updates - state.mu, state.nu, b2, 2)
+    prediction_error = jax.tree_multimap(lambda g, m: g-m, updates, state.mu)
+    nu = _update_moment(prediction_error, state.nu, b2, 2)
     count_inc = _safe_int32_increment(state.count)
     mu_hat = _bias_correction(mu, b1, count_inc)
-    nu_hat = _bias_correction(nu + eps, b2, count_inc)
+    nu_nonzero = jax.tree_map(lambda v: v + eps, nu)
```
mtthss, line 400: Why are you adding an epsilon here as well?
The eps is already add in line 403.

Adam typically only adds it *after* the bias correction.
Why does adabelief need an additional epsilon term *before* the bias correction?

mtthss:
Thanks! Great work!
One (final) comment

optax/_src/transform.py
joaogui1: There's this formula in the appendix:
image
But I believe that's just the `eps_root`, I hadn't noticed it existed before, my bad!

joaogui1:
Thank you very much for the guidance and opportunity :)

mtthss:
Amazing work, thanks for your contribution!

PR #13, opened 2020-10-18
gradient centralization

### LINKED ISSUE #7: Implement gradient centralisation
See paper

https://arxiv.org/abs/2004.01461

This transform is fairly easy to implement.

joaogui1:
Closes #7 
1. Is the library in British English? Should I change it to centralisation?
2. Is the implementation ok? I didn't see an easy way to use `tree_map`
3. Any tips on adding tests?


optax/_src/transform.py
```diff
@@ -605,3 +605,28 @@ def update_fn(updates, state, params=None):
...
+  Returns:
+    An (init_fn, update_fn) tuple.
+  """
+
+  def init_fn(_):
+    return CentralState()
+
+  def update_fn(updates, state, params=None):
+    del params
+    updates = [up - up.mean(tuple(range(1, len(up.shape))), keepdims=True) for up in updates]
```
mtthss, line 629: please use `g` instead

mtthss: updates can be an arbitrary data structure (e.g. a nested list) so just looping with a `for` won't be sufficient,
please try:

```
updates = jax.tree_map(
    lambda g: g - g.mean(tuple(range(1, len(g.shape))), keepdims=True), updates)
```

mtthss: Thanks for the contribution!

I sent you a suggestion on how to use `tree_map`.

Could you also expose the new function in the top level `__init__.py`?
(importing it there, and adding it to the public API defined by `__all__`).

Finally would be great to add a custom test in `transform_test.py`,
stepping the transform once on some dummy inputs,
and then comparing the outputs to some hardcoded values,

You could compute the expected targets either manually,
or even better, using the authors's implementation of the optimiser (if you can find one on github).

mtthss (approved):

optax/_src/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/home/joaogui/miniconda3/envs/optax/bin/python"
```
mtthss, line 2: Can you remove this file?

We added a .gitignore now that should avoid importing this

optax/_src/transform_test.py
```diff
@@ -154,6 +154,18 @@ def test_clip_by_global_norm(self):
       # Check that continuously clipping won't cause numerical issues.
       chex.assert_tree_all_close(updates, updates_step, atol=1e-7, rtol=1e-7)
 
+  @parameterized.named_parameters([
+      ('1d', [1.0, 2.0], [1.0, 2.0]),
+      ('2d', [[1.0, 2.0], [3.0, 4.0]], [[-0.5000,  0.5000], [-0.5000,  0.5000]]),
+      ('3d', [[[1., 2.], [3., 4.]], [[5., 6.], [7., 8.]]],
+        [[[-1.5000, -0.5000], [ 0.5000,  1.5000]], [[-1.5000, -0.5000], [ 0.5000,  1.5000]]]),
```
mtthss, line 161: Remove double spacing in array.

optax/_src/transform_test.py
```diff
@@ -154,6 +154,18 @@ def test_clip_by_global_norm(self):
       # Check that continuously clipping won't cause numerical issues.
       chex.assert_tree_all_close(updates, updates_step, atol=1e-7, rtol=1e-7)
 
+  @parameterized.named_parameters([
+      ('1d', [1.0, 2.0], [1.0, 2.0]),
+      ('2d', [[1.0, 2.0], [3.0, 4.0]], [[-0.5000,  0.5000], [-0.5000,  0.5000]]),
```
mtthss, line 159: Remove double spacing in array

optax/_src/transform.py
```diff
@@ -605,3 +605,30 @@ def update_fn(updates, state, params=None):
...
+  """
+
+  def init_fn(_):
+    return CentralState()
+
+  def update_fn(updates, state, params=None):
+    del params
+    updates = jax.tree_map(
+        lambda g: g - g.mean(tuple(range(1, len(g.shape))), keepdims=True) if len(g.shape) > 1
+          else g, updates)
```
mtthss, line 631: make this a private function

`def _subtract_mean():
    if ...:
       ...
    else:
       ...
```

instead of a lambda.

mtthss (approved):

PR #15, opened 2020-10-23
Radam

### LINKED ISSUE #8: Implement rectified Adam
See paper:

https://arxiv.org/abs/1908.03265

joaogui1:
Closes #8 
Couple things:

1. My implementation with `cond` seems hackish, any tips?
2. The paper uses 4 as threshold, but the official implementation uses 5, which value should we use?
3. It's passing parabel but failing rosenbrock (with a divergence), I will check it out on Monday
4. Any other feedback on the implementation is welcome!

### DIFF
--- a/optax/__init__.py
+++ b/optax/__init__.py
@@ -21,6 +21,7 @@ from optax._src.alias import adam
 from optax._src.alias import adamw
 from optax._src.alias import lamb
 from optax._src.alias import noisy_sgd
+from optax._src.alias import radam
 from optax._src.alias import rmsprop
 from optax._src.alias import sgd
 from optax._src.combine import chain
--- a/optax/__init__.py
+++ b/optax/__init__.py
@@ -58,6 +58,7 @@ from optax._src.transform import scale
 from optax._src.transform import scale_by_adam
 from optax._src.transform import scale_by_belief
 from optax._src.transform import scale_by_fromage
+from optax._src.transform import scale_by_radam
 from optax._src.transform import scale_by_rms
 from optax._src.transform import scale_by_rss
 from optax._src.transform import scale_by_schedule
@@ -124,11 +125,13 @@ __all__ = (
     "pathwise_jacobians",
     "piecewise_constant_schedule",
     "polynomial_schedule",
+    "radam",
     "rmsprop",
     "scale",
     "scale_by_adam",
     "scale_by_belief",
     "scale_by_fromage",
+    "scale_by_radam",
     "scale_by_rms",
     "scale_by_rss",
     "scale_by_schedule",
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -87,6 +87,17 @@ def noisy_sgd(learning_rate: float,
   )
 
 
+def radam(learning_rate: float,
+         b1: float = 0.9,
+         b2: float = 0.999,
+         eps: float = 1e-8,
+         threshold: float = 5.0) -> GradientTransformation:
+  return combine.chain(
+      transform.scale_by_radam(b1=b1, b2=b2, eps=eps, threshold=threshold),
+      transform.scale(-learning_rate),
+  )
+
+
 def rmsprop(learning_rate: float,
             decay: float = 0.9,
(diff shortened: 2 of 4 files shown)


joaogui1:
You know what, I don't like this implementation, it needs to be more modular (like the AdamW weight scaling). I will try to fix it and then ping you when it's better or I need some help

joaogui1:
So, this version almost works, except that if `ro >= threshold` we should return the SGD+momentum updates, not Adam's update, any idea how we could reverse Adam's adaptive stuff? Or do you prefer the original version where there's a `scale_by_radam` transform?  
```python
def rectify(b2: float = 0.999,      
                threshold: float = 5.0) -> GradientTransformation:
  """..."""

  ro_inf = 2./(1 - b2) - 1
  def _radam_update(args):
    ro = args[0]
    updates = args[1]
    r = jnp.sqrt((ro - 4)*(ro - 2)*ro_inf/((ro_inf - 4)*(ro_inf - 2)*ro))
    updates = jax.tree_map(lambda g: r*g , updates)
    return updates

  def init_fn(params):
    return RectifyState(b2t=b2, count=jnp.zeros([], jnp.int32))

    del params
    count_inc = _safe_int32_increment(state.count)
    b2t = state.b2t
    b2t *= b2
    ro = ro_inf - 2*count_inc*b2t/(1 - b2t)
    updates = jax.lax.cond(ro >= threshold, _radam_update, lambda _: updates,
                    (ro, updates))
    return updates, RectifyState(b2t=b2t, count=count_inc)

  return GradientTransformation(init_fn, update_fn)

```

mtthss:
I like the idea of this being decoupled from the standard Adam part similarly to adamw, however I am struggling to see the right way to achieve it. As you mentioned we would need to invert Adam when not entering the branch which differently from weight decay is something that only makes sense if you are following Adam. Therefore the split wouldn't *really* allow you to freely compose it anyway. In which case I would be inclined to go with the first version

joaogui1:
I think I've implemented it correctly, but I'm getting a divergence in the rosebrock test, any idea what could be causing it?

mtthss:
I can't spot anything obviously wrong? By divergence do you mean literal divergence to infinity?

joaogui1:
Well, looks pretty big to me
```python
FAIL: test_rosenbrock_radam (__main__.AliasTest)
...
AssertionError: 
Not equal to tolerance rtol=0.03, atol=0.03
Error in value equality check: Values not approximately equal
Mismatched elements: 2 / 2 (100%)
Max absolute difference: 3069795.5
Max relative difference: 3069795.5
 x: array([3069796.5  ,   23477.021], dtype=float32)
 y: array([1., 1.], dtype=float32)
```

mtthss:
Mmm yeah that is pretty bad indeed :(

mtthss:
Maybe you could compare the output of a single step of your radam implementation to some canonical radam implementation (even in another framework)?

joaogui1:
Hey @mtthss I changed the learning rate to match the common learning rate in TF-addons and now it passes. Think that's enough?

joaogui1:
Ok, the last message was lazy on my part, here's a [colab](https://colab.research.google.com/drive/10yXn4fKNSenkx4k8RQAGtMhZFW4BkUh3?usp=sharing) comparing my implementation and Tenforflow Addon's

1. The results are very similar in the first 10-step test with 1e-1 learning rate
2. Both converge with 1e-3 lr and 10000 steps.

joaogui1:
So @mtthss, what do you think?

mtthss (approved):

mtthss:
Looks good! I'll merge!
Thank you a lot for your contribution!

carlosgmartin:
What's the recommended way to use weight decay with [`radam`](https://optax.readthedocs.io/en/latest/api.html#optax.radam), since it doesn't have a `weight_decay` argument like [`adamw`](https://optax.readthedocs.io/en/latest/api.html#optax.adamw) does? `optax.chain(optax.add_decayed_weights(weight_decay), optax.radam(lr))`?

Issue #11, opened 2020-10-18
More examples


mtthss:
This issue provides a home for work on adding informative examples of using optax (for instance examples reproducing results from interesting optimisation papers).

Reach out on this issue if you are interested and/or have suggestions.

PostQuantum:
I want to use optax, but I don't find any tutorials to learn. Are there any documents for me to learn optax?

SauravMaheshkar:
Hey @mtthss can I work on this? Perhaps try to improve the existing example with RMSprop and SGD, and then move on to LRSchedulers.

mtthss:
@SauravMaheshkar this sounds great thanks!

SauravMaheshkar:
**Update:**
@mtthss, I've used the MNIST dataset and a `hk.module` model, and used/demonstrated how all the optimizers in the `alias.py` could be used and performed hyper-parameter optimization (using other libraries). Now I'm currently working on an example of how one could work with the second-order derivatives. Plan to make a PR by Friday 😊

mtthss:
Awesome! Thank you very much, looking forward to this!

mtthss:
Hi @SauravMaheshkar, how is the PR coming along? do you need any help?

SauravMaheshkar:
Just opened a PR. 

Metrics from all the experiments are available on [this page](https://wandb.ai/sauravmaheshkar/gradient-processing-and-optimization?workspace=).

mtthss:
Hi @SauravMaheshkar 

I am struggling a bit with interpreting the results,
do I see correctly that the train/test accuracy is around 10%?

SauravMaheshkar:
Yes @mtthss, I can't figure that out as well. I've ported the code for training from a [Haiku Example](https://github.com/deepmind/dm-haiku/blob/master/examples/mnist.py) and updated it to use `optax` syntax. 

The only reason that I can think of is perhaps I'm training for fewer epochs. The original example uses `10,000` epochs whereas I've used `{50, 100, 200}`.

What do you think ?

SauravMaheshkar:
Just Updating on my Progress. I've been running experiments with 10000 updates. Will open a PR once again, in the same format. Anything I should keep in mind?

mtthss:
Awesome! Will take a look as soon as you send the new PR!

charmerDark:
[...]

PR #23, opened 2021-02-06
Add Example

SauravMaheshkar:
This is wrt #11.

I've been running tons of experiments (**960 to be exact**) on all of the aliases from the optax library. So far I've been able to optimize using various sets of learning rates, weight decay and batch sizes.

I had some school work to do and haven't been able to work with the most custom features of the library (schedulers, second orders,.. ) That would be the next step. Hopefully, in the next PR, I'd be able to add some more examples using `measure_valued_jacobians` and `control_variates_jacobians`


SauravMaheshkar:
Will Open a new PR.

PR #27, opened 2021-02-15
Add a mask wrapper.

### LINKED ISSUE #25: Option to not decay bias with additive_weight_decay
Currently, `additive_weight_decay` will decay all the parameters. [Jia et al. 2018](https://arxiv.org/abs/1807.11205) show it is beneficial not to decay the bias parameters (and only decay the weights). Many training examples implement this too, such as [Flax's Imagenet example](https://github.com/google/flax/blob/master/examples/imagenet/train.py#L119-L121). 

One way to support this is to add a `decay_bias: bool` argument to `additive_weight_decay`, then within the decay update:

```python
updates = jax.tree_multimap(
    lambda g, p: g + weight_decay * p if p.ndim > 1 or decay_bias else g, 
    updates, params)
```

As far as I'm aware for common NN layers, the bias always has one dimension, so checking if `p.ndim > 1` is sufficient (please correct me if I'm wrong).

I'd be happy to contribute this if it sounds reasonable.

EDIT: I realized batch norm scale parameters also have only one dimension, so this filter would wrongly include those. However, many training pipelines do not regularize batchnorm scale/bias (e.g. the [reference imagenet implementation from mlperf](https://github.com/mlcommons/training/blob/4b0a874e2eb4c4d7c5e9ff299e99c4324d97acd8/image_classification/tensorflow/official/resnet/resnet_run_loop.py#L314-L329)).

n2cholas:
Closes #25

### DIFF
--- a/optax/_src/wrappers_test.py
+++ b/optax/_src/wrappers_test.py
@@ -348,7 +348,7 @@ class MaskTest(chex.TestCase):
     # Negate the updates wherever the mask is True
     def masked_negate(updates):
       return jax.tree_util.tree_multimap(
-        lambda upd, m: -upd if m else upd, updates, mask)
+          lambda upd, m: -upd if m else upd, updates, mask)
     correct_updates = masked_negate(input_updates)
 
     init_fn, update_fn = wrappers.mask(opt_builder(), mask)
@@ -405,7 +405,7 @@ class MaskTest(chex.TestCase):
         mask, input_updates, params)
 
     init_fn, update_fn = wrappers.mask(
-      transform.additive_weight_decay(weight_decay), mask)
+        transform.additive_weight_decay(weight_decay), mask)
     update_fn = self.variant(update_fn)
 
     state = init_fn(params)
--- a/optax/_src/wrappers_test.py
+++ b/optax/_src/wrappers_test.py
@@ -330,5 +330,118 @@ class LookaheadTest(chex.TestCase):
     chex.assert_tree_all_close(final_params.slow, correct_result)
 
 
+class MaskedTest(chex.TestCase):
+  """Tests for the lookahead optimizer."""
+
+  @chex.all_variants
+  @parameterized.parameters(
+      lambda: transform.scale(-1.),  # stateless test
+      _build_sgd  # stateful test
+  )
+  def test_masked(self, opt_builder):
+    mask = {'a': True,
+            'b': [False, True],
+            'c': {'d': True, 'e': (False, True)}}
+    params = {'a': 1, 'b': [2, 3], 'c': {'d': 4, 'e': (5, 6)}}
+    input_updates = jax.tree_util.tree_map(lambda x: x/10., params)
+
+    # Negate the updates wherever the mask is True
+    def masked_negate(updates):
+      return jax.tree_util.tree_multimap(
+        lambda upd, m: -upd if m else upd, updates, mask)
+    correct_updates = masked_negate(input_updates)
+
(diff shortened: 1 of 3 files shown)


optax/_src/wrappers_test.py
```diff
@@ -330,5 +330,118 @@ def test_lookahead_edge_cases(self, sync_period, slow_step_size,
     chex.assert_tree_all_close(final_params.slow, correct_result)
 
 
+class MaskedTest(chex.TestCase):
+  """Tests for the lookahead optimizer."""
```
mtthss, line 334: """Tests for the mask wrapper."""

mtthss (approved):

optax/_src/wrappers_test.py
```diff
@@ -330,5 +330,118 @@ def test_lookahead_edge_cases(self, sync_period, slow_step_size,
...
+    mask = {'a': True,
+            'b': [False, True],
+            'c': {'d': True, 'e': (False, True)}}
+    params = {'a': 1, 'b': [2, 3], 'c': {'d': 4, 'e': (5, 6)}}
+    input_updates = jax.tree_util.tree_map(lambda x: x/10., params)
+
+    # Negate the updates wherever the mask is True
+    def masked_negate(updates):
+      return jax.tree_util.tree_multimap(
+        lambda upd, m: -upd if m else upd, updates, mask)
```
mtthss, line 351: Add 2 spaces

optax/_src/wrappers_test.py
```diff
@@ -330,5 +330,118 @@ def test_lookahead_edge_cases(self, sync_period, slow_step_size,
...
+            'c': {'d': True, 'e': (False, True)}}
+    params = {'a': 1, 'b': [2, 3], 'c': {'d': 4, 'e': (5, 6)}}
+    input_updates = jax.tree_util.tree_map(lambda x: x/10., params)
+
+    correct_updates = jax.tree_util.tree_multimap(
+        lambda m, u, p: u + weight_decay * p if m else u,
+        mask, input_updates, params)
+
+    init_fn, update_fn = wrappers.mask(
+      transform.additive_weight_decay(weight_decay), mask)
```
mtthss, line 408: Add 2 white spaces

mtthss:
There is another linting issue

```
Redefining name 'mask' from outer scope (line 441) [redefined-outer-name]
	//depot/google3/third_party/py/optax/_src/wrappers.py:442:10
```

n2cholas:
Thanks for the review @mtthss and sorry about the whitespace issues. Is there a specific `pylintrc` I can use to be consistent? I'm currently using [Google's](https://google.github.io/styleguide/pyguide.html) and it didn't catch those whitespace issues (though it did catch the redefinition error).

I feel that `optax.partial` is a slightly confusing/misleading name, as it makes me think of `functools.partial` and not partial in the sense that it applies the transform to a subset of the parameters. What about `optax.masked(inner_transform, mask=...)` instead? But if you feel strongly that `optax.partial` is the right name, I'd be happy to change it.

mtthss:
I agree. Let's go with `optax.masked`

n2cholas:
Applied the change @mtthss! Sorry about all these extra commits--I rebased with master to get the `add_decayed_weights` change (so the masked docstring wouldn't be out of date), but didn't realize it would do all this to the PR. I can try to undo that if it's problematic.

n2cholas:
I dropped the commits from the rebase, all the checks pass now.

mtthss (approved):

PR #53, opened 2021-03-12
Implement DPSGD

### LINKED ISSUE #50: Implement Differentially Private Stochastic Gradient Descent
Differentially Private SGD (https://cseweb.ucsd.edu/~kamalika/pubs/scs13.pdf) is an important algorithm in private machine learning. Essentially, it is SGD except you clip and add Gaussian noise to per-example gradients before averaging across the batch. I think this would be a useful addition to Optax. 

The implementation could be based on the example in the JAX repo: https://github.com/google/jax/blob/master/examples/differentially_private_sgd.py

The usage would be slightly different from other transforms, since it requires per-example gradients as inputs.  It can still be composed with other transforms as long as it is the first one in the chain. Alternatively, we can expose a stand-alone utility function that does the clipping/noise/aggregation that the user could then pass to a GradientTransform. I think the former option (making it a transform) is more convenient since this algorithm would have some state (the RNG key).

I'd be happy to work on this if it seems like a good addition.

n2cholas:
Closes #50 

The unit tests are based on [TensorFlow Privacy's](https://github.com/tensorflow/privacy/blob/master/tensorflow_privacy/privacy/optimizers/dp_optimizer_vectorized.py). I verified the performance is similar to the JAX demo [here](https://colab.research.google.com/drive/1JnNAC2BHYJitUCSIIw1TbOSt7paysUO-?usp=sharing).

Some questions:
- For rosenbrock/parabel, I tested dpsgd by setting the norm clipping to `jnp.finfo(jnp.float32).max` and noise multiplier to 0, which makes it identical to SGD (so not really a good test :P). Do I just tune DPSGD to work on these problems with some arbitrary noise/clipping?
- Since all the transform functions are verbs (e.g. _scale_ by adam), I called the dpsgd function `differentially_private_aggregate`. The name sounds awkward to me, I'd appreciate feedback/suggestions on it.
- The loss and accuracy are pretty close to the JAX exemplar implementation, (0.18 vs. 0.22 and 96.61% vs. 96.16%), but is this close enough? Since there is randomness in this algorithm, I'm not sure if we can expect much closer.
- In the implementation, I check that the update shapes passed by the user have the required batch dimension using the shapes of `params` ([here](https://github.com/n2cholas/optax/blob/5b23f42075a34a9127764d2444b40e5173afe84d/optax/_src/transform.py#L941-L949)). Should I make `params` a required argument to `update_fn`? Or no, since the algorithm doesn't technically _need_ params? Or should I remove the check altogether?

Any other feedback would be appreciated!

### DIFF
--- a/optax/_src/transform.py
+++ b/optax/_src/transform.py
@@ -16,6 +16,8 @@
 """Gradient transformations."""
 
 from typing import Any, Callable, NamedTuple, Optional, Sequence, Tuple, Union
+from functools import partial
+
 import chex
 import jax
 import jax.numpy as jnp
@@ -893,3 +895,65 @@ def zero_nans() -> GradientTransformation:
     return updates, opt_state
 
   return GradientTransformation(init=init_fn, update=update_fn)
+
+
+class DifferentiallyPrivateAggregateState(OptState):
+  """State containing PRNGKey for differentially_private_aggregate."""
+  rng_key: jnp.array
+
+
+@partial(jax.vmap, in_axes=(0, None))
+def _clip_per_example_grads(grads_flat: Sequence[chex.Array],
+                            l2_norm_clip: float):
+  global_grad_norm = jnp.linalg.norm([
+      jnp.linalg.norm(g.ravel()) for g in grads_flat])
+  divisor = jnp.maximum(global_grad_norm / l2_norm_clip, 1.0)
+  return [g/divisor for g in grads_flat]
+
+
+def differentially_private_aggregate(l2_norm_clip: float,
+                                     noise_multiplier: float,
+                                     seed: int) -> GradientTransformation:
+  """Aggregates gradients based on the DPSGD algorithm.
+
+  Unlike other transforms, `differentially_private_aggregate` expects the input
+  updates to have a batch dimension in the 0th axis. That is, this function
+  expects per-example gradients as input (which are easy to obtain in JAX:
+  simply use `vmap(grad(fn))` instead of `grad(fn)`). It can still be
+  composed with other transformations as long as it is the first in the chain.
+
+  References:
+    [Song et al, 2013](https://cseweb.ucsd.edu/~kamalika/pubs/scs13.pdf)
+
+  Args:
+    l2_norm_clip: maximum L2 norm of the per-example gradients.
+    noise_multiplier: ratio of standard deviation to the clipping norm.
+    seed: initial seed used for the jax.random.PRNGKey
+
+  Returns:
+    A `GradientTransformation`.
+  """
+  noise_std = l2_norm_clip * noise_multiplier
+
+  def init_fn(_):
+    return DifferentiallyPrivateAggregateState(rng_key=jax.random.PRNGKey(seed))
+
+  def update_fn(updates, state, params=None):
+    grads_flat, grads_treedef = jax.tree_flatten(updates)
+
+    if params is not None and any(g.shape[1:] != p.shape
+        for g, p in zip(grads_flat, jax.tree_leaves(params))):
+      raise ValueError(
+          'Unlike other transforms, `differentially_private_aggregate` expects'
+          ' `updates` to have a batch dimension in the 0th axis. That is, this'
+          ' function expects per-example gradients as input.')
+
+    new_key, *rngs = jax.random.split(state.rng_key, len(grads_flat)+1)
+    clipped_grads_flat = _clip_per_example_grads(grads_flat, l2_norm_clip)
+    updates_flat = [
+        (g.sum(0) + noise_std * jax.random.normal(r, g.shape[1:])) / g.shape[0]
+        for r, g in zip(rngs, clipped_grads_flat)]
+    return (jax.tree_unflatten(grads_treedef, updates_flat),
+            DifferentiallyPrivateAggregateState(rng_key=new_key))
+
+  return GradientTransformation(init_fn, update_fn)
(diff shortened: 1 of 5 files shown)


mtthss:
Could you add a separate `privacy.py` file instead of adding to `transform.py`?
`transform.py` is becoming a bit "packed" so I am thinking of splitting it in a few thematic files
(everything will still be exposed flatly in the user facing API)

mtthss:
> Do I just tune DPSGD to work on these problems with some arbitrary noise/clipping?

Yeah I think that would be better thx

mtthss:
> is this close enough? Since there is randomness in this algorithm, I'm not sure if we can expect much closer.

Could you try running both a few times and compare the mean?

optax/_src/transform.py
```diff
@@ -893,3 +895,65 @@ def update_fn(updates, opt_state, params=None):
...
+  divisor = jnp.maximum(global_grad_norm / l2_norm_clip, 1.0)
+  return [g/divisor for g in grads_flat]
+
+
+def differentially_private_aggregate(l2_norm_clip: float,
+                                     noise_multiplier: float,
+                                     seed: int) -> GradientTransformation:
+  """Aggregates gradients based on the DPSGD algorithm.
+
+  Unlike other transforms, `differentially_private_aggregate` expects the input
```
mtthss, line 919: Add a `WARNING:` at the beginning

mtthss:
> Should I make params a required argument to update_fn? 

I think making it a required argument **just** for the purpose of the check seems a bit weird

mtthss:
> The unit tests are based on TensorFlow Privacy's

If you have bandwidth for this,
would you mind having an actual comparison of applying a couple of steps of your jax implementation vs the TF1 implementation in the `equivalence_test.py`?

mtthss:
For the actual algorithm implementation I'll have to defer to someone who knows the algorithm better :)

optax/_src/transform.py
n2cholas: Done!

n2cholas:
> Could you add a separate `privacy.py` file instead of adding to `transform.py`?
> `transform.py` is becoming a bit "packed" so I am thinking of splitting it in a few thematic files
> (everything will still be exposed flatly in the user facing API)

Done!

> > Do I just tune DPSGD to work on these problems with some arbitrary noise/clipping?
> 
> Yeah I think that would be better thx

Done! I use a relatively low amount of noise to ensure consistent convergence (std ~0.01).

> > is this close enough? Since there is randomness in this algorithm, I'm not sure if we can expect much closer.
> 
> Could you try running both a few times and compare the mean?

Done, I reran the reference JAX version and ours with 20 different seeds for the optimizer RNG. Results below. The +/- is the standard deviation of the values. This looks good to me, what do you think?

|         | JAX Example | Ours |
|-------|-------------------|---------|
| Loss | 0.195 +/- 0.01 |  0.198 +/- 0.01 |
| Accuracy|  0.966 +/- 0.17 | 0.965 +/- 0.16|

> > Should I make params a required argument to update_fn?
> 
> I think making it a required argument **just** for the purpose of the check seems a bit weird

I agree, I'll leave it as is.

> > The unit tests are based on TensorFlow Privacy's
> 
> If you have bandwidth for this,
> would you mind having an actual comparison of applying a couple of steps of your jax implementation vs the TF1 implementation in the `equivalence_test.py`?

Since the algorithm involves adding random noise to the gradients, how can I ensure the outputs are the same as the TF1 implementation? Would I just verify they are approximately the same after many steps...? Also, this would involve introducing a dependency to TensorFlow Privacy, which I don't think is desirable, perhaps we can verify against another source?

> For the actual algorithm implementation I'll have to defer to someone who knows the algorithm better :)

Right, I'll reach out to some folks to see if they have bandwidth then loop them in here!

n2cholas:
@pranavsubramani would you mind reviewing this implementation, focusing specifically on algorithmic correctness? Thanks in advance :D!

optax/_src/privacy.py
```diff
@@ -0,0 +1,77 @@
...
+      raise ValueError(
+          'Unlike other transforms, `differentially_private_aggregate` expects'
+          ' `updates` to have a batch dimension in the 0th axis. That is, this'
+          ' function expects per-example gradients as input.')
+
+    new_key, *rngs = jax.random.split(state.rng_key, len(grads_flat)+1)
+    global_grad_norms = jax.vmap(transform.global_norm)(grads_flat)
+    divisors = jnp.maximum(global_grad_norms / l2_norm_clip, 1.0)
+    clipped = [(jnp.moveaxis(g, 0, -1) / divisors).sum(-1) for g in grads_flat]
+    noised = [(g + noise_std * jax.random.normal(r, g.shape)) / batch_size
```
pranavsubramani, line 72: Shouldn't this be `noise_multiplier * clip_thresh` ? Technically we want to add noise proportional to `N(0, sigma**2 * c**2)`

n2cholas: Right, on line 52 I have `noise_std = l2_norm_clip * noise_multiplier`, so `noise_std * jax.random.normal(r, g.shape)` is `N(0, sigma**2 * c**2)`. Does that make sense or am I missing something still?

pranavsubramani: Yep, this looks good to me, I somehow missed that line, my bad!

pranavsubramani (approved):

mtthss (approved):

mtthss:
@n2cholas Could you add the code you used to run the dpsgd as an example in `optax/examples/`?

n2cholas:
Sure, will do!

PR #54, opened 2021-03-12
Prevent creating unnecessary momentum variables.

### LINKED ISSUE #52: Prevent creating unnecessary momentum variables
Currently, `optax.sgd` and `optax.noisy_sgd` unconditionally create momentum variables for the parameters, since both rely on `trace`. For `optax.noisy_sgd`, this is unnecessary since `decay` is always `0`. For `optax.sgd`, this is unexpected since `momentum=0` by default (and can be wasteful for large models). 

`optax.noisy_sgd` should only require `_scale_by_learning_rate` (with a negation). `optax.sgd` could conditionally add `trace` if `momentum > 0`. 

Below are the lines of code I'm referring to:

https://github.com/deepmind/optax/blob/ba0bc11d172054d65b4387ecae840c04e2bc7035/optax/_src/alias.py#L142-L148

https://github.com/deepmind/optax/blob/ba0bc11d172054d65b4387ecae840c04e2bc7035/optax/_src/alias.py#L105-L113

And here's where trace automatically creates it's state:

https://github.com/deepmind/optax/blob/ba0bc11d172054d65b4387ecae840c04e2bc7035/optax/_src/transform.py#L212-L213

n2cholas:
Closes #52.

### DIFF
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -107,7 +107,6 @@ def noisy_sgd(learning_rate: ScalarOrSchedule,
               gamma: float = 0.55,
               seed: int = 0) -> GradientTransformation:
   return combine.chain(
-      transform.trace(decay=0., nesterov=False),
       _scale_by_learning_rate(learning_rate),
       transform.add_noise(eta, gamma, seed),
   )
@@ -142,10 +141,13 @@ def rmsprop(learning_rate: ScalarOrSchedule,
 def sgd(learning_rate: ScalarOrSchedule,
         momentum: float = 0.,
         nesterov: bool = False) -> GradientTransformation:
-  return combine.chain(
-      transform.trace(decay=momentum, nesterov=nesterov),
-      _scale_by_learning_rate(learning_rate),
-  )
+  if momentum > 0.:
+    return combine.chain(
+        transform.trace(decay=momentum, nesterov=nesterov),
+        _scale_by_learning_rate(learning_rate)
+    )
+  else:
+    return _scale_by_learning_rate(learning_rate)
 
 
 def yogi(learning_rate: ScalarOrSchedule,
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -15,7 +15,7 @@
 # ==============================================================================
 """Aliases for popular optimisers."""
 
-from typing import Union
+from typing import Union, Optional
 
 import jax.numpy as jnp
 from optax._src import combine
@@ -139,15 +139,13 @@ def rmsprop(learning_rate: ScalarOrSchedule,
 
 
 def sgd(learning_rate: ScalarOrSchedule,
-        momentum: float = 0.,
+        momentum: Optional[float] = None,
         nesterov: bool = False) -> GradientTransformation:
-  if momentum > 0.:
-    return combine.chain(
-        transform.trace(decay=momentum, nesterov=nesterov),
-        _scale_by_learning_rate(learning_rate)
-    )
-  else:
-    return _scale_by_learning_rate(learning_rate)
+  return combine.chain(
+      (transform.trace(decay=momentum, nesterov=nesterov)
+       if momentum is not None else transform.identity()),
+      _scale_by_learning_rate(learning_rate)
+  )
 
 
 def yogi(learning_rate: ScalarOrSchedule,


rwightman:
@n2cholas I was looking at this in prep for a tweak to RMSProp that includes a trace, instead of the conditional and multiples chain defs, what about
```
def chain(*args: optax.GradientTransformation) -> optax.GradientTransformation:
    """Applies a list of chainable update transformations.

    Given a sequence of chainable transforms, `chain` returns an `init_fn`
    that constructs a `state` by concatenating the states of the individual
    transforms, and returns an `update_fn` which chains the update transformations
    feeding the appropriate state to each.

    Args:
      *args: a sequence of chainable (init_fn, update_fn) tuples.

    Returns:
      A single (init_fn, update_fn) tuple.
    """

    init_fns, update_fns = zip(*filter(None, args))

    def init_fn(params):
        return [fn(params) for fn in init_fns]

        if len(update_fns) != len(state):
            raise ValueError('The number of updates and states has to be the same in '
                             'chain! Make sure you have called init first!')

        new_state = []
        for s, fn in zip(state, update_fns):
            updates, new_s = fn(updates, s, params)
            new_state.append(new_s)
        return updates, new_state

    return optax.GradientTransformation(init_fn, update_fn)


def rmsprop(
        learning_rate: ScalarOrSchedule,
        decay: float = 0.9,
        momentum: float = 0.,
        eps: float = 1e-8,
        centered: bool = False,
        nesterov: bool = False,
        initial_scale: float = 0.) -> optax.GradientTransformation:

    if centered:
        return chain(
            scale_by_stddev(decay=decay, eps=eps, initial_scale=initial_scale),
            scale_by_learning_rate(learning_rate),
            optax.trace(decay=momentum, nesterov=nesterov) if momentum > 0 else None
        )
    return chain(
        scale_by_rms(decay=decay, eps=eps, initial_scale=initial_scale),
        scale_by_learning_rate(learning_rate),
        optax.trace(decay=momentum, nesterov=nesterov) if momentum > 0 else None
    )
```

n2cholas:
@rwightman good point, multiple chains are a bit ugly, but I'm not sure about adding the `None` filter to chain (seems a bit overfit to this usecase). Perhaps instead we could use:

```python
optax.trace(decay=momentum, nesterov=nesterov) if momentum > 0 else optax.identity()
```

The identity is a no-op, and this keeps `chain` clean. What do you think?

rwightman:
@n2cholas I didn't realize there was already an identity, yeah that works too.

n2cholas:
Great! For this PR, I'll keep the branch, since when momentum=0 you don't need to use `chain` at all and can just return `scale` or `scale_by_schedule`. But I agree that it'll keep the code clean for the RMSProp case.

optax/_src/alias.py
```diff
@@ -142,10 +141,13 @@ def rmsprop(learning_rate: ScalarOrSchedule,
 def sgd(learning_rate: ScalarOrSchedule,
         momentum: float = 0.,
         nesterov: bool = False) -> GradientTransformation:
-  return combine.chain(
-      transform.trace(decay=momentum, nesterov=nesterov),
-      _scale_by_learning_rate(learning_rate),
-  )
+  if momentum > 0.:
```
mtthss, line 144: return combine.chain(
        transform.trace(decay=momentum, nesterov=nesterov) if momentum > 0. else transform.identity(),
        _scale_by_learning_rate(learning_rate)

mtthss: (small preference)

optax/_src/alias.py
```diff
@@ -142,10 +141,13 @@ def rmsprop(learning_rate: ScalarOrSchedule,
 def sgd(learning_rate: ScalarOrSchedule,
         momentum: float = 0.,
```
mtthss, line 142: Lets have `None` as default,
and check `if momentum is not None`
instead of `if momentum > 0.`

this works better if for instance you want to decay the `momentum` to zero over time,

optax/_src/alias.py
n2cholas: Done!

optax/_src/alias.py
n2cholas: Done!

mtthss (approved):

PR #55, opened 2021-03-13
Add momentum and initial_scale to RMSProp

### LINKED ISSUE #49: RMSProp does not match original Tensorflow impl
A lot of papers using RMSProp have hparams that were tuned with the original Tensorflow impl. 

The Optax impl is missing the momentum option and initializes the rms value differently.

TF1 RMSProp (https://github.com/tensorflow/tensorflow/blob/v2.4.1/tensorflow/python/training/rmsprop.py#L126)
* has momentum option
* initializes RMS value to ones

Keras RMSProp (https://github.com/tensorflow/tensorflow/blob/v2.4.1/tensorflow/python/keras/optimizer_v2/rmsprop.py#L35-L299)
* has momentum option
* initializes RMS values to zeros (as with here and PyTorch)

I've spent time replicating results of papers like EfficientNet (and related) in PyTorch and ended up using my own RMSProp impl that matches the TF1 variant (the PyTorch one does not either).

rwightman:
Adds 
* momentum trace to RMSProp
* initial_scale to allow setting initial value of second moment for RMS to 1.0 to match original Tensorflow behaviour

Fixes #49


optax/_src/alias.py
```diff
@@ -126,16 +126,25 @@ def radam(learning_rate: ScalarOrSchedule,
 
 def rmsprop(learning_rate: ScalarOrSchedule,
             decay: float = 0.9,
+            momentum: float = 0.,
```
mtthss, line 129: Lets have `None` as default,
and check `if momentum is not None`

rwightman: change made

mtthss: Could you also move `initial_scale` right after `eps` (so that all the arguments of `scale_by_rms`  stay next to each other), and then move `momentum` at the end just before `nesterov`?

rwightman: done

mtthss (approved):

PR #83, opened 2021-03-28
Optax: Add AdaBelief optimizer description, reference, arg doc strings

8bitmp3:
This PR adds a short description, a paper reference, and arg doc strings to the AdaBelief optimizer in Optax, since it was not documented. AdaBelief seems like a very notable new optimizer we should be using.

```
  ...
  AdaBelief is an adaptive learning rate optimiser that focuses on fast
  convergence, generalisation, and stability. It adapts the step size depending
  on its "belief" in the gradient direction — the optimiser adaptively scales the
  step size by the difference between the predicted and observed gradients.
  AdaBelief is a modified version of Adam and contains the same number of
  parameters.
  ...
```

- Used the recent NeurIPS paper to prepare the docs https://arxiv.org/abs/2010.07468
- Note that the AdaBelief args are the same as in Adam (it's a "modified Adam")
- Added dashes (`-`) to `first- and second-order` in Adam
- Used British English in line with the rest of the docs (`optimiser` 🇬🇧)

Related to https://github.com/deepmind/optax/pull/71 by @mtthss 

LMKWYT @mtthss Cheers 👍

### DIFF
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -44,8 +44,8 @@ def adabelief(
 
   AdaBelief is an adaptive learning rate optimiser that focuses on fast
   convergence, generalisation, and stability. It adapts the step size depending
-  on its "belief" in the gradient direction — the optimiser adaptively scales the
-  step size by the difference between the predicted and observed gradients.
+  on its "belief" in the gradient direction — the optimiser adaptively scales
+  the step size by the difference between the predicted and observed gradients.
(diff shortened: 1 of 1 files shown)


mtthss (approved):

mtthss:
Thanks! The lint fails because of a line too long error

8bitmp3:
Thanks @mtthss, fixed it I think, PTAL 👍

mtthss (approved):

PR #89, opened 2021-04-03
Implement DPSGD example.

n2cholas:
DPSGD example as requested in #53. Verified the accuracies are close with a few seeds [here](https://colab.research.google.com/drive/107SrfBIyZdD9ES67xEvuNy0J8kUEirl_?usp=sharing). This is based on the [`differentially_private_sgd.py`](https://github.com/google/jax/blob/master/examples/differentially_private_sgd.py) example in the JAX repo, but refactored to better match Optax's style.

Questions:
1. How do I run these examples? I couldn't get the `from optax.examples import datasets` import to work when running it via `python3 dpsgd_example.py`
2. Is the name ok? The JAX example named it `differentially_private_sgd.py` I used `dpsgd_example.py` because `differentially_private_sgd_example.py` is kind of long.
3. The default seed of 0 results in a slightly lower accuracy than the reported (just like we saw in #53), should I switch the default seed?
4. Should I create a `requirements.txt` specifically for this example? Or is the comment in the file enough (that specifies how to install `tensorflow_privacy`.

Additional questions/concerns listed as review comments below.

cc: @mtthss

### DIFF
--- a/examples/dpsgd_example.py
+++ b/examples/dpsgd_example.py
@@ -1,0 +1,185 @@
+# Copyright 2021 DeepMind Technologies Limited. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""JAX efficiently trains a differentially private conv net on MNIST.
+This script contains a JAX implementation of Differentially Private Stochastic
+Gradient Descent (https://arxiv.org/abs/1607.00133). DPSGD requires clipping
+the per-example parameter gradients, which is non-trivial to implement
+efficiently for convolutional neural networks.  The JAX XLA compiler shines in
+this setting by optimizing the minibatch-vectorized computation for
+convolutional architectures. Train time takes a few seconds per epoch on a
+commodity GPU.
+This code depends on tensorflow_privacy (https://github.com/tensorflow/privacy)
+  Install instructions:
+    $ pip install tensorflow git+https://github.com/tensorflow/privacy.git
+The results match those in the reference TensorFlow baseline implementation:
+  https://github.com/tensorflow/privacy/tree/master/tutorials
+Example invocations:
+  # this non-private baseline should get ~99% acc
+  python -m examples.differentially_private_sgd \
+    --dpsgd=False \
+    --learning_rate=.1 \
+    --epochs=20 \
+   this private baseline should get ~95% acc
+  python -m examples.differentially_private_sgd \
+   --dpsgd=True \
+   --noise_multiplier=1.3 \
+   --l2_norm_clip=1.5 \
+   --epochs=15 \
+   --learning_rate=.25 \
+  # this private baseline should get ~96.6% acc
+  python -m examples.differentially_private_sgd \
+   --dpsgd=True \
+   --noise_multiplier=1.1 \
+   --l2_norm_clip=1.0 \
+   --epochs=60 \
+   --learning_rate=.15 \
+  # this private baseline should get ~97% acc
+  python -m examples.differentially_private_sgd \
+   --dpsgd=True \
+   --noise_multiplier=0.7 \
+   --l2_norm_clip=1.5 \
+   --epochs=45 \
+   --learning_rate=.25 \
+"""
+
+import time
+import warnings
+
+from absl import app
+from absl import flags
+
+import jax
+import jax.numpy as jnp
+from jax.experimental import stax
+import optax
+
+import datasets
+
+# https://github.com/tensorflow/privacy
+from tensorflow_privacy.privacy.analysis.rdp_accountant import compute_rdp
+from tensorflow_privacy.privacy.analysis.rdp_accountant import get_privacy_spent
+
+NUM_EXAMPLES = 60_000
+FLAGS = flags.FLAGS
+
+flags.DEFINE_boolean(
+    'dpsgd', True, 'If True, train with DP-SGD. If False, '
+    'train with vanilla SGD.')
+flags.DEFINE_float('learning_rate', .15, 'Learning rate for training')
+flags.DEFINE_float('noise_multiplier', 1.1,
+                   'Ratio of the standard deviation to the clipping norm')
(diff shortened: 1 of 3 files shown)


examples/dpsgd_example.py
```diff
@@ -0,0 +1,185 @@
+# Copyright 2021 DeepMind Technologies Limited. All Rights Reserved.
```
n2cholas, line 1: Is this copyright header ok? A big portion of the code is from the JAX repo.

examples/dpsgd_example.py
```diff
@@ -0,0 +1,185 @@
...
+This script contains a JAX implementation of Differentially Private Stochastic
+Gradient Descent (https://arxiv.org/abs/1607.00133). DPSGD requires clipping
+the per-example parameter gradients, which is non-trivial to implement
+efficiently for convolutional neural networks.  The JAX XLA compiler shines in
+this setting by optimizing the minibatch-vectorized computation for
+convolutional architectures. Train time takes a few seconds per epoch on a
+commodity GPU.
+This code depends on tensorflow_privacy (https://github.com/tensorflow/privacy)
+  Install instructions:
+    $ pip install tensorflow git+https://github.com/tensorflow/privacy.git
```
n2cholas, line 25: Note: these installation instructions are slightly different from the [original JAX version](https://github.com/google/jax/blob/master/examples/differentially_private_sgd.py#L27-L30).

examples/dpsgd_example.py
```diff
@@ -0,0 +1,185 @@
...
+   --epochs=60 \
+   --learning_rate=.15 \
+  # this private baseline should get ~97% acc
+  python -m examples.differentially_private_sgd \
+   --dpsgd=True \
+   --noise_multiplier=0.7 \
+   --l2_norm_clip=1.5 \
+   --epochs=45 \
+   --learning_rate=.25 \
+"""
```
n2cholas, line 55: TODO: Change these after figuring out the correct way to run the code.

examples/dpsgd_example.py
```diff
@@ -0,0 +1,185 @@
...
+
+from absl import app
+from absl import flags
+
+import jax
+import jax.numpy as jnp
+from jax.experimental import stax
+import optax
+
+import datasets
```
n2cholas, line 68: TODO: Change this import to `from optax.examples import datasets` after figuring out how to run the code.

examples/dpsgd_example.py
```diff
@@ -0,0 +1,185 @@
...
+                   'Ratio of the standard deviation to the clipping norm')
+flags.DEFINE_float('l2_norm_clip', 1.0, 'Clipping norm')
+flags.DEFINE_integer('batch_size', 256, 'Batch size')
+flags.DEFINE_integer('epochs', 60, 'Number of epochs')
+flags.DEFINE_integer('seed', 0, 'Seed for jax PRNG')
+flags.DEFINE_integer(
+    'microbatches', None, 'Number of microbatches '
+    '(must evenly divide batch_size)')
+flags.DEFINE_string('model_dir', None, 'Model directory')
+flags.DEFINE_float('delta', 1e-5, 'Target delta used to compute privacy spent.')
```
n2cholas, line 91: NOTE: Not `delta` flag not present in original JAX example, added because it was easy to, could remove if we want to stay closer to the original.

examples/dpsgd_example.py
```diff
@@ -0,0 +1,185 @@
...
+@jax.jit
+def test_step(params, batch):
+  loss, logits = loss_fn(params, batch)
+  accuracy = (logits.argmax(1) == batch['label'].argmax(1)).mean()
+  return loss, accuracy * 100
+
+
+def main(_):
+  train_dataset = datasets.load_image_dataset('mnist', FLAGS.batch_size)
+  train_dataset = list(train_dataset.as_numpy_iterator())
```
n2cholas, line 132: When used in the training loop as `train_dataset.as_numpy_iterator()`, the time per private epoch is ~4 seconds, which is significantly slower than this using list (which results in a per private epoch time of ~0.7 seconds). I tried using `prefetch`, which didn't help. How else can I improve this?

examples/dpsgd_example.py
mtthss: Yes this is fine, but please add an attribution in the file doc-string explaining that a big portion of the code is forked from the JAX example

examples/dpsgd_example.py
mtthss: LGTM

mtthss (commented):
2. The name is good
3. Feel free to change the seed
4. Any downside in just adding it to `requirements/requirements-examples.txt`? This is not automatically pulled in when installing optax so its fine to depend on more things in the examples

examples/dpsgd_example.py
n2cholas: Done!

examples/dpsgd_example.py
n2cholas: Will resolve this because I moved the installation to `requirements/requirements-examples.txt`

n2cholas:
> 1. The name is good

I switched the name back to `differentially_private_sgd.py`, since it's in the `examples/` directory, I thought the `_example` suffix was redundant.

> 2. Feel free to change the seed

Done!

> 3. Any downside in just adding it to `requirements/requirements-examples.txt`? This is not automatically pulled in when installing optax so its fine to depend on more things in the examples

That sounds good, done! I also needed to add `tensorflow` and `tensorflow_datasets` to that file to make it complete.

I also reran the colab with the exact commands specified in the docstring, with the dependencies installed via `requirements/requirements-examples.txt` instead of manually.

mtthss (approved):

mtthss:
Apologies for the delay in merging.
There is an issue with the ordering of imports
Could you update it to:
```
import time
import warnings

from absl import app
from absl import flags
import datasets
import jax
from jax.experimental import stax
import jax.numpy as jnp
import optax
from tensorflow_privacy.privacy.analysis.rdp_accountant import compute_rdp
from tensorflow_privacy.privacy.analysis.rdp_accountant import get_privacy_spent
```

n2cholas:
> import time
> import warnings
> 
> from absl import app
> from absl import flags
> import datasets
> import jax
> from jax.experimental import stax
> import jax.numpy as jnp
> import optax
> from tensorflow_privacy.privacy.analysis.rdp_accountant import compute_rdp
> from tensorflow_privacy.privacy.analysis.rdp_accountant import get_privacy_spent

Sure, done! I want to confirm that `import datasets` is ok, since the other examples use `from optax.examples import datasets`. The only reason I used `import datasets` instead is because I couldn't figure how to run the examples with the `from optax.examples import datasets` import.

8bitmp3:
Looks good @n2cholas. Noticed the PR is a bit stale, @mtthss - does it need to be reapproved since 15 April? Thanks all, I'm learning a lot

mtthss (approved):

PR #132, opened 2021-06-08
Implement multi_transform

### LINKED ISSUE #129: Feature Proposal: multi-optimizer / multi-transform
Currently, to apply different optimizers to different sets of parameters, you need to construct multiple masks and chain them. For example, to optionally apply weight decay to some parameters and not to others:

```py
tx = optax.chain(
    optax.masked(optax.adamw(0.01, weight_decay=0.0), 
                 mask=partial(jax.tree_map, lambda p: p.ndim == 1)),
    optax.masked(optax.adamw(0.01, weight_decay=1e-3), 
                 mask=partial(jax.tree_map, lambda p: p.ndim != 1)),
)
```

However, this has a few issues:
1. When you have more than two groups, creating mask functions for them all becomes verbose
2. The burden of verifying the masked groups are mutually exclusive falls on the user (which can be bothersome for groups > 2)
3. This solution isn't obvious to a new user

I'd like to propose a `multi_transform` (or `multi_transformation`, as always naming is hard) that generalizes `optax.masked`:

```py
def multi_transform(transforms: Sequence[GradientTransformation], 
                    partition: Union[PyTree, Callable[[base.Params], PyTree]]):
  ... 
```
Where `partition` is a pytree where the leaves contain an index (or a function that returns such a pytree given the parameters). The index corresponds to a `GradientTransformation` in `transforms`.

With this, our above example would be:
```py
tx = optax.multi_transform(
    [optax.adamw(0.01, weight_decay=1e-3), optax.adamw(0.01, weight_decay=0.0)],
    partition=partial(jax.tree_map, lambda p: int(p.ndim == 1)))  # 0 for weight decay, 1 for no weight decay
```

This solves our above problems:
1. You only need one pytree (function)
2. The groups are mutually exclusive by design, since there's only one pytree with indices
3. The name matches what users would expect from such a feature (due to similar naming in other libraries, e.g. [Flax](https://flax.readthedocs.io/en/latest/flax.optim.html#flax.optim.MultiOptimizer)).

Minor details:
- `partition` could just be `partition_fn: Callable[[base.Params], PyTree]` (I left it as `partition` to mirror the signature of `masked`).
- `partition` could have a different name, perhaps `indices`.

cc: @mtthss @jheek @andsteing

n2cholas:
Closes #129

Should this function also accept a raw PyTree, just like `optax.masked`? I think it makes more sense for the arguments to be independent of the param structure. But, that introduces some API inconsistency between this function and `masked`


mtthss:
I think it would be nice if does support both. Are there any downsides?

n2cholas:
I don't think there are any downsides (apart from a _slightly_ more complex signature). Implementation wise it's pretty simple (same as how we check in `masked`). I'll update the implementation!

mtthss:
Looks great to me,

@jheek
@avital

Any thoughts?

jheek:
LGTM, should there maybe be a sentinel mapping like `None` means "not optimized"?

n2cholas:
@jheek users can use the `optax.identity()` transformation for parameters that shouldn't be updated. I agree a detailed example would be helpful in the docstring, I'll work on that.

mtthss:
Could you move this to `_src/combine.py`?

mtthss (approved):

PR #165, opened 2021-07-28
Explictly pass fast params to fast optimizer in lookahead

jtamanas:
Passing LookaheadParams causes issues with non-adam optimizers


mkunesch (approved):
Thanks a lot for the fix!

PR #183, opened 2021-09-07
Bugfix for optax.maybe_update

lukaszlew:


optax/_src/wrappers.py
```diff
@@ -379,7 +379,7 @@ def do_update(_):
       return inner.update(updates, state.inner_state, params)
 
     def reject_update(_):
-      return updates, state.inner_state
+      return jax.tree_map(jnp.zeros_like, updates), state.inner_state
```
fabioviola, line 382: This proposed change breaks the intended behavior of the function. 
As per the docstring: 

>  When not calling the inner update function, the `updates` and the inner state are left untouched and just passed through.

mkunesch:
Thanks a lot for the comment! Since this is the intended behaviour I'm closing the PR, but we have added an issue #185 to add documentation to help make this clearer.

PR #186, opened 2021-09-14
[Adafactor] add weight decay mask

patrickvonplaten:


optax/_src/alias.py
```diff
@@ -111,6 +112,10 @@ def adafactor(
       weight_decay_rate: (float) optional rate at which to decay weights.
       eps: (float) regularization constant for root mean squared gradient.
       factored: (bool) whether to use factored second-moment estimates.
+      mask: a tree with same structure as (or a prefix of) the params PyTree,
```
patrickvonplaten, line 115:
```suggestion
      weight_decay_mask: a tree with same structure as (or a prefix of) the params PyTree,
```

optax/_src/alias.py
```diff
@@ -83,6 +83,7 @@ def adafactor(
     weight_decay_rate: Optional[float] = None,
     eps: float = 1e-30,
     factored: bool = True,
+    weight_decay_mask: Optional[Union[Any, Callable[[base.Params], Any]]] = None,
```
merrymercy, line 86: I think `weight_decay_mask` is definitely a better name.
Should we update all `mask` in existing optimizers for weight decay to `weight_decay_mask`?

merrymercy (commented):
Thanks for your contribution! I also need this feature. Could you please fix the CI errors and let the mainterins merge this?

optax/_src/alias.py
merrymercy: I mean these lines @mtthss 
https://github.com/deepmind/optax/blob/3d666d92211dc0e2689c3acf82656d6286952733/optax/_src/alias.py#L354
https://github.com/deepmind/optax/blob/3d666d92211dc0e2689c3acf82656d6286952733/optax/_src/alias.py#L223

mkunesch (approved):
Thanks a lot for adding this feature! Looks good to me.

PR #187, opened 2021-09-20
change int64 to int32 to avoid warnings

rdaems:
I have some annoying warnings because the int64 are converted to int32 on my GPU.
They can easily be avoided by changing the dtypes to int32 (which is quite sufficient).

### DIFF
--- a/optax/_src/wrappers.py
+++ b/optax/_src/wrappers.py
@@ -124,9 +124,9 @@ def apply_if_finite(
 
   def init(params):
     return ApplyIfFiniteState(
-        notfinite_count=jnp.zeros([], jnp.int64),
+        notfinite_count=jnp.zeros([], jnp.int32),
         last_finite=jnp.array(True, jnp.bool_),
-        total_notfinite=jnp.zeros([], jnp.int64),
+        total_notfinite=jnp.zeros([], jnp.int32),
         inner_state=inner.init(params))
 
   def update(updates, state, params=None):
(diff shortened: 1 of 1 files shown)


mkunesch (changes requested):
Hi! Thanks a lot for pointing this out and filing a PR!

For completeness and for others who are seeing this, I think the warning you are referring to is:
```
UserWarning: Explicitly requested dtype <class 'jax._src.numpy.lax_numpy.int64'> requested in zeros is not available, and will be truncated to dtype int32.
```
This warning should only appear the first time a jit compiled function is called.

We have to be a bit careful with this change I think because there are models that are trained for long enough to exceed 2^31 steps. Having said this, all other parts of optax use (safely incremented) int32s so I think it makes sense to bring the wrappers in line.

Could you modify the PR to use `numerics.safe_int32_increment` for all the counters you have changed to int32? This ensures that users don't get completely surprising behaviour due to a negative step count.

rdaems:
I hope I did it right now, I changed all `+1` occurences with `safe_int32_increment`.
All tests are ok. And I fetched upstream.

Now I do have errors when I use the faulty dtype. It seems everything is covered.

mkunesch (changes requested):
Hi @rdaems! Thanks a lot for making the changes - this looks great! I have found one more place where a safe increment should be used (total_notfinite wasn't yet incremented safely). Could you add it there too?

Thanks a lot again for raising this! The change doesn't only get rid of warning, it also makes the code safer: previously if int64 wasn't supported, users would get int32 with unsafe increments that might lead to negative numbers. It's good to change this to guarantee "safe" increments on all architectures.

mkunesch:
Btw: this is a small change so I'd be happy to approve the PR and patch the change in myself as it is merged. Let me know whether that's okay for you :-)! Thanks a lot again!

rdaems:
Hi @mkunesch 
Sorry for my slow response. I hope I did that last one correct.

Nice to hear that the change also makes the code safer!

mkunesch (approved):
Looks great! Thank you very much!

PR #258, opened 2021-12-05
Add online newton optimizer

eserie:


mkunesch:
Hi! Thanks a lot for the PR! Could you fix the pylint errors before we review (so that we can focus on the important things in the review)? Thanks a lot!

eserie:
Hi, sure. I fixed the pylint error in an additional commit. Feel free to ask me to change things. 
For now, the dev is more or less as I originally developed it and the test coverage is certainly a bit too limited.

mkunesch:
Hi! Thanks for fixing the pylint error - could you also do the same in the test file? I think that's still holding up the checks.
Due to the holidays it might take us slightly longer to review so there is no rush.

eserie:
Hi! I fixed the tests! Normally all checks should pass now.

mkunesch:
Great, thank you very much for fixing all the checks! I've assigned myself and should have time to review this week.
Thanks a lot again!

optax/_src/alias_test.py
```diff
@@ -125,5 +125,24 @@ def test_explicit_dtype(self, dtype):
     self.assertEqual(expected_dtype, adam_state.mu.dtype)
 
 
+  def test_newton(self):
```
mkunesch, line 128: Nit: I know most tests in this file don't do this, but in theory test names and/or docstrings should explain what the test checks. An example here could be `test_online_newton_step_decreases_loss(self)`, with a similar single-line docstring. This will be shown in the error message from the test when the tests fails and is therefore very helpful when debugging.

optax/_src/transform.py
```diff
@@ -869,6 +869,64 @@ def update_fn(updates, state, params=None):
   return base.GradientTransformation(init_fn, update_fn)
 
 
+class ScaleByNewtonState(NamedTuple):
+  """State holding the inverse of the sum of gradient squares to date."""
+
```
mkunesch, line 874: Nit: I think we don't have this whitespace for other states.

optax/_src/transform.py
```diff
@@ -869,6 +869,64 @@ def update_fn(updates, state, params=None):
...
+  hessian_inv: base.Updates
+
+
+def sherman_morrison(a_inv, u, v):
```
mkunesch, line 878: I think I'd expect this to be a private function or is there a use-case of `sherman_morrison` directly which we would like to support?

mkunesch, line 878: Please add type annotations to this function (even if we decide to make it private in the end).

optax/_src/transform.py
```diff
@@ -869,6 +869,64 @@ def update_fn(updates, state, params=None):
...
+  den = 1.0 + (u @ a_inv) @ v
+  a_inv -= a_inv @ jnp.outer(u, v) @ a_inv / den
+  return a_inv
+
+
+def scale_by_newton(eps: float = 1e-7) -> base.GradientTransformation:
+  """Rescale updates by multiplying by the inverse of an approximation of the
```
mkunesch, line 885: Please shorten this line so that the short description fits onto a single line (with more detail in the detailed docstring if necessary).

optax/_src/transform.py
```diff
@@ -869,6 +869,64 @@ def update_fn(updates, state, params=None):
...
+  hessian.
+
+  References
```
mkunesch, line 888: As above, please format this similar to e.g. `rmsprop` in `alias.py` disabling and re-enabling the `line-too-long` warning if necessary.

optax/_src/transform.py
```diff
@@ -869,6 +869,64 @@ def update_fn(updates, state, params=None):
...
+  def init_fn(params):
+    hessian_inv = jax.tree_map(
+      lambda t: jnp.eye(len(t.flatten()), dtype=t.dtype) / eps, params
+    )
+    return ScaleByNewtonState(hessian_inv=hessian_inv)
+
+  def update_fn(updates, state, params=None):
+    del params
+
+    class Tuple(tuple):
```
mkunesch, line 910: Do I understand correctly that you wouldn't need this if you don't overwrite `updates` in line 917 and instead introduce a new variable e.g. `flattened_updates` and then in line 922:

```
updates = jax.tree_multimap(
  lambda flat_u, u: flat_u.reshape(u.shape), flattened_updates, updates)
```

If that's true I think it would be good to simplify the code and remove the definition the `Tuple`.

optax/_src/transform_test.py
```diff
@@ -202,6 +202,16 @@ def test_centralize(self, inputs, outputs):
...
 
+  def test_sherman_morrisson(self):
+    x = 1
+    a_inv = jnp.eye(3) * 1.0
+    u = v = jnp.ones(3)
+
+    ref = jnp.linalg.inv(a_inv + x * jnp.outer(u, v))
+
+    a_inv = transform.sherman_morrison(a_inv, u, v)
+    assert jnp.allclose(a_inv, ref)
```
mkunesch, line 213: Please use the corresponding unittest/absltest/np.testing asserts so that the error messages from the test are more useful - in this case I think it would be `np.testing.assert_allclose`.

optax/_src/transform_test.py
```diff
@@ -202,6 +202,16 @@ def test_centralize(self, inputs, outputs):
     centralized_inputs, _ = centralizer.update(inputs, None)
     chex.assert_tree_all_close(centralized_inputs, outputs)
 
+  def test_sherman_morrisson(self):
+    x = 1
+    a_inv = jnp.eye(3) * 1.0
```
mkunesch, line 207: What does the `* 1.0` do?

optax/_src/transform_test.py
```diff
@@ -202,6 +202,16 @@ def test_centralize(self, inputs, outputs):
     centralized_inputs, _ = centralizer.update(inputs, None)
     chex.assert_tree_all_close(centralized_inputs, outputs)
 
+  def test_sherman_morrisson(self):
+    x = 1
```
mkunesch, line 206: This is a very special value: in line 210 it will mean that `x` doesn't have an effect. Could you set up the test such that `x` is used in a non-trivial way?

optax/_src/transform_test.py
```diff
@@ -202,6 +202,16 @@ def test_centralize(self, inputs, outputs):
...
+    a_inv = jnp.eye(3) * 1.0
+    u = v = jnp.ones(3)
```
mkunesch, line 208: All of these test values look like special cases to me. Could you add a test with more general conditions? (If the special case is important to test please keep it around as a separate test.)

optax/_src/transform_test.py
```diff
@@ -202,6 +202,16 @@ def test_centralize(self, inputs, outputs):
     centralized_inputs, _ = centralizer.update(inputs, None)
     chex.assert_tree_all_close(centralized_inputs, outputs)
 
+  def test_sherman_morrisson(self):
```
mkunesch, line 205: If we make the function `sherman_morrison` private, it shouldn't be tested here, but instead will be tested indirectly through the test of the public `scale_by_online_newton`.

mkunesch: typo: extra s in morrison.

optax/_src/transform_test.py
```diff
@@ -202,6 +202,16 @@ def test_centralize(self, inputs, outputs):
...
+    x = 1
+    a_inv = jnp.eye(3) * 1.0
+    u = v = jnp.ones(3)
+
+    ref = jnp.linalg.inv(a_inv + x * jnp.outer(u, v))
+
+    a_inv = transform.sherman_morrison(a_inv, u, v)
+    assert jnp.allclose(a_inv, ref)
+
```
mkunesch, line 214: Please add a test for `scale_by_newton` that tests both `init` and `update`. The update function can be tested e.g. by comparing an analytic calculation against a step taken with the code.

optax/_src/transform.py
```diff
@@ -869,6 +869,64 @@ def update_fn(updates, state, params=None):
...
+
+def scale_by_newton(eps: float = 1e-7) -> base.GradientTransformation:
+  """Rescale updates by multiplying by the inverse of an approximation of the
+  hessian.
+
+  References
+  ----------
+  [^1]: [Hazan, E., Agarwal, A. and Kale, S., 2007. Logarithmic regret
+  algorithms for online convex optimization. Machine Learning, 69(2-3),
+  pp.169-192](https://www.cs.princeton.edu/~ehazan/papers/log-journal.pdf)
```
mkunesch, line 892: Could you add which algorithm/formula/section it is in the paper so that it's easier for the reader to cross-reference?

optax/_src/transform.py
```diff
@@ -869,6 +869,64 @@ def update_fn(updates, state, params=None):
...
+
+  Args:
+    eps: A small floating point value to avoid zero denominator.
+
+  Returns:
+    An (init_fn, update_fn) tuple.
+  """
+
```
mkunesch, line 900: Note: I haven't reviewed the rest of this file yet as I'm not sure which equations exactly they implement in the paper. Please point me to the algorithm, equation, or section. Thanks a lot!

mkunesch:
Hi! Thank you so much for the contribution again!

I have been reviewing your code, but in order to check the equations it would be very helpful if you could point me to exactly the equation, algorithm, or section in the paper that you have implemented so that I can compare it to the code more easily.

I will add the comments I have so far below so that they don't get lost in the meantime, but please don't make the changes yet as I'm still reviewing the code.

Thanks a lot for your help and for this PR!

mkunesch (changes requested):
Hi! Thank you so much for the contribution again!

I have been reviewing your code, but in order to check the equations it would be very helpful if you could point me to exactly the equation, algorithm, or section in the paper that you have implemented so that I can compare it to the code more easily.

I will add the comments I have so far below so that they don't get lost in the meantime, but please don't make any changes yet as I'm still reviewing the code.

Thanks a lot for your help and also thanks again for this PR!

optax/_src/transform.py
eserie: Hi @mkunesch, thanks for your review! 
For your last question, the algorithm implemented is the ONS (ONLINE NEWTON STEP) Fig.2 in the paper.

mkunesch: Amazing, thanks a lot - now I understand! I have a few follow-up questions but I'll ask those for specific lines.
(It might be worth adding this to the doc string just because there is additional boilerplate in the paper and it takes a while to see why only the terms in the code remain - thanks a lot!)

optax/_src/transform.py
```diff
@@ -869,6 +869,64 @@ def update_fn(updates, state, params=None):
...
+  def init_fn(params):
+    hessian_inv = jax.tree_map(
+      lambda t: jnp.eye(len(t.flatten()), dtype=t.dtype) / eps, params
```
mkunesch, line 903: I have never implemented these sherman-morrison updates so I don't know whether this is actually an issue, but from looking at the code, I'd be worried about catastrophic cancellation when starting with jnp.eye / eps which has diagonal entries of O(10^7) and doing lots of additions and subtractions in the first step.

Is this something you have thought about? Is this a problem here? Perhaps there is a way to take the limit eps -> 0 explicitly for the first step to avoid this?

Thanks a lot!

optax/_src/transform.py
mkunesch: If we decide to make this private, I'd also change the implementation to specific to our use case where u == v. In any case, please also add a docstring explaining what this does and what sherman-morrison is used for in this case. Thanks a lot!

optax/__init__.py
```diff
@@ -29,6 +29,7 @@
 from optax._src.alias import sgd
 from optax._src.alias import sm3
 from optax._src.alias import yogi
+from optax._src.alias import newton
```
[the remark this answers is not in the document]
eserie: Ok, I will correct it (along with the change of name to online_newton)!
(Suggestion: using isort could permit to manage this automatically.)

optax/_src/alias.py
```diff
@@ -667,3 +667,31 @@ def dpsgd(
        if momentum is not None else base.identity()),
       _scale_by_learning_rate(learning_rate)
   )
+
+
+def newton(learning_rate: ScalarOrSchedule,
+           eps: float = 1e-5) -> base.GradientTransformation:
+  """The online newton optimizer.
+
+  It extends optax optimizers.
```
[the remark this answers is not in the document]
eserie: This is for pep8 formatting. What is the number of characters per line you accept in the project?
(I used "black" which defaults to 88 characters per line).

optax/_src/transform.py
eserie: Yes the correct link is this one: https://link.springer.com/content/pdf/10.1007/s10994-007-5016-8.pdf

optax/_src/transform.py
eserie: You are absolutely right.  In fact, in practice, this parameter may not be small. It should be chosen with some apriori knowledge about the scale of parameters and gradients (e.g. by following the formulas in the article).
Perhaps we should not define a default value for this parameter to let the user choose it?

eserie:
Hi @mkunesch, thanks for your review.
I tried to address your remarks and also made a rebase against the master branch.
Let me know if anything is wrong.

optax/_src/transform.py
mkunesch: Okay, yes, letting the user choose it is helpful - do you have an intuition for how generally applicable the default value would be? If it always has to be tuned, I'd remove the default entirely and force the user to choose it.

Also, is there an explanation (or a paper or reference implementation) we can give the user to help them choose `eps`? I don't think the paper presents experiments with this, does it?

mkunesch:
Hi @eserie!

How you are getting on with the other changes? No rush at all, but I wanted to make sure you are not waiting for my review of the changes you have already made in the meantime! Thanks a lot for making them - I've added one comment on the change re `eps`.

Thanks a lot!

eserie:
Hi @mkunesch !
I think I reviewed all of your comments and I should have addressed them. Can you tell me if sounds good to you?

optax/_src/alias_test.py
```diff
@@ -125,5 +125,24 @@ def test_explicit_dtype(self, dtype):
...
+
+    def loss(w, x):
+      return -(w * x).sum() + (w ** 2).sum()
+
+    w = jnp.ones(3)
+    x = jnp.ones(3)
+
+    opt = alias.newton(1.0e-3)
+    l_, grads = jax.value_and_grad(loss)(w, x)
```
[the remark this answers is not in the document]
mkunesch: @eserie

optax/_src/transform_test.py
mkunesch: @eserie

optax/_src/transform.py
mkunesch: @eserie

optax/_src/transform_test.py
mkunesch: @eserie

optax/_src/transform.py
mkunesch: @eserie

mkunesch:
Hi! 
Ah, I think GitHub collapsed these comments in the discussion view. There were a few more - I've tagged in you in some (but not all) of them so that GitHub shows them for you. They should also be visible in the _Files Changed_ view.

Some of them are with regards to the tests - but we could also merge into experimental where we require less testing for new code.

mtthss:
Hello @eserie, can we get this submitted?

eserie:
Hello @mtthss, @mkunesch , sorry I've been pretty busy the last 3 months. @mkunesch, I can see your comments in  Files Changed view, sorry I missed them before. Hopefully, I can take a look at it by the end of the month, if that's ok with you.

optax/_src/transform.py
eserie: Thank you for the observation! I was a bit messy in my implementation, your proposal seems much more reasonable. I updated the code accordingly and also add a test to cover the multi-dimensional case.

eserie:
Hi, I finally get to run over all your comment and added a new test for the multi-dimensional weights case. 
I finally did a rebase of my changes on the current master branch and fuse all my commits in one.
I also ran the tests and checks as in the test.sh file, everithing looks good.

mtthss:
There is still one change requested by @mkunesch, could you address and then we get this submitted? thanks a lot for the contribution!

mtthss:
This branch has conflicts that must be resolved

mkunesch:
Hi! Thanks a lot @eserie for making the changes. I just returned from holiday and will make sure to review by the end of the week.

eserie:
Hi @mkunesch, thank you! Let me know if everything looks good to you. Unfortunatly, I accidentally closed the pull request while rebasing my branch. Can you open it again? The last commit to reattach is [fa028f5](https://github.com/eserie/optax/commit/fa028f572e4632a0841876e9333e3ed0782f843d).

optax/_src/alias.py
```diff
@@ -431,6 +431,29 @@ def noisy_sgd(
   )
 
 
+def online_newton_step(learning_rate: ScalarOrSchedule,
+                       eps: float) -> base.GradientTransformation:
+  # pylint: disable=line-too-long
```
mkunesch, line 436: Actually, I think as long as the link can fit into a line we could just introduce line breaks in the references. At least in the new theme e.g. for [rmsprop here](https://optax.readthedocs.io/en/latest/api.html#optax.rmsprop) I don't think there is a benefit to disabling and having it in one line.

optax/_src/alias.py
```diff
@@ -431,6 +431,29 @@ def noisy_sgd(
...
+    Hazan, E., Agarwal, A. and Kale, S., 2007. Logarithmic regret algorithms for online convex optimization. Machine Learning, 69(2-3), pp.169-192 : https://link.springer.com/content/pdf/10.1007/s10994-007-5016-8.pdf
+
+  Args:
+    learning_rate: this is a fixed global scaling factor.
+    initial_accumulator_value: initialisation for the accumulator.
+    eps: A floating point value to avoid zero denominator.
+
+  Returns:
+    the corresponding `GradientTransformation`.
+  """
```
mkunesch, line 450: (If we do end up keeping the long line we should enable the warning again immediately after the docstring.)

optax/_src/alias.py
```diff
@@ -431,6 +431,29 @@ def noisy_sgd(
   )
 
 
+def online_newton_step(learning_rate: ScalarOrSchedule,
+                       eps: float) -> base.GradientTransformation:
+  # pylint: disable=line-too-long
+  """Online Newton Step optimizer.
+  (see the description of the ONS in Fig. 2 p. 176 of the reference below).
```
mkunesch, line 438: According to the google python style guide there needs to be a blank line following the summary line.

optax/_src/alias_test.py
```diff
@@ -142,6 +142,25 @@ def test_explicit_dtype(self, dtype):
     adam_state, _, _ = tx.init(jnp.array([0.0, 0.0]))
     self.assertEqual(expected_dtype, adam_state.mu.dtype)
 
+  def test_online_newton_step_decreases_loss(self):
+    """We check that the online newton step update decrease a loss function.
```
mkunesch, line 146: typo: ... newton step update decreases a

But actually, the "We" isn't typical for test docstrings. I'd suggest:

```
"""Checks that the online newton step update decreases the loss."""
```

optax/_src/transform.py
```diff
@@ -953,6 +953,57 @@ def update_fn(updates, state, params=None):
   return base.GradientTransformation(init_fn, update_fn)
 
 
+class ScaleByOnlineNewtonStepState(NamedTuple):
+  """State holding the inverse of the sum of gradient outer products to date."""
+  hessian_inv: base.Updates
+
+
+def sherman_morrison(a_inv, u):
+  """Sherman Morrison formula to compute the inverse of the sum of an invertible
```
mkunesch, line 962: The first line of the docstring should be a <=80 character summary according to https://google.github.io/styleguide/pyguide.html#381-docstrings and followed by a blank line.

In this case you could just keep Sherman Morrison formula as summary and move the rest to the detailed description.

Please also add an Args and Returns section as the internal linter won't allows us to merge non-trivial docstrings (i.e. docstrings that don't just say "Returns X given Y.") without Args and Returns sections.

optax/_src/transform.py
```diff
@@ -953,6 +953,57 @@ def update_fn(updates, state, params=None):
   return base.GradientTransformation(init_fn, update_fn)
 
 
+class ScaleByOnlineNewtonStepState(NamedTuple):
+  """State holding the inverse of the sum of gradient outer products to date."""
+  hessian_inv: base.Updates
+
+
+def sherman_morrison(a_inv, u):
```
mkunesch, line 961: Could you add type annotations to this function please? I assume, here this is just `chex.Array` for all inputs and the return type?

(Not all of optax was typed in the past but we'd like all new contributions to use type annotations and are slowly adding them to the old code too.)

optax/_src/transform.py
```diff
@@ -953,6 +953,57 @@ def update_fn(updates, state, params=None):
...
+  """Sherman Morrison formula to compute the inverse of the sum of an invertible
+   matrix and the outer product of two vectors.
+
+   The formula is used in the "Online Newton Step" gradient update method.
+   """
+  den = 1.0 + (u.T @ a_inv @ u)
```
mkunesch, line 967: Nit: there is enough space to write out den to denominator so I'd do this to make the code easier to read.

optax/_src/transform.py
```diff
@@ -953,6 +953,57 @@ def update_fn(updates, state, params=None):
...
+
+  References:
+    Hazan, E., Agarwal, A. and Kale, S., 2007. Logarithmic regret algorithms for online convex optimization. Machine Learning, 69(2-3), pp.169-192 : https://link.springer.com/content/pdf/10.1007/s10994-007-5016-8.pdf
+
+  Args:
+    eps: A floating point value to avoid zero denominator.
+
+  Returns:
+    An (init_fn, update_fn) tuple.
+  """
```
mkunesch, line 985: (As above, we could actually break the lines in the references I think but in any case we should re-enable the line-too-long warning again here otherwise the rest of the function doesn't get checked.)

optax/_src/transform.py
```diff
@@ -953,6 +953,57 @@ def update_fn(updates, state, params=None):
...
+   """
+  den = 1.0 + (u.T @ a_inv @ u)
+  a_inv -= a_inv @ jnp.outer(u, u) @ a_inv / den
+  return a_inv
+
+
+def scale_by_online_newton_step(eps: float) -> base.GradientTransformation:
+  # pylint: disable=line-too-long
+  """Rescale the updates by multiplying them by the inverse of a hessian approximation
+  (see the description of the ONS in Fig. 2 p. 176 of the reference below).
```
mkunesch, line 975: As above, the docstrings should have a <= 80 character summary line followed by a blank line.

optax/_src/transform.py
```diff
@@ -953,6 +953,57 @@ def update_fn(updates, state, params=None):
...
+  def init_fn(params):
+    hessian_inv = jax.tree_map(
+        lambda t: jnp.eye(len(t.flatten()), dtype=t.dtype) / eps, params
+    )
+    return ScaleByOnlineNewtonStepState(hessian_inv=hessian_inv)
+
+  def update_fn(updates, state, params=None):
+    del params
+    flattened_updates = jax.tree_map(lambda x: x.flatten(), updates)
+    hessian_inv = jax.tree_multimap(
```
mkunesch, line 996: `Tree_multimap` has been deprecated by jax. You can use `tree_map` as a drop-in replacement.

optax/_src/transform.py
```diff
@@ -953,6 +953,57 @@ def update_fn(updates, state, params=None):
...
+        sherman_morrison, state.hessian_inv, flattened_updates
+    )
+    flattened_updates = jax.tree_multimap(lambda hinv, g: hinv @ g, hessian_inv, flattened_updates)
```
mkunesch, line 999: Nit: I think `h_inv` would be easier to read.

optax/_src/transform_test.py
```diff
@@ -269,6 +270,83 @@ def test_add_noise_has_correct_variance_scaling(self):
...
 
+  def test_sherman_morrison(self):
+    rng = jax.random.PRNGKey(42)
+    rng, sub_rng = jax.random.split(rng)
+    a = jax.random.normal(sub_rng, (3, 3))
+
+    rng, sub_rng = jax.random.split(rng)
+    u = jax.random.normal(sub_rng, (3,))
+
+    ref = jnp.linalg.inv(a + jnp.outer(u, u))
```
mkunesch, line 281: Nit: I'd write out ref to `correct_inverse` to make the code and the errors that are raised if the test fails more readable.

optax/_src/transform_test.py
```diff
@@ -269,6 +270,83 @@ def test_add_noise_has_correct_variance_scaling(self):
...
+
+    grad = jax.grad(fun)(params, x)
+
+    state = step.init(params)
+    updates, state = step.update(grad, state)
+
+    u = (w @ x) * x
+
+    # check that step.update applied the sherman_morrison
+    ref = transform.sherman_morrison(jnp.eye(3) / eps, u) @ u
```
mkunesch, line 310: I think these three lines of the test can be removed for a unit test. The last part of the test is great: it checks the calculated gradients against the analytic calculation that we promise to the user. Whether or not sherman_morrison gets called is more of an implementation detail (indeed I think it could be a private function) and testing that isn't typical for a unit test.

optax/_src/transform_test.py
```diff
@@ -269,6 +270,83 @@ def test_add_noise_has_correct_variance_scaling(self):
...
+    x = jax.random.normal(sub_rng, (3, 2))
+
+    grad = jax.grad(fun)(params, x)
+
+    state = step.init(params)
+    updates, state = step.update(grad, state)
+
+    u = grad['weights'].flatten()  # (w * x).sum() * x
+
+    # check that step.update applied the sherman_morrison
```
mkunesch, line 340: (As above, I'd remove these three lines - the important part is the comparison agains the analytical results).

optax/_src/transform_test.py
```diff
@@ -269,6 +270,83 @@ def test_add_noise_has_correct_variance_scaling(self):
...
+    def fun(params, x):
+      return 0.5 * (params['weights'] @ x)**2
+
+    rng, sub_rng = jax.random.split(rng)
+    x = jax.random.normal(sub_rng, (3,))
+
+    grad = jax.grad(fun)(params, x)
+
+    state = step.init(params)
+    updates, state = step.update(grad, state)
```
mkunesch, line 305: Could you wrap `step.update` and `step.init` in `self.variant` and decorate the test with `@chex.all_variants` please? You can look at some of the tests at the top of the file for examples.

This ensures that the optimizer code can be jit compiled, executed on device etc.

mkunesch (changes requested):
Hi! Thanks a lot for making all the changes! 

I did a second pass now with more detailed comments (mostly formatting).

The checks currently fail - but I think that's due to `tree_multimap` being deprecated. You can just replace it by `tree_map` and it should work.

Thanks a lot again!

eserie:
Hi @mkunesch, thanks for all your comments. I tried to address them all, hope all is ok now!

mkunesch:
Hi!

Thanks a lot! Before merging I wanted to make sure I could use it myself, but had difficulty finding parameters that could find the minimum of a parabola.

Would it be possible to add the optimizer to the test `test_optimization` in `alias_test.py` that runs all the optimizers on a parabola and ensures they find the correct minimum? That way there would be examples of the parameters that work.

Other than that, the code looks good to me up to minor formatting edits and capitalization etc that I could fix with the internal linter during merging if that's ok with you! (But if you prefer I'm also happy to make file comments and you can fix them).

Thanks again!

eserie:
Hi!

I could add the `online_newton_step` optimizer to the test `test_optimization` in `alias_test.py`. However, I could not find parameters that fits for the two loss functions. So I addded a special condition in test code to adjust the parameters for one of the loss functions. I hope this is ok.

No problem for me if you want to fix the formatting issues you have seen (on my side the checks done in the test.sh all passed).

NB: to find the parameters, I played with this code in an interactive session:
```python
import jax
import jax.numpy as jnp
import optax
from matplotlib import pyplot as plt
from optax._src import alias
from optax._src import update
from optax._src import numerics

a = 1.0
b = 100.0

initial_params = jnp.array([0.0, 0.0])
final_params = jnp.array([a, a**2])


def fun(params):
    return numerics.abs_sq(a - params[0]) + b * numerics.abs_sq(
        params[1] - params[0] ** 2
    )

opt = alias.online_newton_step(5.0e-1, eps=1.)

params = initial_params
state = opt.init(params)


@jax.jit
def step_(state_params, i):
    state, params = state_params
    val, grad = jax.value_and_grad(fun)(params)
    updates, state = opt.update(grad, state)
    params = update.apply_updates(params, updates)
    return (state, params), (params)


state, params_history = jax.lax.scan(step_, (state, params), jnp.arange(10000))
plt.plot(jnp.linalg.norm(params_history - final_params[None, :], axis=1))
plt.title("$| w - w^* |$")
jnp.linalg.norm(params_history[-1] - final_params)
```
image

mkunesch:
Hi @eserie,

Thanks a lot for adding the optimizer to the tests and finding parameter combinations that work!

We have experimented more with the optimizer in the past month (trying it on various functions and some deep learning work loads such as MNIST) and we have concluded that this optimizer may not be a good fit for the core optax API at this moment.

Optax is currently focused on optimizers that can be substituted into most deep learning training loops (ideally with the default parameters), and we have found that when using ONS on DL problems, finding the right parameters can be tricky and it may run out of memory for typical DL networks.

From the paper and what you have written, ONS is at its best in online learning on time series and streaming data, so we would suggest publishing this optimizer as part of a repository that specializes in these applications. You can of course still use the optax machinery and make this work with optax, but we would suggest providing this as a third-party component that uses optax rather than integrating it into the core API. We would of course be happy to prominently post a link to it here in this PR (or a new issue) so that people can find it easily if they search for it in optax.

Sorry to have decided this towards the end of the code review. We will try to prevent this in the future by creating a series of problems every optimizer should solve with standard parameters before starting the review. Still, thank you very much for making all the changes and we hope that the comments are useful for you for your own open-source version!

Thanks a lot for filing the PR again and we hope you understand the decision!

PR #288, opened 2022-01-25
Fix Yogi optimizer

wdphy16:
When I was implementing #279, I noticed that the Yogi optimizer was incorrectly implemented. We should not take the square of `signed_sq` in `_update_moment`, and the coefficient to update `v` is also different from that in `_update_moment`.

After the fix, I need to change the learning rate to pass the test.


hbq1:
FYI We will only be able to merge this and other CLs after the ICML 2022 deadline (on this Friday). Sorry for the inconvenience!

Issue #281, opened 2022-01-16
add_noise scales by variance, not its root


Rupt:
To match its intended variance, [add_noise](https://github.com/deepmind/optax/blob/5fc395b5e49602e3c0027b976d1368755b20cd0b/optax/_src/transform.py#L725) should scale by its square root.
```diff
-        lambda g, n: g + variance.astype(g.dtype) * n,
+        lambda g, n: g + jnp.sqrt(variance.astype(g.dtype)) * n,
```

I am posting this as an issue rather than a pull request since this is a strongly breaking change which deserves some consideration. Thank you for maintaining and sharing optax!

mkunesch:
Hi @Rupt!

Thank you so much for reporting this! This is indeed a bug and should be fixed. Normally, I'd say we shouldn't quietly change behaviours but I think in this case it's fine (`add_noise` and `noisy_sgd` don't seem to be used very commonly and we are still in the 0.*.* version numbers).

Do you want to file a PR for this (+ a failing test) or shall I? (We'll release a new version after the fix)

Thanks a lot again for reporting!

Rupt:
Thanks @mkunesch for the feedback. This makes sense to me, and I would like to try the PR (with test) now.

mkunesch:
Amazing, thanks a lot! I'll review the code - we might even be able to include it with the version we want to release with the complex number support (#279) this week.

mkunesch:
This was fixed in #294 so I'll close this issue. Thanks @Rupt for spotting and fixing this bug!

PR #294, opened 2022-01-31
Change add_noise to match the target variance by scaling by its sqrt.

Rupt:
Resolve issue #281.

Add a test for this noise scaling which fails prior to this change.

### DIFF
--- a/optax/_src/transform.py
+++ b/optax/_src/transform.py
@@ -744,12 +744,13 @@ def add_noise(
     treedef = jax.tree_structure(updates)
     count_inc = numerics.safe_int32_increment(state.count)
     variance = eta / count_inc**gamma
+    std = jnp.sqrt(variance)
     all_keys = jax.random.split(state.rng_key, num=num_vars + 1)
     noise = jax.tree_multimap(
         lambda g, k: jax.random.normal(k, shape=g.shape, dtype=g.dtype),
         updates, jax.tree_unflatten(treedef, all_keys[1:]))
     updates = jax.tree_multimap(
-        lambda g, n: g + variance.astype(g.dtype) * n,
+        lambda g, n: g + std.astype(g.dtype) * n,
         updates, noise)
     return updates, AddNoiseState(count=count_inc, rng_key=all_keys[0])
 
--- a/optax/_src/transform.py
+++ b/optax/_src/transform.py
@@ -744,13 +744,13 @@ def add_noise(
     treedef = jax.tree_structure(updates)
     count_inc = numerics.safe_int32_increment(state.count)
     variance = eta / count_inc**gamma
-    std = jnp.sqrt(variance)
+    standard_deviation = jnp.sqrt(variance)
     all_keys = jax.random.split(state.rng_key, num=num_vars + 1)
     noise = jax.tree_multimap(
         lambda g, k: jax.random.normal(k, shape=g.shape, dtype=g.dtype),
         updates, jax.tree_unflatten(treedef, all_keys[1:]))
     updates = jax.tree_multimap(
-        lambda g, n: g + std.astype(g.dtype) * n,
+        lambda g, n: g + standard_deviation.astype(g.dtype) * n,
         updates, noise)
     return updates, AddNoiseState(count=count_inc, rng_key=all_keys[0])
 
--- a/optax/_src/transform_test.py
+++ b/optax/_src/transform_test.py
@@ -207,6 +207,35 @@ class TransformTest(parameterized.TestCase):
     centralized_inputs, _ = centralizer.update(inputs, None)
     chex.assert_tree_all_close(centralized_inputs, outputs)
 
+  @parameterized.parameters(
+    (0.01, 0.55, 123),
+    (0.3, 0.54, 234),
+    (1.0, 0.56, 314),
+  )
+  def test_add_noise(self, eta, gamma, seed):
+    # Prepare to compare noise with a rescaled unit-variance substitute.
+    noise = transform.add_noise(eta, gamma, seed)
+    noise_unit = transform.add_noise(1.0, 0.0, seed)
+
+    params = self.init_params
+    state = noise.init(params)
+    state_unit = noise_unit.init(params)
+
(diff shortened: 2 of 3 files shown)


Rupt:
Edited to avoid sqrt after type casting.

optax/_src/transform.py
```diff
@@ -717,12 +717,13 @@ def update_fn(updates, state, params=None):  # pylint: disable=missing-docstring
     treedef = jax.tree_structure(updates)
     count_inc = numerics.safe_int32_increment(state.count)
     variance = eta / count_inc**gamma
+    std = jnp.sqrt(variance)
```
mkunesch, line 720: Nit: std is a standard abbreviation so it's perfectly fine as is, but I'd personally prefer `standard_deviation`, also to be consistent with the use of `variance` over `var`.

optax/_src/transform_test.py
```diff
@@ -202,6 +202,35 @@ def test_centralize(self, inputs, outputs):
...
+    updates = jax.tree_map(jnp.zeros_like, self.per_step_updates)
+
+    for i in range(1, STEPS + 1):
+      updates_i, state = noise.update(updates, state)
+      updates_i_unit, state_unit = noise_unit.update(updates, state_unit)
+
+      scale = jnp.sqrt(eta / i**gamma)
+
+      def rescale(g):
+        return g * scale  # pylint:disable=cell-var-from-loop
```
mkunesch, line 229: I think this is simple enough to be written as lambda in the line below. That way you also wouldn't need to disable the pylint warning or did pylint complain about something else then?

optax/_src/transform_test.py
```diff
@@ -202,6 +202,35 @@ def test_centralize(self, inputs, outputs):
...
+    # Prepare to compare noise with a rescaled unit-variance substitute.
+    noise = transform.add_noise(eta, gamma, seed)
+    noise_unit = transform.add_noise(1.0, 0.0, seed)
+
+    params = self.init_params
+    state = noise.init(params)
+    state_unit = noise_unit.init(params)
+
+    # Check the noise itself by adding it to zeros.
+    updates = jax.tree_map(jnp.zeros_like, self.per_step_updates)
```
mkunesch, line 220: Nit: you could make this more self-contained by replacing `self.per_step_updates` with `params` directly (because that's ultimately the shape we want to match).

optax/_src/transform_test.py
```diff
@@ -202,6 +202,35 @@ def test_centralize(self, inputs, outputs):
     centralized_inputs, _ = centralizer.update(inputs, None)
     chex.assert_tree_all_close(centralized_inputs, outputs)
 
+  @parameterized.parameters(
+    (0.01, 0.55, 123),
+    (0.3, 0.54, 234),
+    (1.0, 0.56, 314),
+  )
+  def test_add_noise(self, eta, gamma, seed):
```
mkunesch, line 210: Could you provide more information on the behaviour being tested? Either by changing the name to e.g. `test_add_noise_has_correct_standard_deviation` or by adding a docstring with something similar (just so that there is more information if the test fails). Thanks a lot!

mkunesch (changes requested):
Looks great! Thank you so much for this bug fix and the associated test!
I just left a few very minor comments/nits.

optax/_src/transform.py
Rupt: Done.
(I chose std to match its use in stochastic_gradient_estimators, but totally agree with your argument)

optax/_src/transform_test.py
Rupt: Yes this is because pylint complains with the lambda form, and with this same warning. This style copies [L186](https://github.com/deepmind/optax/blob/a0d5ed5539a58cc0bfa1d5948c87d2aa3b693d92/optax/_src/transform_test.py#L186).

optax/_src/transform_test.py
Rupt: Nice! Done.

optax/_src/transform_test.py
Rupt: I've gone with `test_add_noise_has_correct_variance_scaling` since the test doesn't actually check the standard deviation, just the scale relative to eta=1, gamma=0.
Thanks for all the nice comments.

mkunesch (changes requested):
Thanks a lot for making the changes! I forgot to check for `chex.all_variants` last time so I have one more comment. Otherwise looks good to me!

optax/_src/transform_test.py
Rupt: Neat solution, done with the lambda.

optax/_src/transform_test.py
```diff
@@ -202,6 +202,35 @@ def test_centralize(self, inputs, outputs):
     centralized_inputs, _ = centralizer.update(inputs, None)
     chex.assert_tree_all_close(centralized_inputs, outputs)
 
+  @parameterized.parameters(
+    (0.01, 0.55, 123),
+    (0.3, 0.54, 234),
+    (1.0, 0.56, 314),
+  )
+  def test_add_noise_has_correct_variance_scaling(self, eta, gamma, seed):
```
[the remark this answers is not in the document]
Rupt: Added the variants.

Agreed there is nothing new the other tests can catch. Kwargs there don't work because the parameters are reused elsewhere (seed for the other transform, eta and gamma for scale), so I've just hard coded them as variables.

mkunesch (commented):
LGTM! Thanks a lot!

There is a test failure now in the (recently added) test for `noisy_sgd` in `alias_test.py` which uses `add_noise`. Do you have time to look into this? It's probably just a matter of changing the parameters so that with the slower decay of the noise it doesn't affect the test results but we should make sure this is indeed the reason.

Rupt:
> It's probably just a matter of changing the parameters
I've rebased and changed eta to match the previous initial variance. It survives the test with the different decay rate.
 
> but we should make sure this is indeed the reason.
The test passes now (and also passed if just removing the noisy_sgd line.

mkunesch (approved):
Great, thanks a lot for doing this!

PR #309, opened 2022-02-28
optax: Add Kullback-Leibler Divergence Loss

holounic:


optax/_src/loss.py
```diff
@@ -358,3 +358,26 @@ def loop_body(prev, x):
   per_seq_loss = -jnp.einsum('bn,bn->b', logalpha_phi_last, one_hot)
 
   return per_seq_loss
+
+
+def kldiv_loss(predictions: chex.Array,
+               targets: chex.Array):
+  """Computes the Kullback-Leibler divergence (relative entropy) loss.
+
+  Measures the information gain achieved if probability distribution P
```
mkunesch, line 367: Nit: since you don't refer to P and Q elsewhere I don't think it's necessary to introduce these names (you could write it in terms of a different distribution and a reference distribution). But as I said, this is a nit - it's perfectly clear what you mean.

optax/_src/loss.py
```diff
@@ -358,3 +358,26 @@ def loop_body(prev, x):
...
+  Args:
+    predictions: Probabilities of predicted distribution with shape [..., dim].
+    targets: Probabilities of target distribution with shape [..., dim].
+
+  Returns:
+    Kullback-Leibler divergence of predicted distribution from target
+    distribution with shape [...].
+  """
+  chex.assert_type([predictions, targets], float)
+  log_p_div_q = jnp.log(jnp.divide(targets, predictions))
```
mkunesch, line 382: Nit: is there a reason for why you use the long forms `jnp.divide` and `jnp.multiply`? I personally find `/` and `*` easier to read.

optax/_src/loss.py
```diff
@@ -358,3 +358,26 @@ def loop_body(prev, x):
...
+  return jnp.sum(jnp.multiply(targets, log_p_div_q), axis=-1)
```
mkunesch, line 383: I think we might need to handle numerical underflow and division by zero somehow ... afaik pytorch implements this as operating on `log_targets` and `log_predictions` and tensorflow does some clipping potentially with a small epsilon? What do you think? If you agree, let's discuss what the better approach is!

(Just so that I don't forget: if it is important we should add tests for underflow and div by zero.)

optax/_src/loss_test.py
```diff
@@ -369,5 +369,28 @@ def test_repeat_with_one_to_one_alignment(self):
           jnp.array(expected_loss), per_seq_loss[n], rtol=self._rtol)
 
 
+class KLDivLossTest(parameterized.TestCase):
```
mkunesch, line 372: Nice tests - thanks a lot!

mkunesch (changes requested):
Hi! Thanks a lot for the contribution and sorry for the slow review!
I have a few small nits and one comment on handling division by zero and underflow. Let me know what you think! 

Thanks a lot again for the PR!

optax/_src/loss.py
holounic: I am not sure that I am right in this but I prefer to maintain a functional-style code when using jax.

optax/_src/loss.py
```diff
@@ -358,3 +358,26 @@ def loop_body(prev, x):
   per_seq_loss = -jnp.einsum('bn,bn->b', logalpha_phi_last, one_hot)
 
   return per_seq_loss
+
+
+def kldiv_loss(predictions: chex.Array,
```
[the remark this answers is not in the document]
holounic: Done!

holounic:
Hi @mkunesch, thanks a lot for the review!
I have made some improvements in the code. Please let me know if I need to fix anything else :)

optax/_src/loss.py
```diff
@@ -358,3 +358,33 @@ def loop_body(prev, x):
   per_seq_loss = -jnp.einsum('bn,bn->b', logalpha_phi_last, one_hot)
 
   return per_seq_loss
+
+
+def kl_div_loss(predictions: chex.Array,
```
mkunesch, line 363: These are now log-predictions aren't they? If my understanding of this is correct we should reflect this in the name and call it `log_predictions`. That will also make it easier to make this clear in the doc string.

Update: I know this would cause confusion with `log_targets` but I think we should remove the flag (see next comment).

optax/_src/loss.py
```diff
@@ -358,3 +358,33 @@ def loop_body(prev, x):
...
+                targets: chex.Array,
+                log_targets: bool = False) -> chex.Array:
```
mkunesch, line 365: I know this is how pytorch does it but since we only have this one function (as opposed to pytorch) I think we should avoid the boolean flag by having two functions: `kl_div_loss` and `kl_div_loss_with_log_targets`.

Aside from the usual arguments against boolean flags, there are two more arguments against having it as a flag in this case in my opinion: the boolean flag changes the meaning of the other parameter (from targets to log_targets) without changing the name which makes the resulting code harder to read and debug. Furthermore, reading from left to right, the meaning of targets will become clear only later at the end of the line or only in the line below.

That way we can also use the name log_predictions (and targets/log_targets) as names for the function inputs without causing confusion with the boolean flag.

optax/_src/loss.py
```diff
@@ -358,3 +358,33 @@ def loop_body(prev, x):
...
+    Expected to be strictly positive.
+    log_targets: Specifies whether targets are given in the log-space.
+
+  Returns:
+    Kullback-Leibler divergence of predicted distribution from target
+    distribution with shape [...].
+  """
+  chex.assert_type([predictions, targets], float)
+  if log_targets:
+    loss = jnp.multiply(jnp.exp(targets), jnp.subtract(targets, predictions))
```
mkunesch, line 387: I know this is a matter of style and I originally called this a _nit_, but we have since decided that we would like to enforce the `+, -, *, /`-style for optax. For large parts of optax, the expressions are too complex mathematically to be readable with `jnp.multiply` etc in our opinion and we'd like to keep the style of mathematical expressions consistent across the codebase.

Could you change your function to use the `*`, `-` operators? Thanks a lot!

optax/_src/loss.py
```diff
@@ -358,3 +358,33 @@ def loop_body(prev, x):
...
+
+  Measures the information gain achieved if target probability distribution
+  would be used instead of predicted probability distribution.
+
+  References:
+    [Kullback, Leibler, 1951](https://www.jstor.org/stable/2236703)
+
+  Args:
+    predictions: Probabilities of predicted distribution with shape [..., dim].
+    Expected to be in the log-space to avoid underflow.
```
mkunesch, line 376: In theory pylint should complain here ... there should be 2 white spaces before "Expected" so that it's indented in the section of "predictions". Once you have synced we can run the testing workflow again which should have flagged this (the workflow broke shortly before you committed but we have since fixed it).

mkunesch (changes requested):
Hi @holounic!

Thanks a lot for making the changes! I think your choice of asking the user to provide log predictions is the right way to go. In most cases the users will get the log predictions directly from a neural network anyway.

I've added three comments on the new changes. We have also fixed the github checks again since your last commit so once you sync the checks should run again.

Thanks a lot!

optax/_src/loss.py
holounic: Done!

optax/_src/loss.py
```diff
@@ -454,3 +454,49 @@ def ctc_loss(logits: chex.Array,
       logits, logit_paddings, labels, label_paddings,
       blank_id=blank_id, log_epsilon=log_epsilon)
   return per_seq_loss
+
+
+def kl_div_loss(log_predictions: chex.Array,
```
mtthss, line 459: Just `kl_divergence`

optax/_src/loss.py
```diff
@@ -454,3 +454,49 @@ def ctc_loss(logits: chex.Array,
...
+  Returns:
+    Kullback-Leibler divergence of predicted distribution from target
+    distribution with shape [...].
+  """
+  chex.assert_type([log_predictions, targets], float)
+  loss = targets * (jnp.log(targets) - log_predictions)
+  return jnp.sum(loss, axis=-1)
+
+
+def kl_div_loss_with_log_targets(log_predictions: chex.Array,
```
mtthss, line 484: kl_divergence_with_log_targets

mkunesch (commented):
This looks great to me - thanks for making all the changes! (I agree with the name changes suggested by @mtthss but otherwise LGTM)

Thanks a lot!

optax/_src/loss.py
holounic: done

optax/_src/loss.py
holounic: done

holounic:
Hi @mkunesch, thanks you for the review!
I have renamed loss functions and the following tests as @mtthss recommended. Please take a look at the PR again :)

optax/_src/loss_test.py
```diff
@@ -416,5 +416,56 @@ def test_repeat_with_one_to_one_alignment(self):
...
+                            [-1.6094, -1.6094, -1.6094, -2.3026, -1.8971, -1.8971]])
+    self.qs = np.array([[0.2, 0.2, 0.2, 0.1, 0.15, 0.15],
+                       [0.05, 0.03, 0.02, 0.3, 0.5, 0.1]])
+    # computed kullback-leibler divergence of P from Q.
+    self.exp = np.array([0.8875625, 0.7187435584901326])
+
+  @chex.all_variants
+  def test_scalar(self):
+    np.testing.assert_allclose(
+        self.variant(loss.kl_div_loss)(self.log_ps[0], self.qs[0]),
```
mkunesch, line 433: This (and the other function names) should be updated for the test to pass.

optax/_src/loss.py
```diff
@@ -454,3 +454,49 @@ def ctc_loss(logits: chex.Array,
       logits, logit_paddings, labels, label_paddings,
       blank_id=blank_id, log_epsilon=log_epsilon)
   return per_seq_loss
+
+
+def kl_divergence(log_predictions: chex.Array,
+                targets: chex.Array) -> chex.Array:
```
mkunesch, line 460: The second line here isn't correctly aligned now (but I can also fix that during merging).

mkunesch (changes requested):
Thanks for changing the function names! I think you forgot to reflect the name change in the tests. Could you change it there too for the tests to pass?

Thanks a lot!

optax/_src/loss_test.py
holounic: done!

optax/_src/loss.py
holounic: done!

mkunesch (approved):
Amazing, thanks a lot! Looks great to me!

It may take a few days to merge as we need to merge #360 first in order to fix the pipeline but I'll take it from here!

PR #318, opened 2022-03-19
Fix docs rendering of loss functions.

### LINKED ISSUE #317: docs: loss functions not rendering properly
Compare https://optax.readthedocs.io/en/latest/api.html#losses with https://optax.readthedocs.io/en/stable/api.html#losses

api.rst has a bug in the autofunction syntax for `ctc_loss`, so docs for ctc_loss and subsequent loss functions aren't rendered.
https://github.com/deepmind/optax/blob/252d152660300fc7fe22d214c5adbe75ffab0c4a/docs/api.rst#L445-L456

grahamgower:
Closes #317.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -447,7 +447,7 @@ Losses
 
 .. autofunction:: cosine_distance
 .. autofunction:: cosine_similarity
--- autofunction:: ctc_loss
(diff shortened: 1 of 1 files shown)


mkunesch (approved):
Thanks a lot!

PR #321, opened 2022-03-28
Enhance `ctc_loss` to return forward probs.

yotarok:
Continuing from #293, we found some use cases where interim alpha probabilities are useful.
This PR renames `ctc_loss` to `ctc_loss_with_foward_probs` and make `ctc_loss` to be an alias to `ctc_loss_with_forward_probs` with omitting auxiliary outputs.

PR #331, opened 2022-04-11
Rename argument names for CTC loss functions.

yotarok:
This PR is fixing some issues mentioned in the follow-up discussions on #321 

The docstring is now confirmed to be correctly rendered to HTMLs.
Since the arguments are renamed, the old code with invocating ctc_loss with keyword arguments will break after merging this change.


hbq1 (approved):
Neat PR, thanks a lot!

PR #351, opened 2022-06-02
[Docs] softmax_cross_entropy_with_integer_label

lkhphuc:
Document for the new loss function is missing from Readthedocs.


mkunesch (approved):
Thanks a lot for spotting this! (and sorry for the slow review)

Issue #250, opened 2021-12-01
Missing type hints


mkunesch:
Some parts of optax do not yet have type hints. It would be great to add them! Files where I've found missing type hints include:

- [x] [utils.py](https://github.com/deepmind/optax/blob/master/optax/_src/utils.py)
- [x] [clipping.py](https://github.com/deepmind/optax/blob/master/optax/_src/clipping.py)
- [x] [linear_algebra.py](https://github.com/deepmind/optax/blob/master/optax/_src/linear_algebra.py)

but this list is not exhaustive. I'm marking this as "good first issue" - PRs that add missing type hints are greatly appreciated!

esraa-saleh:
I would love to start working on this! Has someone been assigned already for all tasks?

mkunesch:
Hi, that would be great! Nobody has been assigned yet so your contribution would be very welcome.
Thanks a lot and let me know if you have any questions!

esraa-saleh:
Great! I'll start working on this then, and come back with questions if any come up :)

yulonglin:
I'm adding another list of files with missing type hints. I'll work on them and add more files along the way :)
- [ ] [base.py](https://github.com/deepmind/optax/blob/master/optax/_src/base.py)
- [ ] [base_test.py](https://github.com/deepmind/optax/blob/master/optax/_src/base_test.py)

Warlord-K:
> I'm adding another list of files with missing type hints. I'll work on them and add more files along the way :)
> 
> * [ ]  [base.py](https://github.com/deepmind/optax/blob/master/optax/_src/base.py)
> * [ ]  [base_test.py](https://github.com/deepmind/optax/blob/master/optax/_src/base_test.py)

I opened these and found that these files have type hints. So, I think this issue can be closed. Thanks

hbq1:
Update: https://github.com/deepmind/optax/pull/413 has added missing type hints to linear_algebra.py.

hbq1:
@mkunesch should we close this issue now?

mkunesch:
Ah yes, thanks - this is done so let's close it for now - anyone who finds missing type hints is still welcome to file a PR to fix them of course!

PR #367, opened 2022-06-28
Add types to utils.py

atgctg:
I added types to `utils.py`. Please review!

In case I add more, should I do them under separate PRs or under one big one?

(#250)

### DIFF
--- a/optax/_src/utils.py
+++ b/optax/_src/utils.py
@@ -14,6 +14,8 @@
 # ==============================================================================
 """Utility functions for testing."""
 
+from typing import Optional, Tuple, List, Sequence
+
 import chex
 import jax
 import jax.numpy as jnp
@@ -23,20 +25,19 @@ from optax._src import linear_algebra
 from optax._src import numerics
 
 
-def tile_second_to_last_dim(a):
+def tile_second_to_last_dim(a: chex.Array) -> chex.Array:
   ones = jnp.ones_like(a)
   a = jnp.expand_dims(a, axis=-1)
   return jnp.expand_dims(ones, axis=-2) * a
 
 
-def canonicalize_dtype(dtype):
+def canonicalize_dtype(dtype: Optional[jnp.dtype]) -> Optional[jnp.dtype]:
   """Canonicalise a dtype, skip if None."""
   if dtype is not None:
     return jax.dtypes.canonicalize_dtype(dtype)
   return dtype
 
-
-def cast_tree(tree, dtype):
+def cast_tree(tree: chex.ArrayTree, dtype: Optional[jnp.dtype]) -> chex.ArrayTree:
   """Cast tree to given dtype, skip if None."""
   if dtype is not None:
     return jax.tree_map(lambda t: t.astype(dtype), tree)
@@ -44,7 +45,7 @@ def cast_tree(tree, dtype):
     return tree
 
 
-def set_diags(a, new_diags):
+def set_diags(a: chex.Array, new_diags: chex.Array):
   """Set the diagonals of every DxD matrix in an input of shape NxDxD.
 
   Args:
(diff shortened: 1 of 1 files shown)


optax/_src/utils.py
```diff
@@ -23,28 +25,29 @@
...
 
-def tile_second_to_last_dim(a):
+def tile_second_to_last_dim(a: chex.Array) -> chex.Array:
   ones = jnp.ones_like(a)
   a = jnp.expand_dims(a, axis=-1)
   return jnp.expand_dims(ones, axis=-2) * a
 
 
-def canonicalize_dtype(dtype):
+def canonicalize_dtype(dtype: Optional[jnp.dtype]) -> Optional[jnp.dtype]:
```
mkunesch, line 34: I think we should make this `chex.ArrayDType` - I'm not sure whether `jnp.dtype` works if a user calls the function with e.g. just float. Also, that way we use chex wherever possible (consistent with `chex.Array`).

optax/_src/utils.py
```diff
@@ -23,28 +25,29 @@
...
   """Canonicalise a dtype, skip if None."""
   if dtype is not None:
     return jax.dtypes.canonicalize_dtype(dtype)
   return dtype
 
 
-def cast_tree(tree, dtype):
+def cast_tree(tree: chex.ArrayTree,
+              dtype: Optional[jnp.dtype]) -> chex.ArrayTree:
```
mkunesch, line 42: (as above, let's use `chex.ArrayDType`)

optax/_src/utils.py
```diff
@@ -67,52 +70,58 @@ def set_diags(a, new_diags):
...
+  def __init__(self, loc: chex.Array, log_scale: chex.Array):
     self._log_scale = log_scale
     self._scale = jnp.exp(log_scale)
     self._mean = loc
     self._param_shape = jax.lax.broadcast_shapes(
         self._mean.shape, self._scale.shape)
 
-  def sample(self, shape, seed):
+  def sample(self, shape: Sequence[int],
+             seed: jax.random.PRNGKey) -> chex.Array:
```
mkunesch, line 84: Let's use `chex.PRNGKey` for consistency with using chex for arrays.

optax/_src/utils.py
```diff
@@ -67,52 +70,58 @@ def set_diags(a, new_diags):
...
     return jnp.sum(log_prob, axis=sum_axis)
 
   @property
-  def log_scale(self):
+  def log_scale(self) -> chex.Array:
     return self._log_scale
 
   @property
-  def params(self):
+  def params(self) -> List[chex.Array]:
```
mkunesch, line 102: I think we should prefer the abstract type Sequence in this case - just so that we don't make any promises where they aren't strictly necessary. We might want to return a tuple or named tuple in the future.

mkunesch (changes requested):
Thank you very much for the contribution! This looks great!

I've just added minor comments on using chex wherever possible and an abstract return type.

optax/_src/utils.py
```diff
@@ -67,52 +70,58 @@ def set_diags(a, new_diags):
...
 
 
+def multi_normal(loc: chex.Array,
+                 log_scale: chex.Array) -> MultiNormalDiagFromLogScale:
+  return MultiNormalDiagFromLogScale(loc=loc, log_scale=log_scale)
+
+
 @jax.custom_vjp
-def _scale_gradient(inputs: chex.Array, scale: float)  -> chex.ArrayTree:
+def _scale_gradient(inputs: chex.Array, scale: float) -> chex.ArrayTree:
```
mkunesch, line 112: Ah sorry, I think these inputs should be `chex.ArrayTree` as the gradients could be a nested tree of arrays.

mkunesch (changes requested):
(thanks for the pointing out `chex.Array` vs `chex.ArrayTree`. I just checked this and there is one function where it should be `ArrayTree` instead of `Array`)

mtthss:
There is still one change requested by @mkunesch, 
@atgctg could you address it so we get this submitted?

mtthss:
This branch has conflicts that must be resolved

atgctg:
Resolved it

mkunesch (approved):
Great, thanks a lot! Looks great to me!

mkunesch:
Hi,

just to give a brief update: the delay in merging is due to internal tests having raised a bug: the + operator in line 85 does not work with the type annotation `Sequence[int]`. I'd propose to just cast shape to `tuple` in line 85. If that's okay with you I can add this small change in during merging (it would still be shown as your commits of course).

atgctg:
Hi, sorry for not catching that. 

Please go ahead with your fix and thanks for letting me know!

PR #382, opened 2022-07-25
Add amsgrad optimizer

merajhashemi:
Hi,
This pr implements Amsgrad—an extension to Adam that improves its convergence properties.


merajhashemi:
Hi @mkunesch, any chance you could review this? Thanks.

mtthss:
@merajhashemi thanks for the PR!
It fails the CI tests (see above) could address the failures?

merajhashemi:
@mtthss Done! Could you run the CI again?

optax/_src/alias.py
```diff
@@ -189,6 +189,7 @@ def adam(
     b2: float = 0.999,
     eps: float = 1e-8,
     eps_root: float = 0.0,
+    amsgrad: bool = False,
```
mkunesch, line 192: Let's introduce a separate alias `amsgrad` for this rather than adding a flag to Adam. My thinking is that there will be other improvements/modifications to Adam in the future and we should avoid an accumulation of options in the simple adam setup. Furthermore, I think it would make it easier for user to find `amsgrad` in optax and more obvious in the code that amsgrad is being used.

optax/_src/transform.py
```diff
@@ -344,6 +344,66 @@ def update_fn(updates, state, params=None):
...
+      `None` then the `dtype is inferred from `params` and `updates`.
+
+  Returns:
+    An (init_fn, update_fn) tuple.
+  """
+
+  mu_dtype = utils.canonicalize_dtype(mu_dtype)
+
+  def init_fn(params):
+    mu = jax.tree_util.tree_map(  # First moment
```
mkunesch, line 383: Nit: Do you think it would be possible / make sense to reuse the adam function for parts of this calculation? I think this might make the difference to Adam more obvious to a reader and improvements/fixes to `scale_by_adam` would automatically carry over.

e.g.
```
adam_state = scale_by_adam(b1, b2, eps, eps_root, mu_dtype).init(params)
nu_max = jax.tree_util.tree_map(jnp.zeros_like, params)
return ScaleByAmsgradState(nu_max=nu_max, **adam_state._asdict())
```

But I'll leave this up to you since it's definitely a matter of taste!

optax/_src/transform.py
```diff
@@ -344,6 +344,66 @@ def update_fn(updates, state, params=None):
...
+        lambda t: jnp.zeros_like(t, dtype=mu_dtype), params)
+    nu = jax.tree_util.tree_map(jnp.zeros_like, params)  # Second moment
+    nu_max = jax.tree_util.tree_map(jnp.zeros_like, params)
+    return ScaleByAmsgradState(count=jnp.zeros([], jnp.int32), mu=mu, nu=nu,
+                               nu_max=nu_max)
+
+  def update_fn(updates, state, params=None):
+    del params
+    mu = update_moment(updates, state.mu, b1, 1)
```
mkunesch, line 392: Nit: As above, it might be possible to use `scale_by_adam` for parts of this and just set `nu_max` and the `updates` with the special amsgrad code. I personally wouldn't worry about the duplicate calculation of the `updates` since the unused code would be removed by `jax.jit`.

But as above, this is definitely a matter of taste and the advantage of writing it out in full is that it then corresponds to exactly to how it's written in the paper. So I'll leave it up to what you prefer!

optax/_src/transform.py
```diff
@@ -344,6 +344,66 @@ def update_fn(updates, state, params=None):
...
 
+class ScaleByAmsgradState(NamedTuple):
+  """State for the AMSGrad algorithm."""
+  count: chex.Array  # shape=(), dtype=jnp.int32.
+  mu: base.Updates
+  nu: base.Updates
+  nu_max: base.Updates
+
+
+def scale_by_amsgrad(
```
mkunesch, line 355: Nice! I agree with keeping this separate from `scale_by_adam` rather than introducing a flat there.

optax/_src/transform.py
```diff
@@ -344,6 +344,66 @@ def update_fn(updates, state, params=None):
...
+    b1: float = 0.9,
+    b2: float = 0.999,
+    eps: float = 1e-8,
+    eps_root: float = 0.0,
+    mu_dtype: Optional[Any] = None,
+) -> base.GradientTransformation:
+  """Rescale updates according to the AMSGrad algorithm.
+
+  References:
+    [Reddi et al, 2018](https://openreview.net/forum?id=ryQu7f-RZ)
```
mkunesch, line 365: Nice! Thanks for adding the reference!

mkunesch (changes requested):

mkunesch (changes requested):
(sorry, I pressed send to early)

Thank you so much for this contribution! This is excellent!

I've just added very minor comments. 

Could you add a test for this PR? The easiest thing might be to just add `amsgrad` to the optimizer list in `alias_test.py` so that it gets tested on a parabola. It might also be nice to test the `nu_max` behavior explicitly in `transform_test.py` but the parabola test is the more important one.

Also, just to say that we will only be able to merge this after the ICLR deadline, but we can approve it before then so that we can merge immediately on the 29th of September.

Thanks a lot again for this excellent contribution!

hbq1:
Hi @merajhashemi, thanks a lot for your contribution! Any change you could address the comments made by @mkunesch, so we could merge this PR in the next version of Optax? Thanks!

optax/_src/transform.py
mkunesch: Actually, let's leave this as is - it's easier to read and doesn't require jit to not do unnecessary calculations.

mkunesch:
Actually, the changes seem quite minimal (name change + adding it to the test). I'd be happy to approve and make these changes upon merging if that's okay with you @merajhashemi?

merajhashemi:
Hi @hbq1 @mkunesch

Thanks for the comments.
The reason I kept amsgrad in Adam was to keep it similar to other libraries; both pytorch and tensorflow have amsgrad as an argument inside their Adam optimizer. Still, I could split it out and add the tests on the weekend. How does that sound?

mkunesch:
Hi! Ah, thanks a lot for that context - that makes a lot of sense. 

I'm still leaning towards splitting as we have generally tried to avoid boolean flags in optax for a while now and don't mirror pytorch and tensorflow in other optimizers too.

@hbq1 : was your 👍  a vote for leaving it as an argument or for splitting it?

Thanks a lot!

mkunesch (approved):
Looks great to me! Thanks a lot for splitting the optimizer from adam.

(I'll make some very minor formatting edits and fix the conflicts with master as I merge the PR)

Thanks a lot for the contribution again!

mkunesch:
(sorry for the 3 commits to your branch - I had to merge master before importing the PR and I messed up the merge in the github editor)

PR #385, opened 2022-07-31
Add NovoGrad optimizer

DT6A:
NovoGrad optimizer was presented in this paper: https://arxiv.org/pdf/1905.11286.pdf
It was used to train the ASR model named Jasper: https://arxiv.org/pdf/1904.03288.pdf


DT6A:
> There are a couple of conflicts, could you sync and merge so we get this submitted?

Hello! Yes, I will try to fix as soon as possible

docs/api.rst
```diff
@@ -149,6 +149,7 @@ Gradient Transforms
     scale_by_adam
     scale_by_belief
     scale_by_factored_rms
+    scale_by_novograd
```
hbq1, line 152: Could you also add `novograd` to the Common Optimizers section?

optax/_src/transform.py
```diff
@@ -952,6 +952,94 @@ def update_fn(updates, state, params=None):
...
+    b1: decay rate for the exponentially weighted average of grads.
+    b2: decay rate for the exponentially weighted average of squared grads.
+    eps: term added to the denominator to improve numerical stability.
+    eps_root: term added to the denominator inside the square-root to improve
+      numerical stability when backpropagating gradients through the rescaling.
+    weight_decay: a scalar weight decay rate.
+    mu_dtype: optional `dtype` to be used for the first order accumulator; if
+      `None` then the `dtype is inferred from `params` and `updates`.
+  Returns:
+    An (init_fn, update_fn) tuple.
```
hbq1, line 982: "A `GradientTransformation` object."

hbq1 (approved):

hbq1:
Thanks a lot for the contribution!

hbq1:
@DT6A could you please add a test for the new optimizer?

DT6A:
@hbq1 is there any reference how the test should be done?

mkunesch:
Hi! We are in the process of writing instructions on this at the moment - in your case it should be enough to just add the new optimizer to the list of optimizers under test in `alias_test.py`. Let me know if you have any questions on this!

DT6A:
@mkunesch Thanks, added it

mkunesch:
(If there are details of the optimizer that would be important to test in isolation it would be good to add a unit test for them, but from a quick look at the documentation and code nothing stood out to me)

hbq1 (approved):

hbq1:
Merged! Thanks again for this great contribution 🎉

Issue #403, opened 2022-08-28
Add Hinge Loss


tanaymeh:
Hi, I would like to add Hinge loss to Optax.

I am hoping to contribute more extensively to optax and hence starting with this loss function addition. 

Can I go ahead with this?

Hinge loss: [Wikipedia](https://en.wikipedia.org/wiki/Hinge_loss), [Sklearn](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.hinge_loss.html)

mtthss:
Sounds great! Looking forward to it

heytanay:
@mkunesch Now, that Binary Hinge Loss is merged, can I open a new PR for Multi-class Hinge Loss?

mkunesch:
Totally! Which method exactly would you like to implement? Crammer-Singer similar to sklearn? If possible, it would be great if you could link a reference in the docstring or ideally even write out the equations in the docstring similar to how it's done in Adam:
https://github.com/deepmind/optax/blob/5f0501bac8015a48fc6cf78a6487a714df2b6553/optax/_src/alias.py#L216-L225

Thanks a lot for this and let me know if you have any questions or if you'd like to discuss the implementation!

heytanay:
@mkunesch I think the Crammer-Singer method would be a good option as it will closely resemble the sci-kit-learn's implementation. I will be opening a PR shortly.

PR #409, opened 2022-08-31
Add Hinge Loss

heytanay:
Hi, this PR intends to add Hinge Loss to Optax as referenced in #403.


mkunesch (commented):
Hi!

Thanks a lot for working on this! I know this is still in draft mode, but I thought it would be useful to provide some high-level comments in the meantime:

Please provide unit tests for the function that are wrapped in `chex.all_variants` and uses `self.variant` to run the loss with the different jax transforms (`jax.jit` etc.). You can take a look at the other tests to see how that's done.

The list comprehensions with if statements don't work with `jax.jit`. Please write them in terms of batch operations using jax functions and operators.

Thanks a lot again!

heytanay:
Hi @mkunesch, a quick question; Should I make 2 different functions for Hinge Loss (one for Binary and one for Multiclass) or coalesce them into one function?

mkunesch:
Having taken a look at sklearn the execution paths look quite different so I think I'd separate them into `hinge_loss` and `multiclass_hinge_loss` but let's merge this binary version first and discuss the multiclass version in a separate PR.

optax/_src/loss.py
```diff
@@ -503,3 +503,19 @@ def kl_divergence_with_log_targets(log_predictions: chex.Array,
   chex.assert_type([log_predictions, log_targets], float)
   loss = jnp.exp(log_targets) * (log_targets - log_predictions)
   return jnp.sum(loss, axis=-1)
+
+def hinge_loss(predictions: chex.Array, targets: chex.Array) -> chex.Array:
+  """
+  Computes the hinge loss for Binary Classification.
```
mkunesch, line 509: I think we can use lower case letters for binary classification. Also, please move this to the first line of the docstring.

optax/_src/loss.py
```diff
@@ -503,3 +503,19 @@ def kl_divergence_with_log_targets(log_predictions: chex.Array,
...
+
+  Args:
+    predictions: Predicted values with shape [..., dim].
+    targets: Target values with shape [..., dim].
```
mkunesch, line 513: Please break this line at 80 characters (I think it's broken earlier than that).

optax/_src/loss.py
```diff
@@ -503,3 +503,19 @@ def kl_divergence_with_log_targets(log_predictions: chex.Array,
...
+      Expected to be in range [-1, 1].
+
+  Returns:
+    Average Hinge Loss
```
mkunesch, line 517: Other optax losses don't average over the batch dimension (see #193) - I think this also means we shouldn't average in this case?

optax/_src/loss.py
```diff
@@ -503,3 +503,19 @@ def kl_divergence_with_log_targets(log_predictions: chex.Array,
...
+  """
+  assert predictions.shape[0] == targets.shape[0], "Predictions and Target lengths must be consistent."
+  # Binary Case
```
mkunesch, line 520: Let's remove this comment for now if we only merge the binary version for now.

optax/_src/loss.py
```diff
@@ -503,3 +503,19 @@ def kl_divergence_with_log_targets(log_predictions: chex.Array,
...
+  return jnp.mean(jnp.maximum(0, 1 - predictions * targets))
```
mkunesch, line 521: As I wrote above, I don't think we should add the `jnp.mean` in analogy to other optimizer (see #193) and also since this would take the mean over the whole array - but the user might want some axis to be excluded.

optax/_src/loss.py
```diff
@@ -503,3 +503,19 @@ def kl_divergence_with_log_targets(log_predictions: chex.Array,
   chex.assert_type([log_predictions, log_targets], float)
   loss = jnp.exp(log_targets) * (log_targets - log_predictions)
   return jnp.sum(loss, axis=-1)
+
+def hinge_loss(predictions: chex.Array, targets: chex.Array) -> chex.Array:
+  """
+  Computes the hinge loss for Binary Classification.
+
+  Args:
+    predictions: Predicted values with shape [..., dim].
```
mkunesch, line 512: Could you add more explanation that predictions aren't the class predictions but the output of the decision function e.g. w*x + b. Perhaps we can also find a name that reflects this? I'm not sure I like the sklearn name `pred_decision` though. Perhaps `predictor_outputs`?

optax/_src/loss.py
```diff
@@ -503,3 +503,19 @@ def kl_divergence_with_log_targets(log_predictions: chex.Array,
...
+    targets: Target values with shape [..., dim].
+      Expected to be in range [-1, 1].
```
mkunesch, line 514: I think they have to be strictly in the set {-1, 1}, i.e. either -1 or 1, not in the interval [-1, 1]. Could you clarify this?

optax/_src/loss.py
mkunesch, line 512: I think in the binary case we do not have to specify that the last dimension must match - that would only be if we have multiple class labels to ensure the number of classes is the same.

optax/_src/loss.py
```diff
@@ -503,3 +503,19 @@ def kl_divergence_with_log_targets(log_predictions: chex.Array,
...
+
+  Returns:
+    Average Hinge Loss
+  """
+  assert predictions.shape[0] == targets.shape[0], "Predictions and Target lengths must be consistent."
```
mkunesch, line 519: I don't think we need this check in the binary case so I'd remove this line. We'd only need it if we want to check that the number of classes matches in the multi-class setting. In this case, we can just let jax numpy raise an error if it can't broadcast the arrays to compatible shapes below.

optax/_src/loss.py
```diff
@@ -503,3 +503,19 @@ def kl_divergence_with_log_targets(log_predictions: chex.Array,
   chex.assert_type([log_predictions, log_targets], float)
   loss = jnp.exp(log_targets) * (log_targets - log_predictions)
   return jnp.sum(loss, axis=-1)
+
+def hinge_loss(predictions: chex.Array, targets: chex.Array) -> chex.Array:
```
mkunesch, line 507: Please add tests for this function that compare against analytically calculated values, ideally for several different array ranks and shapes. If you wrap the test in `chex.all_variants` you can test all the jax-transformed versions of the function (see the other tests for how to do this). Thanks!

mkunesch (changes requested):
Thanks a lot! I added a more detailed review for the binary version - as I wrote earlier, let's merge this first and then consider the multi-class version separately.

heytanay:
Hi @mkunesch ! I was just in the process of committing an updated version. A question though, can I use black to format the code or any other formatter?

heytanay:
I hope I've added all the requested changes. I also ran `bash test.sh` and all tests have passed. Making the PR ready for review.

optax/_src/loss_test.py
```diff
@@ -475,5 +475,20 @@ def test_batched(self):
         self.exp,
         atol=1e-4)
 
+class HingeLossTest(parameterized.TestCase):
+
+  def setUp(self):
+    super().setUp()
+    self.ys = np.array([1, 0, 1, 0, 0, 1, 0, 0, 1, 0])
```
mkunesch, line 482: Do I understand correctly that `ys`, the predictor output can be in the range `(-Inf, Inf)`? I think it would be important to test this, also as a way of documenting the function. In particular, I think your test don't test the `maximum` behaviour in that they would also pass for the implementation `return delta - predictor_outputs * targets`.

mkunesch (changes requested):
Thanks a lot! This is looking great! Just a few very minor comments.

Re the formatting: I am working for better instructions on this for the `Contributing.md` file, but your code looks great so if you are ok with this you can just leave the formatting as is and I'll run the internal linter over it as it's being merged (it will only make very minor changes).

Also, just to say that we will only be able to merge this after the ICLR paper submission deadline on the 28th of September but we can approve and everything else beforehand so that it can be merged immediately on the 29th.

Thanks a lot!

optax/_src/loss_test.py
heytanay: Thanks for the review! From what I understand by this is that I should add a different example here that would test the maximum behavior?

heytanay: Hi @mkunesch, can you please clarify on this?

mkunesch: Yes, please - you could just add some numbers here for which the maximum behaviour is important. Also, currently the ys are only 0 and 1 but if I understand correctly they can be any float in the range (-Inf, Inf) in practice so the test input should reflect this.

mkunesch (changes requested):

optax/_src/loss.py
```diff
@@ -503,3 +503,20 @@ def kl_divergence_with_log_targets(log_predictions: chex.Array,
...
+  predictor_outputs: chex.Array,
+  targets: chex.Array,
+  delta: int = 1) -> chex.Array:
+  """Computes the hinge loss for binary classification.
+
+  Args:
+    predictor_outputs: Outputs of the decision function.
+    targets: Target values.
+      Target values should be strictly in the set {-1, 1}.
+    delta: Margin
```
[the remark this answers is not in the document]
heytanay: @mkunesch Thanks for the suggestions but I just realised that just like in sklearn, most frameworks with loss functions actually only use `1` as the value for delta (or margin) and so maybe it's better if we remove this parameter altogether and just set it manually to `1` in the function, might also remove some potential for confusion if we were to go ahead and name it `margin`.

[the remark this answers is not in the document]
mkunesch: That sounds good - I think we can remove it similar to sklearn and pytorch. Thanks!

mkunesch (changes requested):
Looks great!

Could you please add the hinge_loss to the api docs similar to the [huber_loss](https://github.com/deepmind/optax/blob/7063ce8022ce6165926de369d6fefa207e318127/docs/api.rst#L481) (both to the summary and the detailed docs)?

Also, the license check doesn't pass - you might need to make sure the commits use the right  e-mail address or sign the contributor agreement again with the new email.

Other than that looks good to me!

Thanks!

heytanay:
> Also, the license check doesn't pass - you might need to make sure the commits use the correct e-mail address or sign the contributor agreement again with the new email.

I messed up the whole thing by committing without verifying from my MacBook, will doing a new commit with my Gmail id (which is CLA signed) fix that, or do I need to close this PR and make a new one?

mkunesch:
Hi! Thanks a lot for fixing it! I think it's ready to be merged but the checks don't pass at the moment - could it be because you forgot to add the intended changes to commit [2d92948](https://github.com/deepmind/optax/pull/409/commits/2d929482387f33fe6a8c7000c3b844971c6049c0)?

heytanay:
@mkunesch Yes! No idea why the changes were not reflected here. Let me commit and push it now :)

mkunesch (approved):
Looks great - thanks a lot!

I might have to make some very minor formatting changes as I merge as the internal linter is slightly stricter but I'll take it from here!

Thanks again for this contribution!

PR #410, opened 2022-09-04
Adds Adan Optimizer

### LINKED ISSUE #401: Adan Optimizer
Is there any interest in adding the [Adan optimizer](https://twitter.com/davisblalock/status/1561976182567870465?t=AHwO3of5ivzgE0dW06hZvA&s=35) to optax? If so I can do it

joaogui1:
Closes #401 
Implementation based on the [official code](https://github.com/sail-sg/Adan/blob/main/adan.py)


mtthss:
Thank you very much! If the paper claims stack up, this will be very useful to the jax community.

Btw there is pytorch reference code from the authors of the paper,
https://github.com/sail-sg/Adan

Would you mind loading both the pytorch and the optax implementation in a colab and show that they match?
it might highlight subtle differences that might be hard to spot from just staring at the code

joaogui1:
Hey @mtthss, always happy to help :D 
Yeah, I based my code in the official implementation and here's the [colab](https://colab.research.google.com/drive/16Dx_raEsWq0wrQnGRO2ab88XMAIranMx?usp=sharing)
At the moment I have 2 questions:
1. The current implementation seems to generate slightly different results, for example when using a 256x256 matrix and doing 1000 updates the norm of the difference between the pytorch and jax version is 2.4e-5 on the CPU, is that expected or should I keep looking for more divergences in the implementation
2. The default for the official implementation of Adan seems to use a different [type of weight decay](https://github.com/sail-sg/Adan/blob/main/adan.py#L151), where the params are divided by (1 + lr * wd) instead of multiplied by (1 - lr * wd), this causes a large difference in behavior, but I'm not exactly sure what is meant by the `no_prox` parameters that controls what kind of weight decay is being used

joaogui1:
I've computed the relative error and its in the order of 10^-8 (though it still grows as we do more updates), thoughts? @mtthss

joaogui1:
Hi @Zach-ER! Thanks for the thorough response, for checking with the authors and running my code!
1. From the authors' response to your issue it looks like their results were achieved with `eps` outside the square root, so I think we can leave it as it, wdyt?
2. Makes sense, I believe I'll resuse it
3. Yeah I'm not sure about the best way to proceed here, maybe I can implement the no_prox condition on the transform and create to alias, `adan` and `adan_no_prox`, wdyt?

Once again thanks for the comments :D

Zach-ER:
1. I would prefer leaving the defaults as they are but also having an `eps_root` argument, defaulting to 0.0. This has the benefit of being a closer match to the `adam` signature and also letting people implement it as in the paper (if they so choose).
2. 👍🏻 — just make sure the name is appropriate.
3. I think we _should_ definitely have a method that matches their default implementation. I would put a condition into the `adan` optimizer that matches the reference behaviour. This will be slightly fiddly. Do you want to have a try at this? If not, I would be happy to draft something.

minor nitpicking:
docstring needs fixing for b1, b2 (and b3 needs adding). Something like 
```
    b1: Decay rate for the exponentially weighted average of gradients.
    b2: Decay rate for the exponentially weighted average of difference of
      gradients.
    b3: Decay rate for the exponentially weighted average of the squared term.
```

Zach-ER:
Hi there,
I have posted a new issue in the original repo [here](https://github.com/sail-sg/Adan/issues/12). If they say their experiments were conducted with `no_prox=False`, then I think we can ignore the other condition and your PR reproduces their algorithm and fits well style-wise with the rest of the codebase.

Will update this when the authors respond.

Zach-ER:
OK, the authors have responded.
Their experiments use `no_prox=False`, the condition that you have not implemented, so I think we do need to implement that one and match their algorithm exactly.

what I would do:

1. add a `use_proximal_operator` boolean argument, defaulting to `True`, to match what's in the paper. (this is apparently what `prox` is short for).
2. if `False`, implement exactly as you've already done
3. if `True`, need a slightly less standard implementation

I wrote a `jax` version that matches their implementation, based on your gist and codebase. It is [here](https://gist.github.com/Zach-ER/5204bd4a4d7c4e25061f8653d125205c#file-adan_compare-ipynb) and matches their results completely with weight decay turned on.

If you could integrate what I've written into your PR, that would be great — if not, I will find some time to do it (but quite busy at the moment).

Again, thanks for your work — this will be a great addition to the library 🎖

joaogui1:
Thank you very much for all the help! 
Right now I'm at RIIAA Ecuador, but I'll try to integrate your code as soon as I'm back home, around Sunday or Monday :)

joaogui1:
So it does pass the alias tests, but it looks like sphinx is erroing now, any ideas of a quick fix (I'm not experienced with Sphinx) ? @Zach-ER

hbq1:
@joaogui1 it should be fixed now, could you update the PR?

joaogui1:
@hbq1 update how? Should I merge main?

hbq1:
>Should I merge main?
Yes 👍

docs/api.rst
```diff
@@ -280,6 +288,7 @@ Optax Transforms and States
 .. autofunction:: scale
 .. autofunction:: scale_by_adam
 .. autofunction:: scale_by_adamax
+.. autofunction:: scale_by_adan
```
hbq1, line 291: Could you also add `scale_by_proximal_adan`?

optax/_src/alias.py
```diff
@@ -302,6 +302,65 @@ def adamw(
...
+      or a Callable that returns such a pytree given the params/updates.
+      The leaves should be booleans, `True` for leaves/subtrees you want to
+      apply the weight decay to, and `False` for those you want to skip. Note
+      that the Adam gradient transformations are applied to all parameters.
+
+  Returns:
+    the corresponding `GradientTransformation`.
+  """
+  if use_proximal_operator:
+    return transform.scale_by_proximal_adan(
```
hbq1, line 350: Could you comment that `_scale_by_learning_rate` is not needed here

Zach-ER:
Yes, LGTM. Looking forward to trying it out some more 🙌🏻

joaogui1:
Done @hbq1

mkunesch (commented):
Hi! I am doing an internal review for this PR and have noticed the following potential problem:

The optax transforms are all composable building blocks that can be chained together. While `scale_by_proximal_adan` can be chained I think adding any transforms _on top_ of it might give unexpected results since it calculates new parameters internally and then the `new_updates` based on them but this calculation isn't aware of the other transforms a user might add on top of `scale_by_proximal_adan`.

I think in that sense it might be similar to the lookahead optimizer which we have moved to a separate file (there were other reasons for this too) and added [warnings](https://github.com/deepmind/optax/blob/3d3b8c3a622a9048a3310a30ba44ea3141fc2c51/optax/_src/lookahead.py#L75) to the docstring.

Do you agree that this could be a problem or did I miss something? If it is a problem, can we rewrite `scale_by_proximal_adan` such that it can be chained with further transforms (I don't think so but I haven't spent much time thinking about it)? If not, I think we should discuss a general way to deal with these cases (optimizers that cannot be chained any further) in the optax API.

Thanks a lot and let me know what you think!

Zach-ER:
Yes, I think that this could be a problem.

I agree that it is in a similar boat to the lookahead optimizer.

carlosgmartin:
Any update on this?

fabianp:
@carlosgmartin no updates, this PR is currently orphaned. Do you want to take over?

carlosgmartin:
@fabianp What changes, if any, need to be made to @joaogui1's PR?

And what's the consensus on @mkunesch's questions?

fabianp:
1. A first step would be to update with main, currently there are some conflicts with the current head.
2. I believe the issues highlighted by @mkunesch only apply to the proximal version `scale_by_proximal_adan`. I would suggest focusing first on `scale_by_adan` which shouldn't have those issues if I understood correctly.

carlosgmartin:
@fabianp Created a PR: #1090.

fabianp:
excellent, closing this PR then

PR #431, opened 2022-10-03
Make sigmoid_binary_cross_entropy work for infinite logits.

jondeaton:


jondeaton:
The proposed implementation does not solve the issue.

Issue #429, opened 2022-10-02
adamax and adamaxw not exposed.


nasyxx:
adamax and adamaxw not exposed

I noticed that optax did not expose `adamax` and `adamaxw` defined in `_src/alias.py` (https://github.com/deepmind/optax/blob/2df58b4f85055f69ffbc12d9f0eddbe6edf5a740/optax/_src/alias.py#L703-L776) in `__init__.py`. Is this done on purpose because they have some problems? Or just missed it?

hbq1:
Hi @nasyxx, this is a great catch, thank you! 
Would you like to send a PR which would expose these functions?

nasyxx:
> Would you like to send a PR which would expose these functions?

Hi, sure.  It's #433

PR #433, opened 2022-10-04
Expose adamax and adamaxw

nasyxx:
Issue #429

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -10,6 +10,8 @@ Common Optimizers
     adagrad
     adam
     adamw
+    adamax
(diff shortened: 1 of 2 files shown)


hbq1:
Thanks! Could you also add these symbols to the Sphinx [docs](https://github.com/deepmind/optax/blob/master/docs/api.rst)?

nasyxx:
> Could you also add these symbols to the Sphinx
Done!  Please let me know if there is anything else that needs to be changed.

hbq1 (approved):

PR #465, opened 2022-12-08
Better tests for utils

### LINKED ISSUE #404: Better tests for utils.
Optax tests did not catch a problem with one of the type annotations in #367. This is due to `utils` not having good test coverage. 

I'm marking this as "good first issue". Any tests for `utils` would be very welcome! No need to write tests for all of them at once, PRs with only a single test at a time are very welcome.

acforvs:
Hi, hopefully this one closes #404

### DIFF
--- a/optax/_src/utils_test.py
+++ b/optax/_src/utils_test.py
@@ -20,6 +20,7 @@ from absl.testing import absltest
 from absl.testing import parameterized
 
 import jax
+import jax.numpy as jnp
 
 from optax._src import utils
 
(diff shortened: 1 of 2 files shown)


optax/_src/utils_test.py
```diff
@@ -61,5 +62,70 @@ def fn(inputs):
         sum(jax.tree_util.tree_leaves(
             jax.tree_util.tree_map(lambda x: x**2, inputs))))
 
+
+class MultiNormalDiagFromLogScaleTest(parameterized.TestCase):
+
+  @staticmethod
```
hbq1, line 68: No need to use a @staticmethod here (see https://google.github.io/styleguide/pyguide.html#217-function-and-method-decorators), consider making it a class function.

optax/_src/utils_test.py
```diff
@@ -61,5 +62,70 @@ def fn(inputs):
...
+    def tuple_shape(shape):
+      if isinstance(shape, tuple):
+        return shape
+      return tuple([shape])
+
+    loc, scale = self._get_loc_scale(loc_shape, scale_shape)
+    dist = utils.multi_normal(loc, scale)
+    self.assertIsInstance(dist, utils.MultiNormalDiagFromLogScale)
+    mean, log_scale = dist.params
+    self.assertTrue(tuple(mean.shape) == tuple_shape(loc_shape))
```
hbq1, line 92: Consider using `self.assertEqual(a, b)` instead of `self.assertTrue(a == b)` for better error messages.

optax/_src/utils_test.py
```diff
@@ -61,5 +62,70 @@ def fn(inputs):
...
+
+  @parameterized.parameters(
+    (2, 3),
+    ((2, 3), (3, 2)),
+    ((2, 4), (3, 4)),
+    ((1, 2, 1, 3), (2, 1, 4, 4)),
+  )
+  def test_init_unsuccessful_broadcast(self, loc_shape, scale_shape):
+    loc, scale = self._get_loc_scale(loc_shape, scale_shape)
+    with self.assertRaises(ValueError):
```
hbq1, line 104: `self.assertRaisesRegex(ValueError, <regexp>)` would be more safe here

optax/_src/utils_test.py
```diff
@@ -61,5 +62,70 @@ def fn(inputs):
...
+      utils.multi_normal(loc, scale)
+
+  @parameterized.parameters(list, tuple)
+  def test_sample_input_sequence_types(self, sample_type):
+    sample_shape = sample_type((4, 5))
+    loc_shape = scale_shape = (2, 3)
+    loc, scale = self._get_loc_scale(loc_shape, scale_shape)
+    dist = utils.multi_normal(loc, scale)
+    samples = dist.sample(sample_shape, jax.random.PRNGKey(239))
```
hbq1, line 113: SPB Lyceum 239? :)

optax/_src/utils_test.py
```diff
@@ -61,5 +62,70 @@ def fn(inputs):
...
+    self.assertTrue(samples.shape == tuple(sample_shape) + loc_shape)
```
hbq1, line 114: ditto re: `assertEqual`

optax/_src/utils_test.py
```diff
@@ -61,5 +62,70 @@ def fn(inputs):
...
+    ('2d', (2, 3)),
+    ('4d', (1, 2, 3, 4)),
+  ])
+  def test_log_prob(self, shape):
+    loc, scale = self._get_loc_scale(shape, shape)
+    dist = utils.multi_normal(loc, scale)
+    probs = dist.log_prob(jnp.ones(shape=shape, dtype=jnp.float32))
+
+    self.assertIsInstance(probs, jnp.DeviceArray)
+    self.assertFalse(bool(probs.shape))
```
hbq1, line 127: ditto

hbq1:
Hi @acforvs, many thanks for your PR! 
It looks good to me; I left only a few comments. Could you also add tests for the rest of helpers from utils.py, so we could close the original issue?

acforvs:
Hi @hbq1 , thank you for the review!

Sure thing, I'll add the tests for the remaining functions, but I have some questions about them:
1. The docstring for `set_diags` [states](https://github.com/deepmind/optax/blob/master/optax/_src/utils.py#L56) that the shape of `new_diags` should be `NxD`. However, `new_diags` is only used in one place where it is [flattened](https://github.com/deepmind/optax/blob/master/optax/_src/utils.py#L70).

As a result, examples like
```python
a = jnp.ones(shape=(3, 2, 2))
diags = jnp.array([[[3, 3, 3, 3, 3, 3]]])
print(a.shape, diags.shape)  # prints (3, 2, 2) (1, 1, 6)
optax._src.utils.set_diags(a, diags)
```
or 
```python
a = jnp.ones(shape=(3, 2, 2))
diags = jnp.array([[3, 3, 3], [3, 3, 3]])
print(a.shape, diags.shape)  # prints (3, 2, 2) (2, 3)
optax._src.utils.set_diags(a, diags)
```
work just fine

**Is this by design or should we raise an error in case of a shape mismatch?** 
Currently, the function is used internally in two places: [1](https://github.com/deepmind/optax/blob/master/optax/_src/stochastic_gradient_estimators.py#L237), [2](https://github.com/deepmind/optax/blob/master/optax/_src/stochastic_gradient_estimators.py#L302). The comments specify that the dimension of the passed variable is indeed `NxD` there, so I'd suggest adding another `assert`. What are your thoughts?

2. `cast_tree` uses `.astype` which wouldn't work for simple datatypes, such as 
```python
tree = dict(a=2.5, b=dict(c=-2.5))
tree = jax.tree_util.tree_map(lambda x : x, tree)
optax._src.utils.cast_tree(tree, int)
```
Would it be okay to only test this function for dicts of jnp.arrays of differerent dtypes? Something like 
```python
tree = dict(a=jnp.array(2.5), b=dict(c=jnp.array(-2.5)))
tree = jax.tree_util.tree_map(lambda x : x, tree)
optax._src.utils.cast_tree(tree, int)
```

optax/_src/utils_test.py
acforvs: yep :)

hbq1: very cool :)

optax/_src/utils_test.py
hbq1: Yes, thanks!

hbq1:
Great, thank you!

1. It is used in some internal code, so the correct behaviour would be to raise an exception in case of shape mismatch; it'd be great if you add this to the implementation!
2. Testing only with trees of `jnp.array`s would suffice.

acforvs:
I've fixed the errors and added tests for the rest of the utils.py functions. 
I also changed `assert` to a `ValueError` in `set_diags`. Please let me know how it looks

hbq1 (approved):
Thanks for the changes and for the nice PR 👍

PR #473, opened 2023-01-13
fixed NaN issues in `kl_divergence` loss function

LukasMut:
- The current `kl_divergence` loss function yields NaNs because it ignores the fact that $\log(0) = -\infty$.
- The new `kl_divergence` loss function counteracts this and gives 0 for every entry in `targets` that is 0, which is the numerically correct implementation of `kl_divergence` (see my changes).
- There's currently no way to check whether $\log(p)$ or $\log(q)$ contain negative infs in `kl_divergence_with_log_targets`, and, hence, $\exp(\log(p))$ may yield NaNs. I've added a check that will throw an error if any of the two log probability distributions contain negative `infs` because we don't want to take the exponential of $\log(p)$ if that's actually the case.
- In addition, I've added a convex version of the Kullback-Leibler divergence loss. The implementation is in line with [scipy's version](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.kl_div.html#scipy-special-kl-div) and tested on my end.


LukasMut:
@tomhennigan @froystig Is there anything else I need to do to get this PR merged?

optax/_src/loss.py
```diff
@@ -461,6 +461,35 @@ def ctc_loss(logits: chex.Array,
...
+      [..., dim]. Expected to be in the log-space to avoid underflow.
+    targets: Probabilities of target distribution with shape [..., dim].
+      Expected to be strictly positive.
+
+  Returns:
+    Kullback-Leibler divergence of predicted distribution from target
+    distribution with shape [...].
+  """
+  chex.assert_type([log_predictions, targets], float)
+  chex.assert_tree_all_finite(log_predictions)
```
hbq1, line 486: This value assertion will not work in unchexified code (see [the docs](https://github.com/deepmind/chex#static-and-value-aka-runtime-assertions)), so it's probably better to remove it from the default implementation.

optax/_src/loss.py
```diff
@@ -503,6 +533,7 @@ def kl_divergence_with_log_targets(log_predictions: chex.Array,
     distribution with shape [...].
   """
   chex.assert_type([log_predictions, log_targets], float)
+  chex.assert_tree_all_finite([log_predictions, log_targets])
```
hbq1, line 536: ditto

hbq1 (changes requested):
Thanks for the PR! I added one comment, overall looks good to me.

optax/_src/loss.py
```diff
@@ -461,6 +461,35 @@ def ctc_loss(logits: chex.Array,
   return per_seq_loss
 
 
+def convex_kl_divergence(log_predictions: chex.Array,
```
hbq1, line 464: Could you also add it to the public symbols in __init__.py and to the docs (see `kl_divergence` example)?

optax/_src/loss.py
LukasMut: I've just removed that line.

optax/_src/loss.py
LukasMut: I've removed `chex.assert_tree_all_finite(...)`. Should I additionally remove `chex.assert_type(...)` as well? I see `chex.assert_type(...)` in every loss function.

optax/_src/loss.py
LukasMut: Done.

optax/_src/loss.py
hbq1: This is not necessary because `chex.assert_type` is a static assertions (i.e. it does not need access to actual tensor values).

hbq1:
Thanks! One last thing: could you add a few tests for the new loss? See e.g. `KLDivergenceTest` from loss_test.py.

LukasMut:
> Thanks! One last thing: could you add a few tests for the new loss? See e.g. `KLDivergenceTest` from loss_test.py.

I've added a test for the convex kl-divergence loss function and changed the test for kl-divergence loss function so that it checks whether the 0-case does not yield problems.

optax/_src/loss_test.py
```diff
@@ -421,6 +421,34 @@ def test_repeat_with_one_to_one_alignment(self):
...
+  def test_scalar(self):
+    np.testing.assert_allclose(
+        self.variant(loss.convex_kl_divergence)(self.log_ps[0], self.qs[0]),
+        self.exp[0],
+        atol=1e-4)
+
+  @chex.all_variants
+  def test_batched(self):
+    np.testing.assert_allclose(
+        self.variant(loss.kl_divergence)(self.log_ps, self.qs),
```
hbq1, line 447: convex_kl_divergence

LukasMut: Sorry. I've just fixed that typo.

hbq1 (approved):

hbq1:
Thank you! We'll merge it today or tomorrow.

PR #504, opened 2023-02-24
Fix JIT for `piecewise_interpolate_schedule`, `cosine_onecycle_schedule`, `linear_onecycle_schedule `

brentyi:
Fixes https://github.com/deepmind/optax/issues/503!

Tests might be nice for this kind of thing, but I don't have a good sense for how to best implement them.


hbq1 (approved):
Thanks for the fix!

PR #516, opened 2023-03-22
Allow gradient transform parameters to be dynamic

NeilGirdhar:


NeilGirdhar:
@mtthss  Would you mind taking a look at this?

mtthss:
Hello. I was on paternity leave for most of the past year. Are you still having this issue? Happy to look into it if that's the case

mtthss:
which arguments are causing errors to you?

NeilGirdhar:
(Of course, my dream would be that you adopt the dynamic design so that I don't have to maintain my shim library 😄.)

NeilGirdhar:
@mtthss Do you mind taking a look at this?

rdyro:
I'm a new maintainer. At some point, optax moved to [https://optax.readthedocs.io/en/latest/api/optimizer_schedules.html#optax.schedules.inject_hyperparams](https://optax.readthedocs.io/en/latest/api/optimizer_schedules.html#optax.schedules.inject_hyperparams) to make parameters available in the state.

I think the point of these type annotations is to explicitly indicate that (like you're pointing out) the parameters are closed over and not really modifiable without `inject_hyperparams`.

NeilGirdhar:
>  https://optax.readthedocs.io/en/latest/api/optimizer_schedules.html#optax.schedules.inject_hyperparams to make parameters available in the state.

That's a separate issue from this PR.  (I believe, in fact, it's the first issue #1.)

> I think the point of these type annotations is to explicitly indicate that (like you're pointing out) the parameters are closed over and not really modifiable without inject_hyperparams.

Even if the parameters are closed over, that doesn't make them floats.  A closed-over Jax array is still a Jax array.  The annotations are simply wrong.  It's just that most users probably don't hit this because most users are probably just using fixed values.

rdyro:
Let me talk to @vroulet

vroulet:
That's a completely fair point. We'll make an attempt to change the types (this will affect all internal code so many bugs possible) but a priori it should not pose problems. We'll keep you updated Neil. (It'll be next month a priori)

NeilGirdhar:
@vroulet Any update on this?

PR #519, opened 2023-04-13
Added protobuf versioning to dependencies

andreped:
The current CIs fail due to some protobuf bug. See latest build [here](https://github.com/deepmind/optax/actions/runs/4685065431).

Note that this impacts all CIs, not just the scheduled one, and people might observe the same issue installing `optax` on their local hardware.

This is fixed by restricting protobuf version to `<=3.20`, which was also suggested as a fix in the error log below.

```
ERROR mnist_test.py - TypeError: Descriptors cannot not be created directly.
If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
If you cannot immediately regenerate your protos, some other possible workarounds are:
 1. Downgrade the protobuf package to 3.20.x or lower.
 2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).

More information: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
```


andreped:
This seem to have been fixed in another commit. Hence, this PR is no longer needed. Closing for now.

PR #537, opened 2023-06-14
Add poly_loss_cross_entropy to public api

acforvs:


mtthss:
This is now exposed as part of the losses/ subpackage.

PR #574, opened 2023-08-15
Add an implementation of a sigmoid focal loss

v0lta:
The idea is to support a sigmoid focal loss, as presented in 
https://arxiv.org/pdf/1708.02002.pdf
and implemented in 
https://github.com/facebookresearch/fvcore/blob/main/fvcore/nn/focal_loss.py .

Porting the function was possible because all necessary components already exist in Optax.

### DIFF
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -671,13 +671,18 @@ def rmsprop(
     nesterov: bool = False
 ) -> base.GradientTransformation:
   # pylint: disable=line-too-long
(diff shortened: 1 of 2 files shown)


mtthss:
Apologies for the long delay. We were going through a very busy period internally but are now resuming regular support here on github. Can you please fix the conflict (due to the recent creation of the optax/losses/ subpackage) so we can then merge this in?

v0lta:
okay I am on it.

PR #576, opened 2023-08-17
add kl_divergence_with_log_targets in __init__.py

### LINKED ISSUE #556: kl_divergence_with_log_targets is not imported in __init__.py
Hi.

kl_divergence_with_log_targets is implemented in optax/_src/loss.py, and is tested in optax/_src/loss_test.py.
However, optax/__init__.py does not import the function so it is not available from optax. I think this is a bug.

h-terao:
Fix #556

### DIFF
--- a/optax/__init__.py
+++ b/optax/__init__.py
@@ -88,6 +88,7 @@ from optax._src.loss import ctc_loss_with_forward_probs
 from optax._src.loss import hinge_loss
 from optax._src.loss import huber_loss
 from optax._src.loss import kl_divergence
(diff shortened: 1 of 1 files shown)


mtthss:
Hello, thanks for the proposed change and apologies for the long delay,
we were going through a somewhat busy period internally, but are now resuming regular operations.

This change is subsumed by the recent change I pushed making losses into its own optax.losses subpackage, where you will now found `kl_divergence_with_log_targets` correctly exposed.

PR #599, opened 2023-10-11
sigmoid focal loss

v0lta:
The idea is to support a sigmoid focal loss, as presented in
https://arxiv.org/pdf/1708.02002.pdf
and implemented in
https://github.com/facebookresearch/fvcore/blob/main/fvcore/nn/focal_loss.py .

Porting the function was possible because all necessary components already exist in Optax.

As discussed initially at https://github.com/google-deepmind/optax/pull/574#issuecomment-1754637077,
a revert of the changes to resolve the conflict closed the request, so I am re-opening it here.

I don't know if a segmentation module is the best place possible. It could also be in classification since we are classifying pixels, but I fear new users would find that confusing, so I am thinking a segmentation module might be appropriate.


v0lta:
Converted to draft mode, for now. I would like to add test cases.

v0lta:
I think it's ready. If you have feedback, please let me know.

optax/losses/segmentation.py
```diff
@@ -0,0 +1,49 @@
...
+from optax.losses.classification import sigmoid_binary_cross_entropy
+
+def sigmoid_focal_loss(
+    logits:  chex.Array,
+    labels:  chex.Array,
+    alpha: Optional[float] = None,
+    gamma: float = 2,
+) ->  chex.Array:
+    """Compute a sigmoid focal loss.
+    Implementation of the focal loss as used https://arxiv.org/abs/1708.02002.
```
fabianp, line 16: please add references on its own docstring subsection, something like
```

References
 X et al. 2018, http:/...
```

See _src/alias.py for many examples

optax/losses/segmentation.py
```diff
@@ -0,0 +1,49 @@
+from typing import Optional
```
mblondel, line 1: I don't know about this loss but why can't it go to `classification.py`?

optax/losses/segmentation_test.py
```diff
@@ -0,0 +1,27 @@
...
+
+import chex
+import jax
+import jax.numpy as jnp
+import numpy as np
+
+from optax.losses import segmentation
+from optax.losses import classification
+
+class SoftmaxCrossEntropyTest(parameterized.TestCase):
```
mblondel, line 12: `SigmoidFocalLossTest`

optax/losses/segmentation.py
v0lta: Done.

optax/losses/segmentation_test.py
v0lta: Fixed.

optax/losses/segmentation.py
v0lta: We can think about segmentation as a classification of pixels. The loss expects the network output to be a map with class probabilities for each pixel. Having a segmentation file would stress that this loss is not meant for tasks like image classification, with per-class probabilities for the entire image.

v0lta: Sine the file is not visible to end-users in the documentation, I don't think my initial argument matters. I have moved the function into `classification.py`.

mblondel: Thanks. That was also my opinion.

fabianp:
there are some conflicts with master. Could you resolve these before merging @v0lta ? Thanks!

optax/losses/classification.py
```diff
@@ -430,3 +430,49 @@ def ctc_loss(
...
+
+
+
+def sigmoid_focal_loss(
+    logits:  chex.Array,
+    labels:  chex.Array,
+    alpha: Optional[float] = None,
+    gamma: float = 2,
+) ->  chex.Array:
+  """Compute a sigmoid focal loss as proposed by Lin et al.
```
fabianp, line 442:
```suggestion
  """Compute a sigmoid focal loss.
```

optax/losses/classification.py
```diff
@@ -430,3 +430,49 @@ def ctc_loss(
...
+  The focal loss is a re-weighted cross entropy for unbalanced problems.
+  Use this loss function if classes are not mutually exclusive.
+  See `sigmoid_binary_cross_entropy` for more information.
+
+  References:
+    Lin et al. https://arxiv.org/pdf/1708.02002.pdf
```
fabianp, line 448:
```suggestion
    Lin et al. 2018. https://arxiv.org/pdf/1708.02002.pdf
```

optax/losses/classification.py
```diff
@@ -430,3 +430,49 @@ def ctc_loss(
...
+  
+  Args:
+    logits: A float array of arbitrary shape.
```
fabianp, line 451:
```suggestion
    logits: Array of floats. The predictions for each example.
```

optax/losses/classification.py
```diff
@@ -430,3 +430,49 @@ def ctc_loss(
...
+      The predictions for each example.
+    labels: A float array, its shape must be identical to
```
fabianp, line 453:
```suggestion
    labels: Array of floats of the same shape as
```

optax/losses/classification.py
```diff
@@ -430,3 +430,49 @@ def ctc_loss(
...
+      that of logits. It containes the binary
```
fabianp, line 454:
```suggestion
      Contains the binary
```

fabianp, line 454:
```suggestion
      that of logits. Contains the binary
```

optax/losses/classification.py
```diff
@@ -430,3 +430,49 @@ def ctc_loss(
...
+      classification label for each element in logits
+      (0 for the out of class and 1 for in class).
+      This array is often one-hot encoded.
```
fabianp, line 457: Does this mean that it's often either 0 or 1? If this is true, I think it's a bit confusing since we just said these are logits. I would remove the last phrase since it seems unnecessary. What do you think?

optax/losses/classification.py
```diff
@@ -430,3 +430,49 @@ def ctc_loss(
...
+    gamma: Exponent of the modulating factor (1 - p_t) to
+      balance easy vs hard examples.
+
+  Returns:
+    A loss value array with a shape identical to the logits and target
+    arrays.
+  """
+  chex.assert_type([logits], float)
+  labels = labels.astype(logits.dtype)
+  # see also the original papers implementation at:
```
fabianp, line 469:
```suggestion
  # see also the original paper's implementation at:
```

optax/losses/classification.py
v0lta: It's probably not wise to talk about the logits all the time in the label docstring. [8a85e61](https://github.com/google-deepmind/optax/pull/599/commits/8a85e6112206a77940cdc115cdfa4ebc61027071) attempts to address your concerns by re-writing the docstring. The last phrase is gone now.

v0lta:
> there are some conflicts with master. Could you resolve these before merging @v0lta ? Thanks!

One test says `Waiting for internal safe review approval`, I don't think I can fix this.

optax/losses/_classification.py
```diff
@@ -430,3 +430,48 @@ def ctc_loss(
...
+  """
+  chex.assert_type([logits], float)
+  labels = labels.astype(logits.dtype)
+  # see also the original paper's implementation at:
+  # https://github.com/facebookresearch/fvcore/blob/main/fvcore/nn/focal_loss.py
+  p = jax.nn.sigmoid(logits)
+  ce_loss = sigmoid_binary_cross_entropy(logits, labels)
+  p_t = p * labels + (1 - p) * (1 - labels)
+  loss = ce_loss * ((1 - p_t) ** gamma)
+  if alpha:
```
the-xentropy, line 474: Won't this prevent this from being JIT'able? I ran it briefly and it errored on this. It also seems like in the case that alpha=0.0, the loss is the same - so can we just set alpha=0.0 by default and drop the conditional?

v0lta: Dear @the-xentropy, thank you for reviewing the code. JIT support is essential. Unfortunately, `alpha=0` and `alpha=None` are not the same. Therefore we can't drop the conditional, but we can use `jax.lax.cond`, [7368964](https://github.com/google-deepmind/optax/pull/599/commits/7368964be9b2713600870eb88d24f3d741a6ef9c) does this. I hope it addresses your concern.

optax/losses/_classification.py
```diff
@@ -430,3 +430,58 @@ def ctc_loss(
...
+  alpha = -1 if alpha is None else alpha
+
+  chex.assert_type([logits], float)
+  labels = labels.astype(logits.dtype)
+  # see also the original paper's implementation at:
+  # https://github.com/facebookresearch/fvcore/blob/main/fvcore/nn/focal_loss.py
+  p = jax.nn.sigmoid(logits)
+  ce_loss = sigmoid_binary_cross_entropy(logits, labels)
+  p_t = p * labels + (1 - p) * (1 - labels)
+  loss = ce_loss * ((1 - p_t) ** gamma)
```
fabianp, line 475: Thanks for the contribution @v0lta ! Our internal code review tool complains that the tests are not exhaustive enough. It rightfully points out that changing this line to ``p_t = (-(p * labels + (1 - p) * (1 - labels)))`` doesn't make any tests to fail.

Could you add other tests that would catch this (and potentially similar) failures?  

The original package has some more exhaustive tests from where you could get inspiration: https://github.com/facebookresearch/fvcore/blob/main/tests/test_focal_loss.py

optax/losses/_classification.py
fabianp: thanks for catching this @the-xentropy and @v0lta for the fix. Could you also add a test that JITs the function so it doesn't break in the future?

v0lta: Sure, I think `@chex.all_variants` takes care of jit, too. The problem was that the default value of alpha never changed. [74d17f0](https://github.com/google-deepmind/optax/pull/599/commits/74d17f0a76113d7dedd46f8e750064b4c0651c97) fixes this issue.

fabianp:
Great, thanks

fabianp:
Thanks for all the work @v0lta . I'm now just waiting for the internal review, which should be done within the next days

fabianp:
This has now been merged. Thanks @v0lta and apologies for the slow review on this one!

Note: I made some cosmetic changes when submitting due to internal errors. Feel free to open another pull request if something is missing!

PR #618, opened 2023-11-06
Use sphinx-gallery to display gallery of examples

fabianp:
As part of this change, I have:

  * Moved folder examples/ to doc/examples. This is unfortunately
    necessary if we want to have all examples in the same directory,
    as myst_nb does not allow notebooks to be outside of the docs/
    folder (although sphinx-gallery does allow python scripts to be
    outside of the docs/ folder).

  * Removed empty and unlinked file support.md

  * Wrote docs/examples/index.rst directory, with the code to display
    the gallery of examples.

  * I shortened the name of some examples to make them display nicer in the
    gallery.


mtthss:
Looks great!

PR #646, opened 2023-11-27
Update combine.py to use typing

Beomi:
- use `typing.Tuple` instead of `tuple` to previous python compability


fabianp:
We only support the Python versions that JAX supports. This means that we're no longer support Python 3.8 (https://jax.readthedocs.io/en/latest/changelog.html#jaxlib-0-4-14-july-27-2023)

PR #651, opened 2023-12-01
D-Adaptation and Prodigy contrib implementations

adefazio:
Implementations of D-Adaptation AdamW and the related method Prodigy based on the official PyTorch implementations. I have verified that they give the same outputs as the PyTorch version on an example problem. Unit tests similar to those used on other optimizers in contrib.

https://github.com/facebookresearch/dadaptation
https://github.com/konstmish/prodigy

These two new optimizers perform learning rate adaptation, similar to Mechanic and COCOB, two optimizers already included in contrib, but by a different mechanism, and so I think these are relevant to Optax and interesting to the community. D-Adaptation won an ICML outstanding paper award and is already gaining a lot of traction in the ML community, particularly for fine-tuning diffusion models with the Prodigy variant.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -776,6 +776,10 @@ scale_gradient
     SAMState
     cocob
     COCOBState
+    dadapt_adamw
+    DAdaptAdamWState
+    prodigy
+    ProdigyState
 
 
 Complex-Valued Optimization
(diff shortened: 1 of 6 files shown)


fabianp:
Thanks @adefazio  for the contribution!

vroulet (commented):
Thanks for the contribution!
Minor comment: if s and d could have more explicit names, that could be helpful for a newcomer to investigate the algorithm but numerous algorithms have non-readable parameters like b1 or beta so that's fine as is too.

vroulet (commented):
Looks perfect, thank you! Final request: could you squash your commits into one?

mtthss (approved):

PR #681, opened 2023-12-21
Add the SAM example to the example gallery

fabianp:
Along the way, I've also reorganized a bit the gallery to have all examples that make use of the contrib directory together.


holounic (approved):
Thank you!

Issue #479, opened 2023-02-01
Documentation Rendering


lockwo:
The API documentation has headings "common optimizers", "optax transformations", "apply updates", etc. However, the only thing that shows up on the API documentation sidebar in the read the docs is "common optimizers" (tested on chrome and safari). 

<img width="270" alt="Screen Shot 2023-02-01 at 2 26 06 PM" src="https://user-images.githubusercontent.com/42878312/216166879-53d6afeb-8b92-48c8-9c26-7019814938e6.png">

It seems like each of these headers should also be on the sidebar (since they are not under the concept of common optimizers and represent distinct things).

itstalmeez:
Hi,
I agree with your observation. Based on the screenshot you provided, it does seem like the API documentation sidebar is missing entries for "Optax Transformations," "Apply Updates," and potentially other headings.

Here are some steps you can take to address this:

Check for configuration errors:

If you have access to the Read the Docs configuration, double-check that the table of contents for the API section is set up correctly. There might be a setting that's causing only "Common Optimizers" to be displayed.
Report the issue:

If you don't have access to the configuration or can't find any issues there, consider reporting the problem to the Optax documentation maintainers. They'll be able to investigate and make the necessary corrections.
Workaround using search:

In the meantime, you can still access the content under those headings using the search functionality within the docs. However, having them directly in the sidebar would definitely improve navigation.
I'll be waiting for your positive response!!!!

MaanasArora:
Hi! This seems to be a limitation of Sphinx Book Theme. In line 21 of [this file](https://github.com/executablebooks/sphinx-book-theme/blob/1d067ae2101906cc89200897a887c2ddcf77af73/src/sphinx_book_theme/theme/sphinx_book_theme/components/sbt-sidebar-nav.html), which defines the navigation, `titles_only` (wrapping the `:titlesonly:` directive) is set to true.

Unfortunately, the only solution, saving to fork the theme, seems to be to divide `api.rst` into several files and reference all of them, which can get cumbersome. I don't believe the alternative of having a single "API Documentation" heading in the navigation will look very good?

Let me know how to proceed and I'd love to work on this issue. Thank you!

PR #688, opened 2023-12-29
Divide API documentation

MaanasArora:
Addresses issue #479.

I divided `api.rst` into one file for each section and referenced them individually from `index.rst`.

Thank you!


fabianp:
Thanks @MaanasArora ! This is great. 

Just one minor comment:

For some reason the contrib module appears different than the rest (module members appear in the left hand side while in the others they don't). Can we make the contrib module appear the same as the others?

image

MaanasArora:
Hi @fabianp thanks for the review! 

It seems that the source of the Contrib subtree was a `toctree` directive that generated stub for each `autosummary`, so I removed the nested toctrees. Let me know if this change is okay, thank you!

fabianp (approved):
Excellent, thanks! This looks good to me.

MaanasArora:
Thank you! Let me know if I need to do anything else for this to be merged.

fabianp:
Its now waiting for internal review. Should be done early next week. Thanks again for the contribution!

MaanasArora:
Got it, thanks a lot!

Issue #507, opened 2023-03-07
Add missing optimizers


carlosgmartin:
Add the following optimizers:
- [x] Nadam: [keras](https://keras.io/api/optimizers/Nadam/), [torch](https://pytorch.org/docs/stable/generated/torch.optim.NAdam.html). Pull request: https://github.com/google-deepmind/optax/pull/529
- [x] Adadelta: [keras](https://keras.io/api/optimizers/adadelta/), [torch](https://pytorch.org/docs/stable/generated/torch.optim.Adadelta.html)
- [x] Rprop: [torch](https://pytorch.org/docs/stable/generated/torch.optim.Rprop.html): Pull request: https://github.com/google-deepmind/optax/pull/695
- [x] Ftrl: [keras](https://keras.io/api/optimizers/ftrl/) : #738

SauravMaheshkar:
I'd like to work on the Adadelta implementation.

fabianp:
@SauravMaheshkar : that would be most welcome if you're still interested! Please open a new issue to keep track of that :-)

PR #695, opened 2024-01-08
Add RProp optimizer

MaanasArora:
Added the RProp optimizer as requested in #507.

Thank you!

### DIFF
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -887,3 +887,47 @@ def adamaxw(
       transform.add_decayed_weights(weight_decay, mask),
       transform.scale_by_learning_rate(learning_rate),
   )
+
+
+def rprop(
+    learning_rate: float,
+    eta_minus: float = 0.5,
+    eta_plus: float = 1.2,
+    min_step_size: float = 1e-6,
+    max_step_size: float = 50.0,
+) -> base.GradientTransformation:
+  """The Rprop optimizer.
+
+  Rprop, short for resillient backpropogation, is a first order variant of
+  gradient descent. It responds only to the sign of the gradient by increasing
+  or decreasing the step size selected per parameter exponentially to speed up
+  convergence and avoid oscillations.
+
+  References:
+    PyTorch implementation: 
+      https://pytorch.org/docs/stable/generated/torch.optim.Rprop.html
+    Riedmiller and Braun, 1993: https://ieeexplore.ieee.org/document/298623
+    Igel and Hüsken, 2003:
+      https://www.sciencedirect.com/science/article/abs/pii/S0925231201007007
+
+  Args:
+    learning_rate: The initial step size.
+    eta_minus: Multiplicative factor for decreasing step size. This is applied
+      when the gradient changes sign from one step to the next.
+    eta_plus: Multiplicative factor for increasing step size. This is applied
+      when the gradient has the same sign from one step to the next.
+    min_step_size: Minimum allowed step size. Smaller steps will be clipped to
+      this value.
+    max_step_size: Maximum allowed step size. Larger steps will be clipped to
+      this value.
+  """
+  return combine.chain(
+    transform.scale_by_rprop(
+      learning_rate=learning_rate,
+      eta_minus=eta_minus,
+      eta_plus=eta_plus,
+      min_step_size=min_step_size,
+      max_step_size=max_step_size,
+    ),
+    transform.scale(-1.0),
+  )
--- a/optax/_src/transform.py
+++ b/optax/_src/transform.py
@@ -748,6 +748,48 @@ def scale_by_radam(
   return base.GradientTransformation(init_fn, update_fn)
 
 
+class ScaleByRpropState(NamedTuple):
+  step_sizes: base.Updates
+  prev_updates: base.Updates
+
+
+def scale_by_rprop(
+    learning_rate: float,
+    eta_minus: float = 0.5,
+    eta_plus: float = 1.2,
+    min_step_size: float = 1e-6,
+    max_step_size: float = 50.0,
+) -> base.GradientTransformation:
+
+  def init_fn(params):
+    step_sizes = jax.tree_util.tree_map(
+      lambda p: learning_rate * jnp.ones_like(p), params)
+    prev_updates = jax.tree_util.tree_map(jnp.zeros_like, params)
+    return ScaleByRpropState(step_sizes, prev_updates)
+
+  def update_fn(updates, state, params=None):
+    del params
+    sign = jax.tree_util.tree_map(
+      lambda g, prev_g: g * prev_g, updates, state.prev_updates)
+    step_sizes = jax.tree_util.tree_map(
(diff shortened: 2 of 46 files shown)


vroulet:
Thank you @MaanasArora ! Could you add a docstring with a reference for the algorithm?

MaanasArora:
This is done, thanks!

optax/_src/alias.py
```diff
@@ -887,3 +887,40 @@ def adamaxw(
...
+    eta_minus: float = 0.5,
+    eta_plus: float = 1.2,
+    min_step_size: float = 1e-6,
+    max_step_size: float = 50.0,
+) -> base.GradientTransformation:
+  """The Rprop optimizer.
+
+  Rprop, short for resillient backpropogation, is a first order variant of
+  gradient descent. It responds only to the sign of the gradient by increasing
+  or decreasing the step size exponentially to speed up convergence and avoid
```
vroulet, line 903: Maybe add that the stepsize is selected per parameter

optax/_src/transform.py
```diff
@@ -748,6 +748,44 @@ def update_fn(updates, state, params=None):
   return base.GradientTransformation(init_fn, update_fn)
 
 
+class ScaleByRpropState(NamedTuple):
+  nu: base.Updates
```
vroulet, line 752: Can we call it stepsizes or lrs to better understand what the variable is?

optax/_src/transform.py
```diff
@@ -748,6 +748,44 @@ def update_fn(updates, state, params=None):
...
+    sign = jax.tree_util.tree_map(
+      lambda g, prev_g: g * prev_g, updates, state.prev_updates)
+    nu = jax.tree_util.tree_map(
+      lambda s, nu: jnp.where(
+        s == 0,
+        nu,
+        jnp.clip(nu * jnp.where(s < 0, eta_plus, eta_minus),
+                 a_min=min_step_size, a_max=max_step_size)),
+      sign, state.nu)
+    updates = jax.tree_util.tree_map(
```
vroulet, line 781: In the ref, they reverse the previous update if the sign flips no? And they also reset the previous updates to be zero in that case.

vroulet (commented):
Some additional comments :)

optax/_src/alias.py
```diff
@@ -887,3 +887,40 @@ def adamaxw(
...
+  References:
+    Riedmiller and Braun, 1993: https://ieeexplore.ieee.org/document/298623
+
+  Args:
+    learning_rate: The initial step size.
+    eta_minus: Multiplicative factor for decreasing step size. This is applied
+      when the gradient changes sign from one step to the next.
+    eta_plus: Multiplicative factor for increasing step size. This is applied
+      when the gradient has the same sign from one step to the next.
+    min_step_size: Minimum allowed step size. Larger steps will be clipped to
```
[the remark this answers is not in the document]
MaanasArora: Oops, thanks! Fixed

optax/_src/alias.py
```diff
@@ -887,3 +887,40 @@ def adamaxw(
...
+      this value.
+    max_step_size: Maximum allowed step size. Smaller steps will be clipped to
```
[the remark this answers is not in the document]
MaanasArora: Fixed

optax/_src/transform.py
MaanasArora: They do, I should have mentioned that I'm actually using the torch implementation as a reference; I thought it would be a more concrete baseline. Should I mention this in the docstring? Or is it better to stick to the paper?

optax/_src/transform.py
MaanasArora: Sure! This is done.

optax/_src/alias.py
vroulet: Perfect

optax/_src/transform.py
vroulet: Ah! Yep if you could simply add it in the doctring that would be good.
By curiosity, has there been recent investigations of the algorithm and such variant?

MaanasArora: Thanks! I added it. There have been some [further investigations](https://www.sciencedirect.com/science/article/abs/pii/S0925231201007007) but I'm not aware of anything very recent.

vroulet: Yes, just for some context. You may add that paper even if it's a bit outdated, that's still interesting. After that, if you can squash all commits I think we'll be good.

MaanasArora: I just added the paper and squashed. Thank you for the review!

vroulet (approved):

vroulet:
Hello @MaanasArora,
Sorry to ping you again. I'm merging it in the codebase, fixing some linting and I noted a potential mistake: in the updates of the stepsizes, I think it should be 
```
    step_sizes = jax.tree_util.tree_map(
        lambda s, step_size: jnp.where(
            s == 0,
            step_size,
            jnp.clip(
                step_size * jnp.where(s > 0, eta_plus, eta_minus),
                a_min=min_step_size, a_max=max_step_size)),
        sign, state.step_sizes
    )
 ```
 rather than
 ```
     step_sizes = jax.tree_util.tree_map(
        lambda s, step_size: jnp.where(
            s == 0,
            step_size,
            jnp.clip(
                step_size * jnp.where(s < 0, eta_plus, eta_minus),
                a_min=min_step_size, a_max=max_step_size)),
        sign, state.step_sizes
    )
```
But then the tests do not pass...

MaanasArora:
@vroulet No problem at all and thank you for the feedback! I agree that was a mistake, sorry for missing that. I think the issue with the correct setup is that Rprop is oscillating. I think we might need to manage the double-update case as mentioned in the paper. I will test if the pytorch implementation is stable as well. Thanks again!

MaanasArora:
This is actually quite strange. The version with `s < 0` seems to converge appropriately with a variety of hyperparameters.

MaanasArora:
Sorry, I think I found the mistake! The previous updates were being scaled by their negative, which flipped their sign. I instead moved the negative scale into `alias.py` which is also good for composability.

vroulet:
Thanks! Could you test quickly that this implementation returns exactly the same thing as the pytorch implementation? (no need to add such test in the code but just for the record as we're reviewing). If yes, you can squash everything and I'll take care of the merge.

MaanasArora:
Thank you for the suggestion! It was indeed important. I tried testing torch and our implementation on some random arrays, and they actually differed by 2e-1 in some places. It seems that the reversing double updates feature you mentioned was missing was indeed implemented in torch. After adding that feature, asserting an allclose with a tolerance of 1e-6 works properly. Thank you!

vroulet:
We needed to bump the sphynx version we used for the doc which made the tests fail.
Can you sync with the main?

MaanasArora:
This is done and squashed!

vroulet:
Merged in 108bf1e4cb00a7d2b415035d57391270b6bc5a35. Thanks again @MaanasArora !

PR #714, opened 2024-01-16
Test on Python 3.12

fabianp:


fabianp:
This fails since tensorflow doesn't (yet) have a package for Python 3.12

PR #567, opened 2023-08-13
Add the Nadam optimizer

Rupt:
Nadam is Adam with Nesterov's correction. It is recommended by the [Deep Learning Tuning Playbook](https://github.com/google-research/tuning_playbook#choosing-the-optimizer).

This PR adapts @AntoinePlumerault's excellent work in https://github.com/deepmind/optax/pull/529 to
- Pass the tests, by tree mapping, safe incrementing, and reducing `learning_rate` in `alias_test`.
- Not include an extra `b2` scale factor on the second moment estimate.
  This multiplier only appears in the [2016 OpenReview paper](https://openreview.net/forum?id=OM0jvwB8jIp57ZJjtNEZ), and I believe it might be a typographic error because I see no explanation. [Other](https://github.com/keras-team/keras/blob/b3ffea6602dbbb481e82312baa24fe657de83e11/keras/optimizers/nadam.py#L174) [versions](https://pytorch.org/docs/stable/generated/torch.optim.NAdam.html) [also](https://arxiv.org/abs/1910.05446) [exclude](https://cs229.stanford.edu/proj2015/054_report.pdf) it.
- Document the mathematical form we implement.
  The versions linked above differ by their "bias correction" factors go. This implementation matches [Torch](https://pytorch.org/docs/stable/generated/torch.optim.NAdam.html) and [Keras](https://github.com/keras-team/keras/blob/b3ffea6602dbbb481e82312baa24fe657de83e11/keras/optimizers/nadam.py#L171) (for constant `b1`).

In this PR, we:
- Add the Nadam transform and optimizer alias.
- Make minor tidy-ups: remove some inconsistent whitespace, place `scale_by_lion` alphabetically.

Merges #570 so that tests pass.

Ticks one off #507

PR #529, opened 2023-05-13
Added Nadam optimizer

AntoinePlumerault:
Followed implementation descscribed in Algorithm 2 (https://openreview.net/pdf?id=OM0jvwB8jIp57ZJjtNEZ). https://github.com/deepmind/optax/issues/507

PR #719, opened 2024-01-18
Add the Nadam optimizer

Rupt:
_Mimics #567, because I deleted its original fork._

Nadam is Adam with Nesterov's correction. It is recommended by the [Deep Learning Tuning Playbook](https://github.com/google-research/tuning_playbook#choosing-the-optimizer).

In this PR, we add the Nadam transform and optimizer alias.

This PR adapts @AntoinePlumerault's excellent work in #529 to
- Pass the tests, by tree mapping, safe incrementing, and reducing `learning_rate` in `alias_test`.
- Exclude an extra `b2` scale factor on the second moment estimate.
  This multiplier only appears in the [2016 OpenReview paper](https://openreview.net/forum?id=OM0jvwB8jIp57ZJjtNEZ) (Algorithm 2), and I suspect that it might be a typo because I see no explanation. [Other](https://github.com/keras-team/keras/blob/b3ffea6602dbbb481e82312baa24fe657de83e11/keras/optimizers/nadam.py#L174) [versions](https://pytorch.org/docs/stable/generated/torch.optim.NAdam.html) [also](https://arxiv.org/abs/1910.05446) [exclude](https://cs229.stanford.edu/proj2015/054_report.pdf) it.
- Document the mathematical form we implement.
  The versions linked above differ by their "bias correction" factors. This implementation matches [Torch](https://pytorch.org/docs/stable/generated/torch.optim.NAdam.html) and [Keras](https://github.com/keras-team/keras/blob/b3ffea6602dbbb481e82312baa24fe657de83e11/keras/optimizers/nadam.py#L171) (for constant `b1`).

Ticks one off #507

### DIFF
--- a/docs/api/common_optimizers.rst
+++ b/docs/api/common_optimizers.rst
@@ -16,6 +16,7 @@ Common Optimizers
     lamb
     lars
     lion
+    nadam
     noisy_sgd
     novograd
     optimistic_gradient_descent
@@ -79,6 +80,10 @@ SM3
 ~~~
 .. autofunction:: sm3
 
+Nadam
+~~~~
+
+.. autofunction:: nadam
 
 Noisy SGD
 ~~~~~~~~~
--- a/docs/api/optax_transformations.rst
+++ b/docs/api/optax_transformations.rst
@@ -39,6 +39,7 @@ Optax Transformations
     scale_by_belief
     scale_by_factored_rms
     scale_by_lion
+    scale_by_nadam
     scale_by_novograd
     scale_by_optimistic_gradient
     scale_by_param_block_norm
@@ -173,6 +174,8 @@ Optax transformations and states
 .. autofunction:: scale_by_lion
 .. autoclass:: ScaleByLionState
     :members:
+
+.. autofunction:: scale_by_nadam
 
 .. autofunction:: scale_by_novograd
 .. autoclass:: ScaleByNovogradState
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -244,6 +244,48 @@ def adam(
       transform.scale_by_learning_rate(learning_rate),
   )
 
+def nadam(
+   learning_rate: base.ScalarOrSchedule,
+    b1: float = 0.9,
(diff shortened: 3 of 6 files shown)


fabianp (approved):

vroulet:
Thank you very much @Rupt !
Quick question: did you try to compare the outputs of this implementation and the pytroch one to see if they match?

Rupt:
> Thank you very much @Rupt ! Quick question: did you try to compare the outputs of this implementation and the pytroch one to see if they match?

No, I'm afraid I don't have any numerical valuation to share. But I strongly agree that more checks would be useful, so I'm happy to add some.

vroulet:
No worries! I think it would be for another PR to think about other tests.
For now, it passes the standard tests, so it should be good to go.

Rupt:
I can't access details on the `feedback/copybara — Google internal checks FAILED for runs with create time 2024-01-18T21:16:46.079` failure.

vroulet:
Yes, we are investigating that, it's not clear yet why some errors are triggered.

fabianp:
hey @Rupt , yes, we have some internal failures that we're looking into.

Another thing that we're exploring is to implement this solver as a keyword argument to adam (and perhaps also exposing a nadam solver for convenience). The advantage of this is that we could add a `nesterov` kwarg for free to similar solvers like adamw. I took a stab at implementing this based on your code here (https://github.com/google-deepmind/optax/pull/723). What would you think about this solution?

Rupt:
> Another thing that we're exploring is to implement this solver as a keyword argument to adam (and perhaps also exposing a nadam solver for convenience). The advantage of this is that we could add a `nesterov` kwarg for free to similar solvers like adamw. I took a stab at implementing this based on your code here (#723). What would you think about this solution?

Beautiful. That removes all of this duplicated-but-must-be-identical logic, and you've written nicer documentation too. I've left [one tiny suggestion](https://github.com/google-deepmind/optax/pull/723/files#r1459362802) for a character that escaped.

Rupt:
Closing in favour of #723

PR #737, opened 2024-01-25
feat(ci): drop `setup.py` from publishing CI

SauravMaheshkar:
## Context

This PR aims to continue the work done in #513 to remove `setup.py` and migrate to a pure `pyproject.toml` based build. While the PR was merged in (https://github.com/google-deepmind/optax/pull/513#issuecomment-1551532728) the older legacy files such as `setup.py` and `MANIFEST.in` still exist. 

The comment to delete such legacy files was raised again https://github.com/google-deepmind/optax/pull/513#issuecomment-1902488225 and thus the issue to drop `setup.py` from CI raised in https://github.com/google-deepmind/optax/pull/513#issuecomment-1902670911.

---

Changes made my this PR can be summarised as follows:
* switch from dynamic to statically defined version metadata in `pyproject.toml`.
* Modify publishing workflow to read version metadata from `pyproject.toml`.

Request for Review: @fabianp

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -16,11 +16,12 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install setuptools wheel twine build
+        pip install setuptools wheel twine build toml
     - name: Check consistency between the package version and release tag
       run: |
         RELEASE_VER=${GITHUB_REF#refs/*/}
-        PACKAGE_VER="v`python setup.py --version`"
+        VER = python -c "import toml; PYPROJECT = toml.load('pyproject.toml'); print(PYPROJECT['project']['version'])"
+        PACKAGE_VER="v`VER`"
         if [ $RELEASE_VER != $PACKAGE_VER ]
         then
           echo "package ver. ($PACKAGE_VER) != release ver. ($RELEASE_VER)"; exit 1
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -16,11 +16,12 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install setuptools wheel twine build toml
+        pip install setuptools wheel twine build
     - name: Check consistency between the package version and release tag
       run: |
+        pip install .
         RELEASE_VER=${GITHUB_REF#refs/*/}
-        VER = python -c "import toml; PYPROJECT = toml.load('pyproject.toml'); print(PYPROJECT['project']['version'])"
+        VER = python -c "import optax; print(optax.__version__)"
         PACKAGE_VER="v`VER`"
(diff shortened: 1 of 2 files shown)


fabianp:
thanks @SauravMaheshkar . This has the big disadvantage that the versions are now duplicated in both optax/__init__.py and in the pyproject.toml . How about getting the version with `python -c "import optax; print(optax.__version__)"`? Would something like that work?

SauravMaheshkar:
> thanks @SauravMaheshkar . This has the big disadvantage that the versions are now duplicated in both optax/**init**.py and in the pyproject.toml . How about getting the version with `python -c "import optax; print(optax.__version__)"`? Would something like that work?

Yes that would work, however that would require for us to do a editable install in that step i.e. a `pip install -e .`

Although with some cache I'm sure even that could be sped up. Would you recommend I proceed with the proposed fix ?

fabianp:
do we really need to install optax for that? I think as long as we run it from the optax root directory it should work

SauravMaheshkar:
We would run into import error's right ? Following is my terminal output for running `python -c "import optax; print(optax.__version__)"`

```bash
❯ python -c "import optax; print(optax.__version__)"
  File "<string>", line 1, in <module>
  File "/Users/saurav/github/SauravMaheshkar/optax/optax/__init__.py", line 17, in <module>
    from optax import contrib
  File "/Users/saurav/github/SauravMaheshkar/optax/optax/contrib/__init__.py", line 17, in <module>
    from optax.contrib.cocob import cocob
  File "/Users/saurav/github/SauravMaheshkar/optax/optax/contrib/cocob.py", line 23, in <module>
    import jax.numpy as jnp
ModuleNotFoundError: No module named 'jax'
```

fabianp:
works for me, it seems you're just missing some dependencies

SauravMaheshkar:
The python environment in the Github Actions workflow won't have `jax` installed, right ? It only installs setuptools, wheel, twine and build in the previous step.

fabianp:
yes, you're absolutely right. I was thinking we could run `pip install .` right before `python -c "import optax; print(optax.__version__)"`

fabianp:
which I see is what you proposed a few comments earlier 🙃 . Anyway, we seem to be on the same page 😄

SauravMaheshkar:
Sounds good, will revert the change to `pyproject.toml` and do a editable install before fetching the version.

fabianp:
My point before is that we don't really need to install optax, just the dependencies. Not sure if there's an easy way to do that, and TBH doesn't change much

fabianp:
I thought `pip install . ` does that, install the dependencies and not the package, but I could be wrong

fabianp:
> Sounds good, will revert the change to `pyproject.toml` and do a editable install before fetching the version.

Thanks! Let me know when I should take another look at it

SauravMaheshkar:
CI looks green. Should be good now @fabianp

fabianp (approved):

fabianp:
Thanks! Looks good to me

PR #738, opened 2024-01-26
Implementation of FTRL Optimizer

amosyou:
Adds the FTRL optimizer as requested in #507 according to [this paper](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/41159.pdf).


amosyou:
@fabianp I've read through the paper as well as the Keras implementation in the past couple days, I feel like I have a good understanding of the algorithm now. However, I'm a little stuck on the implementation due to the closed form update rule.

The paper gives the a update rule for $w_{t+1}$, but it's a function of some other saved variables in the optimizer state (ie. z and n), not in terms of $w_{t}$. The Keras implementation has an update function that has the gradients and parameters as inputs, so it's not an issue for them to use the closed form update rule to find the new parameters. In our case, I was thinking of just finding $w_{t}$, updating the saved variables, calculating $w_{t+1}$, and then taking the difference between $w_{t+1} - w_{t}$ as our gradient transformation. It's a bit round-about (and defeats the purpose of some of the findings of the paper), but this was the only solution that came to mind.

Let me know what your thoughts are! Would appreciate some input on this

fabianp:
hey @amosyou , thanks for starting this!

> The paper gives the a update rule for wt+1, but it's a function of some other saved variables in the optimizer state (ie. z and n), not in terms of wt. The Keras implementation has an update function that has the gradients and parameters as inputs, so it's not an issue for them to use the closed form update rule to find the new parameters. In our case, I was thinking of just finding wt, updating the saved variables, calculating wt+1, and then taking the difference between wt+1−wt as our gradient transformation. It's a bit round-about (and defeats the purpose of some of the findings of the paper), but this was the only solution that came to mind.

Yes, I would do as you suggest. I don't see an easy way around this. Unfortunately optax has the implicit assumption that optimizer updates are additive, but this is not always the case. I anyway think that the overhead of computing  wt+1−wt is negligeable compared to the cost of computing the gradient, so I don't think this is a big deal.

> 
> Let me know what your thoughts are! Would appreciate some input on this

A few of high level comments on the code:

1. This is a very classical and highly cited solver, so let's put in the main package, not in contrib (we'll soon be adding some guidelines to the doc on what should go into contrib vs the main package). In practice what this means is that the ftrl function would go (probably) under optax/_src/alias.py (and potentially also optax/_src/transform.py if it makes sense to factor it as other optimizers) and that the solver should pass the standard tests in alias_test.py (and potentially also transform_test.py)
2. It would be nice if you can put the pseudocode of the algorithm in the docstring, as we did for the `adam()` algorithm.
3. See also the docstring of `adam()` to see how to format the references so that they appear nicely in the website.

Thanks again, this is a very valuable contribution!

amosyou:
thanks for your suggestions! 

> This is a very classical and highly cited solver, so let's put in the main package, not in contrib (we'll soon be adding some guidelines to the doc on what should go into contrib vs the main package). In practice what this means is that the ftrl function would go (probably) under optax/_src/alias.py (and potentially also optax/_src/transform.py if it makes sense to factor it as other optimizers) and that the solver should pass the standard tests in alias_test.py (and potentially also transform_test.py)

I've moved the files from contrib to _src!

> It would be nice if you can put the pseudocode of the algorithm in the docstring, as we did for the adam() algorithm.

Sounds good, I'll add an explanation for how we implemented the w_t+1 - w_t.

I'm currently failing 1 test case for alias_test.py which is either due to an error in my implementation or I haven't found the right hyperparameters for the test case (will investigate this more). I'm also missing a few cases for transform_test.py which is due to the fact that the default parameter values lead to a divide by 0 error. Maybe I could just set a default beta value that is positive which fixes the divide by 0 error?

amosyou:
I found the appropriate hyperparameters for the test cases in alias_test.py and transform_test.py. ready for review!

fabianp:
Amazing! What about the division by zero? No longer an issue?

amosyou:
fixed! I set the default beta=1 instead of 0. (paper also suggests this too, not sure why Keras used 0)

optax/_src/transform.py
```diff
@@ -1373,3 +1376,85 @@ def _tx(g, d, g_sos):
...
+  initial_accumulator_value: float = 0,
+  lambda_1: float = 0,
+  lambda_2: float = 0,
+  beta: float = 1,
+) -> base.GradientTransformation:
+  # pylint: disable=line-too-long
+  """Scale updates according to FTRL algorithm.
+  
+  References:
+    [McMahan et al](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/41159.pdf)
```
fabianp, line 1398: same formatting comments as in the alias.py file

optax/_src/alias.py
```diff
@@ -1082,12 +1082,73 @@ def rprop(
...
+  
+  FTRL, or Follow the Regularized Leader, is an optimization algorithm developed
+  at Google for click-through rate prediction in the early 2010s. It is most
+  suitable for shallow models with large and sparse feature spaces.
+  
+  McMahan et al. gives a closed form update for the parameters as,
+  
+  .. math::
+  
+    \begin{align*}
```
fabianp, line 1114: use & to align equations:

```
    \begin{align*}
      w_{t,i} &= \begin{cases} 
        0 & \text{if } |z_i| \leq \lambda_1 \\
        -\left( \frac{\beta + \sqrt{n_i}}{\alpha} + \lambda_2 \right)^{-1} 
          \left( z_i - \text{sgn}(z_i)\lambda_1 \right) & \text{otherwise}.
      \end{cases} \\
      \sigma_i &= \frac{1}{\alpha} \left( \sqrt{n_i + g_i^2} - \sqrt{n_i} \right) \\
      z_i &\leftarrow z_i + g_i - \sigma_i w_{t,i} \\
      n_i &\leftarrow n_i + g_i^2
    \end{align*}
```

 this way it looks like 
image
instead of 
image

fabianp (changes requested):
Thanks for the contribution! Some minor comments on the doc for now

fabianp:
One comment regarding citation: while the keras implementation references the McMahan 2013 paper, reading through that paper and related literature, it seems to me that the algorithm was initially proposed in https://arxiv.org/pdf/1002.4908.pdf , which appears 3 years prior. I suggest we mention both papers, what do you think?

amosyou:
the initial proposal of the algorithm in 2010 seems right according to this [paper](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/37013.pdf), good call on including it too! on a separate note, perhaps we should make it clear that this is the FTRL-Proximal algorithm? not the most well-versed with the literature but it seems that FTRL encompasses many algorithms (eg. dual averaging according to [here](https://arxiv.org/pdf/1002.4908.pdf))?

fabianp:
Absolutely @amosyou ! I agree with writing in the docstring that this is the  FTRL-Proximal algorithm, but I would keep the name of ftrl for consistency with keras (and because its shorter 😉 )

konstmish:
Hi @amosyou, I had a discussion with @fabianp and @mblondel about FTRL, and the plan is to put this pull request on hold for now. The reason is that after digging through the papers on FTRL, I concluded that the current implementation is not going to be useful because of the problem that you mention with having to compute $w_{t+1}-w_t$. The history of FTRL and more details are below:
1. Firstly, I want to point out a good description of FTRL in the [book by Orabona](https://arxiv.org/pdf/1912.13213.pdf). You can see that the method is defined there (see Algorithm 7.1) as $w_{t+1}= \arg \min_{w} \left( \psi_t(w) + \sum\limits^{t-1}_{i=1} \ell_t(w) \right)$. Here $\psi_t$ is an almost arbitrary convex regularizer. The method is mostly theoretical, but it is quite general and some of its special cases are implementable.
2. The original reference given in the book is Abernethy et al. [2008], "Competing in the dark: An efficient algorithm for bandit linear optimization". The 2013 paper is more cited, but I'm not sure why it is the case. Orabona in the book says 

> There is an interesting bit about FTRL-Proximal [McMahan, 2011]: FTRL-Proximal is an instantiation of FTRL
that uses a particular proximal regularizer. It became very famous in internet companies when Google disclosed in
a very influential paper that they were using FTRL-Proximal to train the classifier for click prediction [McMahan et al., 2013]. This generated even more confusion because many people started conflating the term FTRL-Proximal (a
specific algorithm) with FTRL (an entire family of algorithms).

3. The Keras implementation seems to use linearized losses $\ell_t$ and Adagrad-based stepsizes, which are in line with your implementation and the 2013 paper. However, it is just a special case of FTRL, and shouldn't have had this name. Keras implementation also uses the explicit update rule because it doesn't need $w_{t+1} - w_t$.
4. Unlike Keras, Optax can't handle the explicit update, which, according to 2013 paper, is what makes it useful (saving time thanks to the sparsity of the updates). Optax, in contrast, would benefit much more from Proximal SGD and Proximal Adagrad, since they explicitly compute $w_{t+1}$ from $w_t$. Such Prox-based methods should be added to Optax at some point in the future, and they should support $\ell_1$ and $\ell_2$ regularization, among other regularizers.

Issue #507 mentioned FTRL probably because of how many citations it had, but I think that adding the special case of FTRL would only introduce confusion. The implementation wouldn't be as efficient as the 2013 paper claims, and it wouldn't be as general as proposed in 2008 paper.

amosyou:
Hi @konstmish, thanks for your detailed response! The excerpts from Orabona's book are a great read, really informative and points out an interesting point of how people conflate FTRL-Proximal for FTRL (literally me earlier in this PR)!

On your points about not including this implementation for now, totally agree with your concerns about efficiency and exacerbating the confusion surrounding FTRL. It seems that it was a fundamental design choice for Optax to use gradient transformations that are additive, but if for some reason in the future Optax is restructured to handle explicit updates, we should be able to merge this in with minimal modifications.

Thanks for taking the time to discuss this with @fabianp @mblondel :)

In terms of next steps, I can open new issues for Proximal SGD and Proximal Adagrad and try implementing those as well.

fabianp:
Thanks @amosyou for the understanding! I feel bad that at the end we're not merging this code, but I do think it was necessary to have the implementation to fully realize all the pros and cons.

As for the next steps, I would wait a bit to implement proximal SGD as we'd like to have something generic, similar to jaxopt, where the user could pass any projection or proximal operator (https://jaxopt.github.io/stable/constrained.html), and that requires thinking a bit more about the API.

But proximal Adagrad I think should be implemented just for the L1 + L2 penalties, as it will not be efficient for general proxes due to the diagonal rescaling. So I think that something (API-wise) similar to [this TF1 implementation](https://www.tensorflow.org/api_docs/python/tf/compat/v1/train/ProximalAdagradOptimizer) would be a welcome addition! 

What do others think? CC @mblondel @konstmish @vroulet ?

konstmish:
I agree, Adagrad with support for $\ell_1 + \ell_2$ regularization should achieve exactly what FTRL from Keras does.

fabianp:
Closing this one as per the discussion above

PR #743, opened 2024-01-29
downgrade pytest to 7.4.4

fabianp:


fabianp:
seems to be related to https://github.com/pytest-dev/pytest/issues/11807

fabianp:
I think a better solution that doesn't require to downgrade pypi is https://github.com/google-deepmind/optax/pull/746

fabianp:
closing in favor of https://github.com/google-deepmind/optax/pull/746

Issue #656, opened 2023-12-05
remove tensorflow as dependency for the examples


fabianp:
Currently we're importing tensorflow for some examples such as in https://optax.readthedocs.io/en/latest/_collections/examples/cifar10_resnet.html. 

It should be possible to avoid this import and instead rely exclusively on tfds and (if necessary [grain](https://github.com/google/grain))

See for example https://www.tensorflow.org/datasets/tfless_tfds on how to import tfds datasets without tensorflow.

mmhamdy:
Anyone working on this? if not, I'd be happy to work on it.

fabianp:
All yours @mmhamdy !

fabianp:
hey @mmhamdy , any progress on this? I might have some cycles to work on this issue next week if it's not yet solved

mmhamdy:
Hi @fabianp, sorry for taking that long, January with such a MONTHFUL😅 for me! But anyway, I'll submit a PR in a couple of days.

I was leaving `grain` as a last resort and trying to use TFDS only. The thing is, the promise of a **TFless TFDS** in the documentation isn't a complete one, quoting the doc:

> In future versions, we are also going to make the dataset preparation TensorFlow-free.

So, we are getting a TF-free dataset but we still need TF to process it. That's nice, but still not that helpful.

I used `grain` in the end, as you might've expected. It's working fine, just a small problem with MNIST. When loaded using the ArrayRecord-based data source, we get a chunk header hash mismatch and can't access it. I've opened an issue and still working on it.

The other datasets are fine. Now, the examples notebooks that use TF and TFDS are:
```
adversarial_training.ipynb
lookahead_mnist.ipynb
mlp_mnist.ipynb
differentially_private_sgd.ipynb
cifar10_resnet.ipynb
reduce_on_plateau.ipynb
```
The first 4 examples use MNIST, and `cifar10_resnet.ipynb` still uses TF for image augmentation.
Also, a small note, `mlp_mnist.ipynb` and `differentially_private_sgd.ipynb` appear on the gallery but don't appear on the side menu.

So, I think I'm going to start with `reduce_on_plateau.ipynb`. One last question, do you prefer a separate PR for each notebook or just one PR for all?

Thank you, and sorry for the long reply. Was just thinking out loud.

fabianp:
Excellent, and thanks for the update! I'm trying to push for this as I'd like to test on Python 3.12 but unfortunately TF is blocking us from doing so 😞 

[...]

PR #768, opened 2024-02-04
Remove TF dependency in reduce_on_plateau example

mmhamdy:
This PR removes Tensorflow as a dependency in `reduce_on_plateau.ipynb`, and uses only TFDS, with [grain](https://github.com/google/grain) for loading and preparing the dataset.

Addressing issue #656


fabianp (changes requested):
This looks amazing! Way to go @mmhamdy! 🚀 

Just one thing, could you please remove the %pip install statements, and instead add these dependencies to the [pyproject.toml](https://github.com/google-deepmind/optax/blob/main/pyproject.toml) file, under section [docs].

mmhamdy:
Sure, much more cleaner this way.

mmhamdy:
Hey @fabianp, looks like there is a problem with python 3.9 and grain.

fabianp:
ouch! yeah, I'm seeing from https://github.com/google/grain/blob/main/pyproject.toml that they require Python 3.10 ....

mmhamdy:
ouch indeed! 😅 What do you think? should we stick to notebook downloads until we get rid of either TensorFlow or Python 3.9?

fabianp:
I'm unsure. Let me talk today with the other optax devs to see what they
think

mmhamdy:
Sounds great!

fabianp:
BTW I submitted a PR fixing python 3.9 errors: https://github.com/google/grain/pull/338

mmhamdy:
That's QUITE a PR! 😅😅

fabianp:
yeah, one thing that worries me a bit more about pygrain is that it it doesn't seem to build on OSX (tried and failed), and probably also Windows (haven't tried there). Before committing replacing TF with pygrain it would be important to make sure that those platforms are supported (just to say that it might take a bit before we merge these PRs unfortunately)

mmhamdy:
Well, replacing TensorFlow wasn't expected to be a walk in the park anyway 😃. But we've come a long way and I see that the people at pygrain are working daily on it.

vroulet:
For the record, in #1270 we removed this notebook (and others) from the docs building [configuration](https://github.com/google-deepmind/optax/blob/main/docs/conf.py) so we don't depend on tensorflow anymore during the continuous integration. 
This PR is still great to actually migrate the examples to grain for the sake of users learning how to use grain etc... And by the way, grain now supports mac os (not windows I believe but that's fine). So we could try merging this @mmhamdy if you want

mmhamdy:
hi @vroulet, yeah sure that would be great!

rajasekharporeddy:
Closing this PR since the TensorFlow dependency issue in `reduce_on_plateau` example is resolved with the PR #1616.

PR #780, opened 2024-02-07
Remove TF dependency in adversarial_training example

mmhamdy:
This PR removes Tensorflow as a dependency in `adversarial_training.ipynb`, and uses only TFDS, with [grain](https://github.com/google/grain) for loading and preparing the dataset.

Addressing issue #656


fabianp:
@mmhamdy : now that grain has builds for 3.9, can you update with master to see if the tests pass?

mmhamdy:
Great! But I think we need to start with #768 in which I've added grain as doc dependency. I've updated it and it has passed all tests successfully 🚀

rajasekharporeddy:
Closing this PR since the TensorFlow dependency issue in adversarial_training example is resolved with the PR #1609.

PR #785, opened 2024-02-08
correct type annotation for `params`

yixiaoer:
Correct type annotation for params in the `net` function in [docs/optax-101.ipynb](https://github.com/google-deepmind/optax/blob/54c238c2ca16f1ffedcfe8dc8e5a00a72d97e300/docs/optax-101.ipynb#L309)


fabianp (approved):
thanks!

PR #792, opened 2024-02-09
Add missing examples to contrib examples section.

fabianp:
Also add flax as dependency to examples, which should fix the error in

https://optax.readthedocs.io/en/latest/_collections/examples/mlp_mnist.html

### DIFF
--- a/examples/contrib/README.md
+++ b/examples/contrib/README.md
@@ -2,7 +2,6 @@
 
 Examples that make use of the `optax.contrib` module.
 
-<!-- include all files in the current directory except README.md -->
 ```{toctree}
 :glob:
 :maxdepth: 1
--- a/examples/contrib/README.md
+++ b/examples/contrib/README.md
@@ -7,5 +7,5 @@ Examples that make use of the `optax.contrib` module.
 :glob:
 :maxdepth: 1
 
-[!README.md]*
+*
 ```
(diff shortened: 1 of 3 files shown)


examples/contrib/README.md
```diff
@@ -7,5 +7,5 @@ Examples that make use of the `optax.contrib` module.
 :glob:
 :maxdepth: 1
 
-[!README.md]*
+*
```
vroulet, line 10: Remove then the comment above

pyproject.toml
```diff
@@ -38,8 +38,8 @@ dependencies = [
 ]
 
 [project.urls]
-homepage = "https://github.com/deepmind/optax"
-repository = "https://github.com/deepmind/optax"
+homepage = "https://github.com/google-deepmind/optax"
```
vroulet, line 41: As we are on it, shouldn't we change the authors as 
authors = [
    {name = "Google DeepMind", email = "optax-dev@google.com"},
]
?

fabianp: indeed, done!

examples/contrib/README.md
fabianp: good catch!

fabianp:
done, please take another look :-)

vroulet (approved):

Issue #669, opened 2023-12-10
NamedTuples not being documented properly


fabianp:
Note the redundancy and all the "Alias for field number X" in classes such as optax.MultiStepsState: https://optax.readthedocs.io/en/latest/api/optimizer_wrappers.html#optax.MultiStepsState

![image](https://github.com/google-deepmind/optax/assets/277639/3c755c6e-4f3b-45b6-bb58-e6e3e9be2ead)

amosyou:
I took a look at using [sphinx-toolbox](https://sphinx-toolbox.readthedocs.io/en/latest/extensions/more_autodoc/autonamedtuple.html) to autodoc the namedtuples, but I wasn't able to resolve the duplicate fields showing up. It seems that this is a known issue according to [here](https://github.com/sphinx-doc/sphinx/issues/8717), not too sure how else to go about fixing this.

fabianp:
Thanks for looking into it @amosyou . Eliminating the "Alias for field ___" would already be better, even if we loose the type information IMO. Could you post a screenshot of how that would look like?

amosyou:
<img width="831" alt="Screenshot 2024-02-03 at 12 48 46 AM" src="https://github.com/google-deepmind/optax/assets/91300605/ceda0433-9046-48cd-b21a-8e8b9ee8dc00">

fabianp:
I think that looks good! (module some formatting issues that I thin are orthogonal to this issue). Could you submit a pull request for this?

You're on 🔥 @amosyou !

amosyou:
revisiting this! so removing :members: from the autoclass directives isn't a viable solution since it also removes fields that are accurately documented like for LookaheadState

<img width="824" alt="Screenshot 2024-02-14 at 11 19 07 PM" src="https://github.com/google-deepmind/optax/assets/91300605/e1a9a3cf-4113-4989-9636-7c23f3a1715b">

i've tried using the autodoc-skip-member event as suggested [here](https://stackoverflow.com/questions/61572220/python-sphinx-namedtuple-documentation) but it doesn't seem to be removing the duplicate alias fields. i've also tried modifying the _monkey_patch_doc_strings in conf.py to replace lines with "Alias for field number" with empty string, also didn't work. not sure where to go from here 😅

vroulet:
Thank you @amosyou for looking into this!
I've just tried removing :members: from the LookAheadState and the `fast_state` and `steps_since_sync` were still in the doc. One can even still have types if we are really brave:
```
[...]

PR #803, opened 2024-02-15
Fix NamedTuples Formatting

amosyou:
fixes duplicate formatting for classes inheriting from NamedTuples as seen in #669


amosyou:
@vroulet most of the classes / states in transform.py and combine.py don't have proper docstrings. would it still be worthwhile to remove :members: for those files (many of those classes won't have much rendered bc of no docstring)

vroulet:
Hello @amosyou (sorry got too many mails didn't catch this notification). It's better not to have a docstring than having a confusing one. So yes, you can remove :members: to all classes and we may fill all the docstrings later (one file at a time probably).

amosyou:
@vroulet i removed the :members: from all the states with the exception of SAMState in sam.py since SAMState uses the chex.dataclass decorator instead of inheriting from NamedTuple. for consistency should we change that to be a namedtuple?

vroulet:
That's great, it will really make the doc more readable! 
For SAM's State it should have been a NamedTuple from the start... Pytypes could have checked that but the author disabled pytype in that file... I'm going to reach out to the author and ask why. For now, let's (unfortunately) keep it as is.

optax/_src/constrain.py
```diff
@@ -57,13 +57,15 @@ def update_fn(updates, state, params):
...
-  Each leaf is a single boolean which contains True iff a NaN was detected in
-  the corresponding parameter array at the last call to `update`.
+  Attributes:
+    found_nan (``jax.Array``): tree that has the same structure as that of the
+      parameters. Each leaf is a single boolean which contains True iff a NaN
+      was detected in the corresponding parameter array at the last call to
+      `update`.
   """
-  found_nan: Any
+  found_nan: jax.Array
```
vroulet, line 68: Should rather be a chex.ArrayTree (it can be a tree not just an Array)

optax/_src/wrappers.py
```diff
@@ -175,23 +174,24 @@ def _zeros_tree_like(inp_tree: chex.ArrayTree) -> chex.ArrayTree:
...
       `ShouldSkipUpdateFunction`.
   """
-  mini_step: Array
-  gradient_step: Array
-  inner_opt_state: Any
-  acc_grads: Any
+  mini_step: Union[jax.Array, int]
+  gradient_step: Union[jax.Array, int]
+  inner_opt_state: base.OptState
+  acc_grads: jax.Array # TODO: double check this one
```
vroulet, line 194: Should be a base.Updates I think

docs/api/optimizer_wrappers.rst
vroulet, line 1: Since you made a pass on all States here to have a nice formatting you can keep the members attribute. 
My idea was: 
- if you're brave and you change the docstring such that it's properly formatted, then let's keep `:members:` for that state.
- otherwise, simply remove `:members:` and we may take care of the formatting of the states later.

vroulet (commented):
First few comments, I'm going to build the doc and see for the pytypes.

vroulet:
Could you also merge with main to see latest changes? (for example I'm not sure why States such as ClipState are displayed as alias of tuple instead of alias of EmptyState)

docs/api/combining_optimizers.rst
```diff
@@ -15,4 +15,3 @@ Multi-transform
 ~~~~~~~~~~~~~~~
 .. autofunction:: multi_transform
 .. autoclass::  MultiTransformState
```
vroulet, line 17: Add a one line docstring to MultiTransformState and add MultiTransformState to the autosummary

docs/api/optimizer_wrappers.rst
vroulet: No sorry scratch that, what you did is perfect. Then simply remove :members: for the LookaheadParams

amosyou:
@vroulet apologies on the delay! been quite a month for me but we're back. ready for another review :)

docs/api/optimizer_wrappers.rst
amosyou: sounds good! i brought back the members

vroulet (commented):
Thank you very much @amosyou. It's close to be good!
I have (hopefully last) comments:
- Remove `:members:` from SAMState
- Remove `:members:` of all classes in `optimizer_wrappers.rst` (I checked it works, that is we still have the handwritten attributes even after removing members)

amosyou:
apologies for the delay! It looks like a lot of restructuring happened since my last changes. I can open a new pull request that does these changes on the new transform subpackage?

fabianp:
thanks @amosyou ! Yes, that's fine, do it however way its most convenient for you 😄

vroulet:
Thank you again for your efforts @amosyou. I believe this PR can be closed but feel free to reopen it if needed

Issue #757, opened 2024-02-02
Add a mathematical description of the algorithms


vroulet:
Like adam or nadam, it would be nice to have mathematical descriptions of as many algorithms as possible. 
To start with, having a clear description of what sgd with momentum and nesterov is would be very good.
Algorithms to do if possible below (sometimes the description may be too long). Refer to the reference each time (on arxiv you can even extract the source to potentially simply copy-paste the algorithm but make sure it matches the implementation).
- [x] SGD: https://github.com/google-deepmind/optax/pull/830
- [x] AdaBelief: https://github.com/google-deepmind/optax/pull/869
- [ ] Adagrad
- [ ] Adafactor
- [x] Adamax, Adamaxw: https://github.com/google-deepmind/optax/pull/918
- [x] AdamW: https://github.com/google-deepmind/optax/pull/894
- [ ] AMSGrad
- [ ] Fromage
- [ ] Lamb
- [ ] Lars
- [ ] Lion
- [x] Noisy SGD: https://github.com/google-deepmind/optax/pull/857
- [ ] Novograd
- [ ] OptimisticGD
- [ ] DifferentiallyPrivateSGD
- [ ] Radam
- [ ] RMSProp
- [ ] SM3
- [ ] Yogi

vroulet:
And by the way could be good to align equations on the left, with nice indent.
Adding 
`mathjax3_config = {'chtml': {'displayAlign': 'left',
                             'displayIndent': '2em'}}`
to conf.py should work (to be tested).

fabianp:
+100 to fix the equations. Could you please open a separate issue for that?

IronwallxR5:
Hi @vroulet  , I'm going to start working on this issue, could you assign it to me?

vroulet:
Thanks for taking this up @IronwallxR5.
Consider a lightweight and clear presentation each time, just as it has been done for Adam for example.
Harmony through the optimizers will be key for having a clear documentation.
Thanks again

julurisaichandu:
Hi @vroulet , I would like to add descriptions for a few of the methods. Could you also assign to me?
I think the adadelta and rprop methods are not mentioned in the above list, and they don't have the mathematical descriptions. So, could you assign me this issue so I can add descriptions for both methods?

jhaayush2004:
@vroulet , Kindly check the PR  #1239   RMSProp  mathematical description added. Screenshots of the mathematical equations final view are also attached.

[...]

PR #857, opened 2024-03-08
Added mathematical description to Noisy SGD

hmludwig:
I added mathematical description to noisy SGD to improve clarity (see #757)

The results look like this:
Screenshot 2024-03-08 at 12 46 34

### DIFF
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -1043,8 +1043,8 @@ def noisy_sgd(
     learning_rate: A global scaling factor, either fixed or evolving along
       iterations with a scheduler, see :func:`optax.scale_by_learning_rate`.
     eta: Initial variance for the Gaussian noise added to gradients.
-    gamma: A parameter controlling the annealing of noise over time, the
-      variance decays according to `(1+t)^-\gamma`.
+    gamma: A parameter controlling the annealing of noise over time ``t``, the
+      variance decays according to ``(1+t)**(-gamma)``.
     seed: Seed for the pseudo-random generation process.
 
   Returns:
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -992,9 +992,9 @@ def noisy_sgd(
     gamma: float = 0.55,
     seed: int = 0
 ) -> base.GradientTransformation:
-  r"""Noisy SGD is a variant of :func:`optax.sgd` that incorporates Gaussian noise 
-  into the updates. It has been found that adding noise to the gradients can 
-  improve both the training error and the generalization error in very deep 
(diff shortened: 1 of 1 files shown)


fabianp (approved):
looks great to me, thanks!

optax/_src/alias.py
```diff
@@ -992,11 +992,27 @@ def noisy_sgd(
     gamma: float = 0.55,
     seed: int = 0
 ) -> base.GradientTransformation:
-  r"""A variant of SGD with added noise.
```
fabianp, line 995: First line needs to end with ".", so I would leave the first line as it is, and add your description below

fabianp: and the line after should be blank, so something like:

```
r"""A variant of SGD with added noise.

Noisy SGD is ...
"""

Comment (hmludwig) [reply optax/_src/alias.py:995]: Could you clarify if the requirement for the first line to end with "." is driven by stylistic guidelines or technical reasons? The list of optimizers at the beginning of the documention isn't effected and displays the first sentence of the docstring correctly.

Comment (fabianp) [reply optax/_src/alias.py:995]: its a requirement of the python style: https://peps.python.org/pep-0257/#multi-line-docstrings

Internally we get an error when importing the code if the docstring doesn't follow this convention

Comment (fabianp) [reply optax/_src/alias.py:995]: you're right though that this would be something worth adding to https://optax.readthedocs.io/en/latest/development.html

File: optax/_src/alias.py [hunk 2]
```diff
@@ -994,9 +994,27 @@ def noisy_sgd(
...
-  
+  Noisy SGD is a variant of :func:`optax.sgd` that incorporates Gaussian noise
+  into the updates. It has been found that adding noise to the gradients can
+  improve both the training error and the generalization error in very deep
+  networks.
+
+  The update :math:`u_t` is modified to include this noise as follows:
+
+  .. math::
+    u_t \leftarrow -\alpha_t g_t + N(0, \sigma_t^2),
```
vroulet, line 1005: Thanks for doing that!
Shouldn't it rather be the following?
`u_t \leftarrow -\alpha_t (g_t + N(0, \sigma_t^2))`

vroulet (commented):
Thanks for doing this! 
As you are on it, could you change the description of `gamma` lines 1046-1047 of this file by 
```
gamma: A parameter controlling the annealing of noise over time ``t``, the
      variance decays according to ``(1+t)**(-gamma)``.
```
Thank you again!

optax/_src/alias.py
hmludwig: Yes, you're right. We first add the noise to the gradient and then scale by the learning rate. I will review my more carefully in the future.

PR #862, opened 2024-03-12
Test on Python 3.12

fabianp:


fabianp:
seems that is pytype now that isn't compatible with python 3.12 ...

vroulet:
Done in #1266

PR #895, opened 2024-03-28
Simplification for transforms with empty state

fabianp:


vroulet (commented):
Great idea!
Some comments:
- one may remove all corresponding states from the docs
- I think you may remove ScaleState (and modify scale(...))

I don't know why the tests are not finishing though... 
I can see that test_centralize has a weird `update(inputs, {})` might need to be `update(inputs, base.EmptyState())`

Issue #896, opened 2024-04-01
softmax_cross_entropy handles -inf logits incorrectly


carlosgmartin:
[`softmax_cross_entropy`](https://optax.readthedocs.io/en/latest/api/losses.html#optax.softmax_cross_entropy_with_integer_labels) handles a `-inf` logit incorrectly when the corresponding label is 0. For example:

```python3
import optax
from jax import nn, numpy as jnp
from jax.scipy.special import xlogy

def softmax_cross_entropy_alt_0(logits, labels):
    return -(xlogy(labels, nn.softmax(logits, -1))).sum(-1)

def softmax_cross_entropy_alt_1(logits, labels):
    return -(labels * nn.log_softmax(logits, -1)).sum(-1, where=labels != 0)

def softmax_cross_entropy_alt_2(logits, labels):
    x = labels * nn.log_softmax(logits, -1)
    return -jnp.where(labels == 0, 0, x).sum(-1)

logits = jnp.array([-jnp.inf, 0])
labels = jnp.array([0, 1])
print(optax.softmax_cross_entropy(logits, labels))  # nan
print(optax.softmax_cross_entropy(logits.clip(-1e10), labels))  # -0.0
print(softmax_cross_entropy_alt_0(logits, labels))  # -0.0
print(softmax_cross_entropy_alt_1(logits, labels))  # -0.0
print(softmax_cross_entropy_alt_2(logits, labels))  # -0.0
```

In principle, it should be doing what `softmax_cross_entropy_alt_0` does, which uses [`xlogy`](https://jax.readthedocs.io/en/latest/_autosummary/jax.scipy.special.xlogy.html), which treats `x * log(y)` as `0` when `x == 0`.

Two possible solutions are given by `softmax_cross_entropy_alt_1` and `softmax_cross_entropy_alt_2` above.

I can submit a PR for either.

(For context, I encountered this issue when doing action masking of logits in an RL setting.)

vroulet:
Hello @carlosgmartin,
Great catch! If you are up for doing a PR that would be great.
Why wouldn't you use the function xlogy, that is softmax_cross_entropy_alt_0? This seems to me to be the safest option.

carlosgmartin:
[...]

PR #898, opened 2024-04-02
Fix softmax_cross_entropy to handle -inf logits correctly when corresponding label is 0.

carlosgmartin:
#896

### DIFF
--- a/optax/losses/_classification.py
+++ b/optax/losses/_classification.py
@@ -128,7 +128,8 @@ def softmax_cross_entropy(
     distributions, with shape `[...]`.
   """
   chex.assert_type([logits], float)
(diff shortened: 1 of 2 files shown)


fabianp:
Would perhaps using logsumexp (https://jax.readthedocs.io/en/latest/_autosummary/jax.scipy.special.logsumexp.html) be a solution here?

Also, please add tests :-)

carlosgmartin:
@fabianp Added tests.

`log_softmax` uses `logsumexp` internally: https://github.com/google-deepmind/optax/issues/896#issuecomment-2032688959.

vroulet (approved):
Thanks again!
I understood also my confusion, you are modifying cross_entropy and not cross_entropy_with_integer_labels so it makes a lot of sense. Sorry for the confusion and thanks again!

PR #945, opened 2024-04-23
Weight decay for COCOB

albcab:
Adding L2 weight decay for COCOB, see Orabona's implementation: https://github.com/bremen79/parameterfree/blob/main/parameterfree/cocob.py#L85


albcab:
@vroulet

vroulet (approved):
Perfect, thank you very much @albcab !!

PR #953, opened 2024-04-27
Fix RST formatting issues.

fabianp:
I also removed the * from the product of DoG,
as in latex it seemed to me more like a hadamart
product, and I don't think we use this notation
for other solvers.


vroulet (approved):
Thanks for doing this!

Issue #969, opened 2024-05-21
Memory leak in optax.radam


lukekulik:
Number of live buffers (and memory use) continues to grow when using `optax.radam` optimizer. The issue goes away when the optimizer is switched to `optax.adam`. 

Potentially related to https://github.com/google-deepmind/optax/issues/580

CUDA Version: 12.0
JAX version: 0.4.1
Optax version: 0.1.4

Repro:
```python
import jax
import jax.numpy as jnp
import optax

from flax import linen as nn
from collections import namedtuple

client = jax.lib.xla_bridge.get_backend()

class MLP(nn.Module):
    features : namedtuple('features', ['outputsize'])
    
    def setup(self):
        self.linear = nn.Dense(self.features[0])
        self.relu = nn.activation.relu
    
    def __call__(self, x):
        return self.relu(self.linear(x)).max(axis=1)

key = jax.random.PRNGKey(0)

model = MLP(features = [10])

test_input = jax.numpy.ones((10,10,10))
test_label = jax.numpy.ones((10))

model_params = model.init(key, test_input)

optimizer = optax.radam(learning_rate=1e-3)  # this opt will overflow
# optimizer = optax.adam(learning_rate=1e-3)  # this opt will not overflow

opt_state = optimizer.init(model_params)

def mse_for_model(model):
  @jax.jit
  def mse(params, x_batched, y_batched):
    def squared_error(x, y):
      pred = model.apply(params, x)
      return jnp.inner(y-pred, y-pred) / 2.
    return jnp.mean(jax.vmap(squared_error)(x_batched,y_batched), axis=0)
  return mse

loss_grad_fn = jax.value_and_grad(mse_for_model(model))

for i in range(100):
    loss_val, grads = jax.block_until_ready(
        loss_grad_fn(model_params, test_input, test_label)) 
        
    bs = [b.shape for b in client.live_buffers()]
    print('num_live_buffers=', len(bs))
    
    updates, opt_state = optimizer.update(grads, opt_state) 
    params = optax.apply_updates(model_params, updates)
```

vroulet:
Hello @lukekulik,
Thanks for catching this and giving a reproducible example.
I think there is a simple fix for this: replace the jax.lax.cond in [here](https://github.com/google-deepmind/optax/blob/dd1dac479e8bb49f58fec98c0c34316cba9d2c99/optax/_src/transform.py#L764) by a jnp.where.
[...]

PR #974, opened 2024-05-25
Fix memory leak in radam optimizer

lukekulik:
Reference: https://github.com/google-deepmind/optax/issues/969

Replace jax.lax.cond with jnp.where in the radam optimizer update function. This increases computational time, but fixes the memory leak. This solution was originally proposed by vroulet@.


vroulet (approved):
Looks great, thank you very much @lukekulik.
Just to make sure: did you try and did it solve the memory leak issue?

lukekulik:
Yes, it looks like it fixed it. The number of live buffers in the the repro example from https://github.com/google-deepmind/optax/issues/969 doesn't continue to grow anymore.

vroulet:
Perfect thank you again!

Issue #968, opened 2024-05-20
Intended usage of the Sophia optimiser


vvvm23:
Hi all,

I noticed in [this commit](https://github.com/google-deepmind/optax/commit/7d43c5c0cc1ab343229c3c394b3179a1404e97e8) that the Sophia optimiser (see [paper](https://arxiv.org/abs/2305.14342)) has been integrated into `optax`.

However, I noticed the estimate of the diagonal hessian is far simpler than either variant proposed in the paper. 
![image](https://github.com/google-deepmind/optax/assets/44398246/2a9fdd82-5d3d-4f10-a5c1-3e52a3cc0c47)

The first uses a Hessian-vector product and the second uses label resampling. However, in the merged version of the code, [we only seem to use the square of the current gradient](https://github.com/google-deepmind/optax/blob/7d43c5c0cc1ab343229c3c394b3179a1404e97e8/optax/contrib/_sophia.py#L94-L103). Furthermore, this comment:
```
update_hessian_every: How often to update the second order terms. As these
      are very cheap to compute, in fact (just squaring gradients), we can leave
      these at 1 by default. Must be >= 1.
```
seems to further contradict the paper, which says the update is quite expensive to compute and is limited to (usually) every ten steps. This is pretty confusing as the api of `optax` makes it seem that all optimisers can be used as drop-in replacements for one another, but here it seems additional work is required to get the behaviour we want.

How should I use the optax version of the optimiser in order to replicate the behaviour in the paper? Can you share example code for doing this?

vroulet:
Thanks for catching this. You are absolutely right.
The implementation of recent optimizers in the contrib folder are the responsibility of the authors adding them.
@rumschuettel can you fix your implementation of the sophia optimizer?

jbausch-gdm:
Good catch indeed! The current version implements the special case of the GNB estimator for a squared loss only (as mentioned in the paper on p. 9) for which we can use the squared gradients to estimate the diagonal of the Hessian. I'll take a look at the other estimators as well.

evanatyourservice:
[...]

PR #979, opened 2024-05-29
Add sophia-h optimizer

evanatyourservice:
PR to add sophia optimizer. It's mostly based on levanter's implementation with some changes/added features here and there.

One note is that I had to change the contrib common test file a couple times, once to pass the loss_fn out of the parabola and rosenbrock functions (could be useful later for other optimizers that need loss function), and a second time to bypass the check for update arguments to be values (the loss function is not). Please advise if these changes are not ok or the most correct.


evanatyourservice:
fixes #968

optax/contrib/_sophia_h.py
```diff
@@ -0,0 +1,307 @@
...
+          lambda x, y: x + y,
+          jax.tree.map(
+              lambda u: jnp.sum(jnp.abs(u) < clip_threshold), updates
+          ),
+      )
+      total_tree_size = sum(x.size for x in jax.tree.leaves(updates))
+      win_rate = sum_not_clipped / total_tree_size
+      jax.lax.cond(
+          jnp.logical_and(
+              print_win_rate_every_n_steps > 0,
```
vroulet, line 154: I would wrap the logic associated to ``print_win_rate_every_n_steps > 0`` in an if statement to keep clear what operations are necessary for the algorithm and what are operations are simply for debugging if ``print_win_rate_every_n_steps > 0``

optax/contrib/_sophia_h.py
```diff
@@ -0,0 +1,307 @@
...
+          seed=seed,
+          print_win_rate_every_n_steps=print_win_rate_every_n_steps,
+      ),
+      transform.add_decayed_weights(weight_decay, mask=mask),
+      transform.scale_by_learning_rate(learning_rate),
+  ]
+  return chain(*tx)
+
+
+def _tree_rademacher_like(key, tree):
```
vroulet, line 292: Could you use `optax.tree_utils.tree_random_like`?
If the API of `optax.tree_utils.tree_random_like` does not match your needs, just keep what you have but feedback on `optax.tree_utils.tree_random_like` would be useful for us.

optax/contrib/_sophia_h.py
```diff
@@ -0,0 +1,307 @@
...
+
+    key, nu = update_hessian(state.key, state.count, state.nu, params, obj_fn)
+
+    mu = otu.tree_cast(mu, mu_dtype)
+    state = SophiaHState(count=count_inc, mu=mu, nu=nu, key=key)
+    return updates, state
+
+  def update_hessian(key, count, nu, params, obj_fn):
+    def _do_update(key):
+      if pmap_axis_name is not None and jax.local_device_count() > 1:
```
vroulet, line 173: I am not sure to understand the logic here.
I suppose the idea is that, by using pmap, we can compute an average of $m$ estimates where $m$ is the number of local devices?
But (i) we need an example on how to use it, in particular, how to set pmap_axis_name, (ii) we need to test it.
Concerning the second point, pmap is free computationally but a priori not free in terms of memory. We need to be able to have $m$ random copies of the parameters which may blow up the memory. Moreover, with an implementation using hvps as done right now, the graph of computation of the function would be replicated $m$ times too, which, I would believe, is infeasible for large models.
Moreover, I don't know how this logic will interact with shard_map or automatic parallelism/sharding with jit.
So either we keep this option but document and test it, or we leave it out for now. As you pointed out, we may keep the code close to the levanter implementation which does not incorporate such a logic.

optax/contrib/_sophia_h.py
```diff
@@ -0,0 +1,307 @@
...
+  to `loss_fn(params) -> loss` like below:
+
+  `sophia_obj_fn = lambda params: loss_fn(params, batch)[0]`
+
+  Then it can be passed to sophia's update function:
+
+  `updates, state = sophia.update(updates, state, params, obj_fn=sophia_obj_fn)`
+
+  Notes:
+    - TODO filter non-differentiable inputs to jvp
```
vroulet, line 77: What do you mean by that?
Are you referring to the ability to take additional input arguments of the function as done in e.g. optax.scale_by_backtracking_linesearch?

optax/contrib/_sophia_h.py
```diff
@@ -0,0 +1,307 @@
...
+def scale_by_sophia_h(
+    b1: float = 0.965,
+    b2: float = 0.99,
+    eps: float = 1e-8,
+    gamma: float = 0.01,
+    clip_threshold: Optional[float] = 1.0,
+    update_interval: int = 10,
+    mu_dtype: Optional[Any] = None,
+    pmap_axis_name: Optional[str] = None,
+    seed: PRNGKey = PRNGKey(0),
```
vroulet, line 57: I would probably make seed be either an int or a PRNGkey with default being seed=0.
If it is an int then convert it to a key later.

optax/contrib/_sophia_h.py
```diff
@@ -0,0 +1,307 @@
...
+from optax._src import base
+from optax._src import transform
+from optax._src.numerics import safe_int32_increment
+from optax._src.utils import canonicalize_dtype
+from optax._src.combine import chain
+from optax import tree_utils as otu
+
+
+class SophiaHState(NamedTuple):
+  """State for Sophia-H and similar."""
```
vroulet, line 40: What does similar refer to?

optax/contrib/_sophia_h.py
```diff
@@ -0,0 +1,307 @@
...
+
+  def update_hessian(key, count, nu, params, obj_fn):
+    def _do_update(key):
+      if pmap_axis_name is not None and jax.local_device_count() > 1:
+        # get current replica's key
+        idx = jax.lax.axis_index(pmap_axis_name)
+        key = jax.lax.dynamic_index_in_dim(key, idx, keepdims=False)
+
+      key, subkey = jax.random.split(key)
+      hess = _stochastic_hessian_diagonal(subkey, obj_fn, params)
```
vroulet, line 179: Call it `hess_diag` rather than `hess`

vroulet (commented):
Thank you very much @evanatyourservice! And sorry for the delay.
I left you some comments we can discuss about.

evanatyourservice:
Hi Vincent, thank you for the notes! They all make perfect sense to me and I'll get to updating the code/answering them tomorrow

optax/contrib/_sophia_h.py
evanatyourservice: I did have it that way but changed it to this to pass the injected hyperparams test, but if it's ok it fails that test I'll switch it back to the if statement. I could also modify the inject hyperparams test (in contrib._common_test.py) to not check every argument, just let me know what would be best

optax/contrib/_sophia_h.py
evanatyourservice: Ah I didn't know this existed, definitely

optax/contrib/_sophia_h.py
evanatyourservice: Yes exactly, it performs a separate monte carlo sample on each device by using a device-specific rng_key instead of a replicated rng_key, essentially giving free estimates. Without pmap_axis_name, if using pmap, the exact same sample would be carried out on each device because the same rng_key would be replicated on every device. This could be viewed as wasteful because the only thing needed for each device to perform its own unique sample is a unique rng_key. The params or vectors do not have to be multiplied memory-wise on each device to do this, only the rng_keys.

For example, if we are using pmap with 4 devices and provide pmap_axis_name="batch", 4 unique rng_keys are made in init and all 4 are held on every device. When it comes time to update the hessian diag, each device slices its own of the 4 rng_keys using these lines:
```
        idx = jax.lax.axis_index(pmap_axis_name)
        key = jax.lax.dynamic_index_in_dim(key, idx, keepdims=False)
```
Then the hutchinson's sample is carried out with the device's single unique rng_key, the samples are averaged using pmean, and then the rng_keys are all_gathered back together. The costs are the memory of holding 4 rng_keys on each device instead of 1, the pmean for averaging the hessian diagonal estimates across devices, and the all_gather for the 4 rng_keys.

It shouldn't mess with jit at all because if one isn't using pmap the user can leave the argument as None and it will not be used and act as the normal optimizer. I think it's a very useful feature because, if using pmap, performing the exact same monte carlo sample on each device could be seen as wasteful when the only thing needed for them to be unique samples are unique rng_keys for the random vector creation. I originally got this idea from distributed shampoo where in pmap mode computations are split across devices using jax.lax.axis_index, slicing, and all_gathering. I'll document the feature better so users can understand what's going on and make using it easier, and I can add a test for it using xla_force_host_platform_device_count.

optax/contrib/_sophia_h.py
evanatyourservice: This is referring to how levanter's version uses equinox.filter_grad but I think we can forget about this for now unless optax wants to host a version of filter_grad similar to equinox's https://github.com/stanford-crfm/levanter/blob/fe414e2fc9116eb7a8bec4f639d86b1bf10be988/src/levanter/optim.py#L67

I think for now I can just delete this line

optax/contrib/_sophia_h.py
evanatyourservice: Ok can do

optax/contrib/_sophia_h.py
evanatyourservice: Just a line from levanter, I can change it

optax/contrib/_sophia_h.py
```diff
@@ -0,0 +1,307 @@
...
+      list(keys),
+      leaves,
+  )
+  r = jax.tree.unflatten(structure, r)
+  return r
+
+
+def _stochastic_hessian_diagonal(key, obj_fn, model):
+  random_signs = _tree_rademacher_like(key, model)
+  product = jax.jvp(jax.grad(obj_fn), (model,), (random_signs,))[1]
```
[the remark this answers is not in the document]
evanatyourservice: When I try this with forward in a forward it only returns a scalar, or am I missing something?

```
def _hqp(key, obj_fn, model):
    random_signs = otu.tree_random_like(
        key, model, partial(jax.random.rademacher, dtype=jnp.float32)
    )

    def jvp_fun(params):
        return jax.jvp(obj_fn, (params,), (random_signs,))[1]

    return jax.jvp(jvp_fun, (model,), (random_signs,))[1]
```

optax/contrib/_sophia_h.py
evanatyourservice: Oh this is another argument I changed to pass injecting hyperparams test, i think it has to be only one or the other, should it be a key or int?

optax/contrib/_sophia_h.py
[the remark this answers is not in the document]
vroulet: My bad, I was thinking of the trace estimator. So let's keep the hvp way. But then the memory footprint is not negligible.

optax/contrib/_sophia_h.py
vroulet: Ok, thanks for the explanations!
We would mostly need one example to show how to use it.
I would still believe that at one point the m samples need to be created in the memory before being fed to the hvp so a priori we would incur a memory cost of m d (for m the number of devices and d the size of the parameters). I don't see how this can be avoided. In terms of computations, it should indeed be no more costly than not using pmap, so as long as the memory allows it it's a great feature. I would just mention the memory cost (if you agree with me).

optax/contrib/_sophia_h.py
vroulet: Yes, maybe simpler for now to delete this. Thanks for the answer!

optax/contrib/_sophia_h.py
vroulet: Let's make it an optional PNRGKey then. So if the user does not want to know all the details of the algorithm, he/she does not need to specify it (we may just instantiate it at the start if it is a None).

optax/contrib/_sophia_h.py
vroulet: Ah I see. Let's keep it the way you had it then.

optax/contrib/_sophia_h.py
evanatyourservice: ok actually I think for the sake of simplicity I will just remove this feature for now haha, but what would be the most concise way to average the hessian diagonals across devices/sharded data axis when using either pmap or jit with sharding, akin to averaging gradients across devices? Would an argument asking for a named axis suffice?

vroulet: I think the hutchinson estimators (trace or diag) should be part of another functionality in optax.
As an example [matfree](https://github.com/pnkraemer/matfree) has a rather neat api that we could either directly use or take inspiration of.
Concretely, I'm wondering if it would be possible for sophia_h to take as an argument a function `hutch_hess_diag_fn` given from something like `hutch_hess_diag_fn = hutch_hess_diag(fun, **extra_args)` with a signature `hess_diag = hutch_hess_diag_fn(params)` (semantic akin to a `jax.grad`). I'd suppose that such a function needs also an inner state (to store the key) so maybe more something like `hess_diag, hutch_state = hutch_hess_diag_fn(params, hutch_state)`. 
That way we can postpone for now the implementation of a clever hutchinson estimator and such a clever estimator could be used for other purposes. The `extra_args` above could incoporate pmap axis etc... Also that way it would be easier to make unit tests on the pmap/shard_map behaviors.

evanatyourservice: Excellent yeah that sounds like a great idea for the long run, and matfree's api does look really nice. I'll add the argument for a custom hvp fn, and have a default one in the same file using the `hess_diag, hutch_state = hutch_hess_diag_fn(params, hutch_state)` format, just so people can get started using the optimizer and have a good example of how to make their own fn if they want. Then later optax could add hvp fns into their main API if they wanted. I'll also rename the optimizer to sophia instead of sophia-h, because then a user can write their own fn for sophia-g, or any other kind of preconditioner they'd like, it won't be hardcoded to be sophia-h.

vroulet: Sounds perfect! Thanks for all the work @evanatyourservice !

fabianp:
@evanatyourservice please ping us whenever you're ready for another round of reviews :-)

evanatyourservice:
@fabianp Will do! Sorry been moving but will try to get this going asap

fabianp:
there's no rush, just wanted to make sure you were not waiting on us :-)

evanatyourservice:
@vroulet @fabianp Got some updates pushed, let me know if anything needs to be changed! Thanks

optax/contrib/_sophia_h.py
evanatyourservice: I think I got a good enough setup with this at least for a contrib but let me know your thoughts! Not sure if I've documented it well enough too

vroulet: That's great thank you!

vroulet:
Hello @evanatyourservice,
Sorry for the very long delay on our end. I think your code looks great!
If it's ok with you, can you merge the code with head once #1060 is merged (#1060 adds more tests for the contrib optimizers to ensure compatibilities). Then I should approve and finish on our side if there are still minor details to fine-tune.
Thank you again!

evanatyourservice:
sounds good!

optax/contrib/_sophia.py
```diff
@@ -0,0 +1,339 @@
...
+    return HutchinsonState(key=key)
+
+  def update_fn(updates, state, params=None, obj_fn=None):
+    if params is None:
+      raise ValueError("params must be provided to hutchinson update function.")
+    if obj_fn is None:
+      raise ValueError("obj_fn must be provided to hutchinson update function.")
+    del updates
+    key, subkey = jax.random.split(state.key)
+    random_signs = otu.tree_random_like(
```
fabianp, line 71: As far as I can tell from the paper (https://arxiv.org/pdf/2305.14342, section 2.3), it computes the Hutchinson estimator using a Normal distribution, while here we use a Rademacher distribution. The Rademacher distribution should be lower variance, but perhaps it's worth to add a comment in that this deviates from what is specified in the paper?

vroulet:
Hello @evanatyourservice,
#1060 got merged. You may merge the tests, address Fabian's comment, and I can approve (and maybe fix minor issues on our end if there are).
Thank you again!

evanatyourservice:
Ok sounds good! Sorry I should get to this tomorrow

optax/contrib/_sophia.py
evanatyourservice: Hi Fabian! Yes it seemed some math worked out to lower variance with rademacher when dealing with diag hessian like in sophia but I'll actually just revert to normal for simplicity and staying true to the paper for now.

evanatyourservice:
@fabianp @vroulet wow ok 1 day turned into 2 weeks, let me revert rademacher to normal and make sure the tests are all good and I'll push update :)

fabianp:
gentle ping @evanatyourservice :-)

no rush but it would be a really nice feature to have landed in optax ❤️

vroulet:
Thank you again @evanatyourservice ! We added tests and made some final fixes necessary for the tests to pass.

evanatyourservice:
Awesome! Thank you for finishing it up! I got kinda stumped at some of the tests and was debating another strategy but then it fell out of my mind for other tasks. Glad we were able to get it merged!

PR #982, opened 2024-06-08
Fix LaTeX rendering issue in docstring of lbfgs in aliases for optimi…

mcharrak:
Fix LaTeX rendering issue in docstring of lbfgs in aliases for optimizers file alias.py

### DIFF
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -65,7 +65,7 @@ def adabelief(
 
     \begin{align*}
       m_t &\leftarrow \beta_1 \cdot m_{t-1} + (1-\beta_1) \cdot g_t \\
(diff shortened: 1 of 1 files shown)


fabianp (approved):
thanks!

vroulet:
Thank you! The error in the lbfgs docstring is not solved by this pr.
The real error stems from using mbox. One can use text instead. This error appears both in lbfgs and scale_by_lbfgs.
If that's ok with you, I'd like to merge a PR that sweeps over current doc errors.
I've not made a pass on trailing spaces as you did, so if you want you can keep this PR that fixes these errors (once the aforementioned PR is merged).

mcharrak:
Of course, no problem!

PR #987, opened 2024-06-14
CI: add test against oldest supported JAX version

jakevdp:
Also bump minimum supported version to v0.4.27, because the initial test run revealed that optax is no longer compatible with jax v0.4.26

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,13 +10,18 @@ on:
 
 jobs:
   build-and-test:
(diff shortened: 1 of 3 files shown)


vroulet (approved):
Thank you for doing that!
So would you like us to ensure compatibility with older JAX versions?

jakevdp:
I don't think there's any need for compatibility with particular JAX versions, but I think it's important that the conditions in the requirements file reflect actual compatibility. Hopefully this new test will help maintain that. Thanks!

PR #1012, opened 2024-07-15
feat: add mathematical notation docs of SM3 optimizer

Abhinavcode13:
Addresses: #757 
cc: @fabianp

### DIFF
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -1501,20 +1501,21 @@ def sm3(
   SM3-I Algorithm
 
   .. math::
-    \begin{align*}
-    & \text{parameters: learning rate } \eta \\
-    & \text{initialize } w_1 = 0; \forall r \in [k]: \mu_0(r) = 0 \\
-    & \text{for } t = 1, \ldots, T \text{ do} \\
-    & \quad \text{receive gradient } g_t = \nabla \ell_t(w_t) \\
-    & \quad \text{for } r = 1, \ldots, k \text{ do} \\
-    & \quad \quad \mu_t(r) \leftarrow \mu_{t-1}(r) \\
-    & \quad \quad \quad \quad + \max_{j \in S_r} g_t^2(j) \\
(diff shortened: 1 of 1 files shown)


optax/_src/alias.py
```diff
@@ -1488,7 +1488,63 @@ def sm3(
...
+  cumulative squared gradients :math:`\mu_t(r)` for each weight component 
+  :math:`r`, and updates :math:`w_t` using received gradients :math:`g_t`, 
+  adjusting each component based on the minimum accumulated gradient 
+  :math:`\nu_t(i)`. It employs a convention where division by zero yields zero, 
+  optimizing convergence by adapting to varying gradients across different 
+  parts of the weight vector.
+
+  SM3-I Algorithm
+
+  .. math::
```
vroulet, line 1502: 1. Use subscripts instead of parenthesizing unless it's a function.
2. Make the description follow the function implemented: here we implement SM3 as a gradient transformation, so the doc should detail the gradient transformation, not T iterations of the algorithm. You may detail then the `init_fn` and the `update_fn` (so that the initial values are described clearly).

optax/_src/alias.py
```diff
@@ -1488,7 +1488,63 @@ def sm3(
...
+    & \quad \quad w_{t+1}(i) \leftarrow w_t(i) \\
+    & \quad \quad \quad \quad - \eta \frac{g_t(i)}{\sqrt{\nu_t(i)}} \\
+    & \quad \quad \text{with the convention that } 0/0 = 0
+    \end{align*}
+
+
+
+  SM3-II Algorithm
+
+  The algorithm initializes with parameters such as the learning rate
```
vroulet, line 1522: Explains the difference with the SM3-I algorithm rather than rewriting a similar description

vroulet (commented):
Few comments. Thanks for doing that!

Abhinavcode13:
Rendered HTML:

image
image

Abhinavcode13:
ready for review @vroulet

fabianp (approved):

PR #1013, opened 2024-07-16
Fix docs for dowg

jungtaekkim:
Similar to the DoG optimizer, the example for dowg also needs to be fixed.


vroulet (approved):
Thanks for doing that!

PR #1021, opened 2024-07-24
keep a local .pylintrc file

fabianp:
This makes it easier to use tools like vscode that rely on the local .pylintrc file. Otherwise one would see a lot of warnings the first time opening the file, as we use non-standard (google) python idioms.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,6 @@ docs/modules/generated
 .mypy_cache/
 .pytype/
 .ipynb_checkpoints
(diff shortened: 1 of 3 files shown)


vroulet (approved):
Great idea, thanks!
Btw don't know why the benchopt paper cannot be loaded with its absolute file, but you may replace it with the pdf link (https://arxiv.org/pdf/2206.13424 instead of https://arxiv.org/abs/2206.13424 )

PR #1024, opened 2024-07-24
keep a local .pylintrc file

fabianp:
This makes it easier to use tools like vscode that rely on the local .pylintrc file. Otherwise one would see a lot of warnings the first time opening the file, as we use non-standard (google) python idioms.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,6 @@ docs/modules/generated
 .mypy_cache/
 .pytype/
 .ipynb_checkpoints
-.pylintrc
 
 # Editors
 .idea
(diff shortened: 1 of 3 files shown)


vroulet (approved):

carlosgmartin:
@fabianp @vroulet If `.pylintrc` is meant to stay, why does `test.sh` delete it? This "tug-of-war" seems to be causing issues when running tests: https://github.com/google-deepmind/optax/issues/1067.

IMO: Running tests should be free of side effects; we don't want running tests to yield different results depending on whether they were run previously.

fabianp:
Absolutely, that's a bug. We should not be deleting the .pylintrc file. Could you send a PR fixing that? Thanks 🙏!

carlosgmartin:
@fabianp Should test.sh also be amended to not modify .pylintrc, as it currently does on [line 64](https://github.com/google-deepmind/optax/blob/fad9a2028b06ad2d9d8805cc3de5c52b00963700/test.sh#L64)?

PR #1030, opened 2024-07-30
Fix typo

gil2rok:
Change scheduels to schedules. Also remove unnecessary comma.

### DIFF
--- a/optax/_src/base.py
+++ b/optax/_src/base.py
@@ -163,7 +163,7 @@ class GradientTransformation(NamedTuple):
 
   Since gradient transformations do not contain any internal state, all stateful
   optimizer properties (such as the current step count when using optimizer
-  schedules or  momemtum values) are passed through optax gradient
+  schedules or  momentum values) are passed through optax gradient
   transformations by using the optimizer _state_ pytree. Each time a gradient
   transformation is applied, a new state is computed and returned, ready to be
   passed to the next call to the gradient transformation.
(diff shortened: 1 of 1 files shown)


vroulet (approved):
Thanks!

gil2rok:
Would love some advice on small doc improvements for the `GradientTransformation` class in `base.py`.

1. The doc refers to Optax's meta-learning example. I added a link to this in the doc: [https://optax.readthedocs.io/en/latest/_collections/examples/meta_learning.html](https://optax.readthedocs.io/en/latest/_collections/examples/meta_learning.html). Would a relative path to the meta-learning example be preferred? If so, how is this formatted?
2. If the link format is preferred, it goes past the end of line that the rest of the doc respects. How are long links handled?
3. There are a couple weirdly formatted underscores. Not sure what they are trying to accomplish, but wanted to highlight this:

Screenshot 2024-07-30 at 3 54 25 PM

vroulet:
Hello @gil2rok,

1. The docs use reStructuredText whose semantics are not the same as usual markdown. See e.g. https://docs.anaconda.com/restructuredtext for some tips. In this case, I would simply use
```
`meta-learning example <https://optax.readthedocs.io/en/latest/_collections/examples/meta_learning.html>`_
 ```
There may be a way to link internal refs but that's not very important.
2. You may add
```
# pylint: disable=line-too-long
(The doc string...)
# pyline: enable=line-too-long
```
to avoid linting issues because of too long lines. (see e.g. [rmsprop in the alias file](https://github.com/google-deepmind/optax/blob/d7f3e4bee8637645926709c50d29a5319ea6d8ea/optax/_src/alias.py#L1346C3-L1346C34))
3. I suppose the intent was to emphasize these elements if the docs were in markdown. But as said above the docs use reStructuredText (rst) so one should use the syntax of rst to emphasize things here. (Or we may simply remove these underscores, it's not super important as long as it is displayed in a readible fashion which is not the case right now).

Thanks again!

gil2rok:
Silly mistake. I didn't realize docs are in rst and not in Markdown!

All fixed now.

vroulet:
Thanks again, can you send a screenshot of the resulting doc?
Also as you said you may replace the underscores _ by the corresponding emphasizing syntax in rst (wrap text between asterisks *).

gil2rok:
All done!

Screenshot 2024-08-01 at 10 43 05 PM

vroulet (approved):

Issue #905, opened 2024-04-04
Documentation for cosine decay schedule


gjhuizing:
Hello,

The formula in the documentation for the `cosine_decay_schedule` (https://optax.readthedocs.io/en/latest/api/optimizer_schedules.html#optax.cosine_decay_schedule) would suggest that the learning rate increases again after T steps.

A quick look at the code confirms this is not the case, but it may be good to write it explicitly, as in `linear_schedule`.

Happy to make a short PR! I also could propose a short formula/pseudocode for functions like `piecewise_constant_schedule` that do not have one.

Best

GJ

vroulet:
Hello @gjhuizing,

Thanks for catching this! If you are willing to do such a PR that would be great!

gjhuizing:
Great!

fabianp:
Solved in https://github.com/google-deepmind/optax/pull/1032

PR #1032, opened 2024-08-12
Updated docs cosine_decay_schedule

bhargavyagnik:
Updated the documentation of cosine_decay_schedule from the previous 
https://optax.readthedocs.io/en/latest/api/optimizer_schedules.html#optax.cosine_decay_schedule.

I mentioned 2 different cases of decay based on no. of steps as per the source code.
✅ Solves #905

Previously the formula in the documentation was confusing in the scenario when t > decay_steps.


vroulet (approved):
Thanks for doing that!

PR #1037, opened 2024-08-30
add doctest for polynomial_schedule

fabianp:


vroulet (approved):
Thanks!
You could have an example with `transition_begin=5` and print more values. The doc is not great at explaining what happens before and after so an example could help clarifying.
```
>>> schedule_fn = optax.polynomial_schedule(
...    init_value=1.0,
...    end_value=0.01,
...    transition_steps=100,
...    transition_begin=5,
...    power=2,
... )
>>> counts = [0, 5, 6, 104, 105, 110]
>>> print(*[f'count:{i} value:{schedule_fn(i):.4f}' for i in counts], sep='\n')
count:0 value:1.0000
count:5 value:1.0000
count:6 value:0.9803
count:104 value:0.0101
count:105 value:0.0100
count:110 value:0.0100
```

Issue #1038, opened 2024-08-30
Error when using `optax.Multisteps` with `optax.contrib.schedule_free`


ozanarmagan:
Hello,

I am getting the following exception when I try to wrap a schedule free optimizer with multisteps. Can you help me?

```python
        learnin_rate_fn = optax.schedules.warmup_constant_schedule(peak_value=config.learning_rate, warmup_steps=config.num_warmup_steps, init_value=0.0)
        optimizer = optax.adamw(learning_rate=learnin_rate_fn, b1=0.)
        optimizer = optax.contrib.schedule_free(optimizer, learning_rate=learnin_rate_fn, b1=config.b1, state_dtype=jnp.bfloat16)
        optimizer = optax.MultiSteps(optimizer, every_k_schedule=config.accum_steps)
```

Exception message:
`
jax.errors.SimplifiedTraceback: For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/ozan/cloud/train/trainer.py", line 116, in train_step
    new_state = state.apply_gradients(grads=grads, train_rngs=new_train_rngs)
  File "/home/ozan/.local/lib/python3.10/site-packages/flax/training/train_state.py", line 101, in apply_gradients
    updates, new_opt_state = self.tx.update(
  File "/home/ozan/.local/lib/python3.10/site-packages/optax/transforms/_accumulation.py", line 380, in update
    new_updates, new_state = lax.cond(
TypeError: true_fun and false_fun output must have identical types, got
[...]

Issue #377, opened 2022-07-21
Adafactor + MultiStep with bfloat16 model doesn't work


Sea-Snell:
If I use Adafactor with MultiStep on a bfloat16 model I get this strange error (note the error is extremely long, so I truncated it to fit in the issue; the model is T5-small):

```
Traceback (most recent call last):
  File "/home/charliesnell/jax_v_pytorch/large_lm_finetune/flax/main.py", line 135, in <module>
    train.unroll(metaconfig)
  File "/home/charliesnell/miniconda3/envs/jax_v_torch/lib/python3.9/site-packages/micro_config.py", line 39, in new_unroll
    result = unroll(self, metaconfig)
  File "/home/charliesnell/jax_v_pytorch/large_lm_finetune/flax/train_loop.py", line 372, in unroll
    logs, params, opt_state = p_step_fn(params, opt_state, new_rng, items)
  File "/home/charliesnell/miniconda3/envs/jax_v_torch/lib/python3.9/site-packages/jax/experimental/pjit.py", line 352, in wrapped
    args_flat, _, params, _, out_tree, _ = infer_params(*args, **kwargs)
  File "/home/charliesnell/miniconda3/envs/jax_v_torch/lib/python3.9/site-packages/jax/experimental/pjit.py", line 330, in infer_params
    jaxpr, canonicalized_out_axis_resources_flat = _pjit_jaxpr(
  File "/home/charliesnell/miniconda3/envs/jax_v_torch/lib/python3.9/site-packages/jax/experimental/pjit.py", line 490, in _pjit_jaxpr
    jaxpr, global_out_avals, consts = pe.trace_to_jaxpr_dynamic(fun, global_in_avals)
  File "/home/charliesnell/jax_v_pytorch/large_lm_finetune/flax/train_loop.py", line 337, in t5_step_fn
    updates, opt_state = optim.update(grads, opt_state, params)
  File "/home/charliesnell/miniconda3/envs/jax_v_torch/lib/python3.9/site-packages/optax/_src/wrappers.py", line 413, in update
    new_updates, new_state = jax.lax.cond(
  File "/home/charliesnell/miniconda3/envs/jax_v_torch/lib/python3.9/site-packages/jax/_src/traceback_util.py", line 162, in reraise_with_filtered_traceback
    return fun(*args, **kwargs)
[...]

Issue #337, opened 2022-04-26
`noisy_sgd` adds noise after scale by learning rate.


HanGuo97:
Thanks for the effort in this awesome library!

According to [these](https://github.com/deepmind/optax/blob/master/optax/_src/alias.py#L408#L435), it seems like the noise is added after scaling by the learning rate. That is, the noise is independent of the learning rate. I'm curious if this is an intentional design?

n2cholas:
Good catch, this does not seem correct. My understanding is the noise is supposed to be added to the raw gradient, independent of the actual optimization method. That is, learning rate and any other scaling should take place after the noise is added to the gradient. Below is a snippet from the paper: https://arxiv.org/pdf/1511.06807.pdf

![image](https://user-images.githubusercontent.com/12474257/175794719-716714a2-1d7c-4b07-83bc-3d62defbc600.png)

The order has been this way since the library was originally published on GitHub (https://github.com/deepmind/optax/blame/d0ba208a300d68a3f4e380ed2a0b70848380988b/optax/_src/alias.py#L37-L39), so perhaps @mtthss has some insight? If there's no reason to keep this order, I think it makes sense to correct this.

mtthss:
I spoke with one of the authors and I can confirm that the noise should be added before multiplying by the learning rate.
@HanGuo97 do you want to put together a PR fixing this?

HanGuo97:
Unfortunately, I'm swamped with other responsibilities at the moment. I can come back to this later. In the meanwhile, contributions from others are more than welcome!

atgctg:
Shouldn't the fix just be this?

![Screen Shot 2022-06-29 at 20 42 52](https://user-images.githubusercontent.com/105969161/176512133-9f444cfe-0526-4950-8313-08bd90c04c23.png)

Not really sure how to validate this, so just ran the `test_optimization` test in `alias_test.py`. Here's the before & after:
<details>
<summary>
_setup_parabola params at every 1000th step
</summary>

#### Before

```
[-1. 10.  1.]
[0.7257005 0.4779135 1.02636  ]
[ 0.96638435 -0.8105099   0.98194563]
[ 1.0049064  -0.97684264  1.0158088 ]
[ 0.98075825 -0.99098873  1.0355433 ]
[ 0.99105674 -0.9999055   0.99992895]
[ 0.99034065 -1.0062125   0.99989396]
[ 1.0017453 -1.0287409  1.0216172]
[...]

PR #1039, opened 2024-09-01
fix: MultiSteps with mixed precision params and updates.

hlzl:
Aims at solving #377 and #1038 

## Problem
`MultiSteps` throws the following error when using different precision for parameters and the corresponding gradient values (e.g. `bfloat16` parameters but casting the parameters' gradient to `float32` for scaling such as in `scale_by_factored_rms` and `flax.training.dynamic_scale`):
```bash
File "/home/user/miniforge3/envs/name/lib/python3.12/site-packages/optax/transforms/_accumulation.py", line 380, in update
    new_updates, new_state = lax.cond(
                             ^^^^^^^^^
TypeError: true_fun and false_fun output must have identical types, got
({'decoder_image_output': {'bias': 'DIFFERENT ShapedArray(bfloat16[768]) vs. ShapedArray(float32[768])', 'kernel': ...
```
This is due to `jax.lax.cond` expecting the outputs of both conditions to have the same `dtype`.

## Code to Reproduce
Taken from #377 :
```python
import os; os.environ['JAX_PLATFORMS'] = 'cpu'
import jax
import jax.numpy as jnp
import optax

@jax.jit
@jax.value_and_grad
def f(params, x, labels):
    logits = params @ x
    loss = optax.softmax_cross_entropy_with_integer_labels(logits, labels)
    return loss.mean()

params = jnp.zeros((5, 18), dtype=jnp.bfloat16)
x = jnp.zeros((18, 4), dtype=jnp.bfloat16)
labels = jnp.zeros((5,), dtype=jnp.uint16)
value, grad = f(params, x, labels)

lr = 0.00005
n_accumulation_steps = 4

optimizer = optax.adafactor(learning_rate=lr)
optimizer = optax.MultiSteps(optimizer, n_accumulation_steps)
updates, opt_state = optimizer.update(grad, opt_state, params)
print(updates)
```

## Proposed Fix
Casting new updates to the `dtype` of the original optimizer state. Allows to solve the problem independent of the update `dtype`.

### DIFF
--- a/optax/transforms/_accumulation.py
+++ b/optax/transforms/_accumulation.py
@@ -334,7 +334,9 @@ class MultiSteps:
     # Note: we do not enclose variables to allow JAX to re-use memory buffers.
     def _do_update(updates, state, params):
       acc_grads = jtu.tree_map(
-          lambda upd, acc: self._acc_update(upd, acc, n_acc=state.mini_step),
+          lambda upd, acc: self._acc_update(
+              upd, acc, n_acc=state.mini_step
+          ).astype(acc.dtype),
           updates,
           state.acc_grads,
       )
@@ -350,7 +352,7 @@ class MultiSteps:
           * numerics.safe_int32_increment(state.gradient_step)
           + (1 - emit) * state.gradient_step,
           inner_opt_state=jtu.tree_map(
-              lambda st, nst: jnp.where(emit, nst, st),
+              lambda st, nst: jnp.where(emit, nst, st).astype(st.dtype),
               state.inner_opt_state,
               new_inner_state,
           ),
@@ -361,7 +363,9 @@ class MultiSteps:
       )
 
       final_updates = jtu.tree_map(
-          lambda ga: emit * ga, final_updates
+          lambda ga, st: jnp.array(emit * ga, dtype=st.dtype),
+          final_updates,
+          state.acc_grads
       )
       return final_updates, new_state
 


optax/transforms/_accumulation.py
```diff
@@ -334,7 +334,9 @@ def update(self,
     # Note: we do not enclose variables to allow JAX to re-use memory buffers.
     def _do_update(updates, state, params):
       acc_grads = jtu.tree_map(
-          lambda upd, acc: self._acc_update(upd, acc, n_acc=state.mini_step),
+          lambda upd, acc: self._acc_update(
```
hlzl, line 337: Now that I look at it again, a slightly better version would be the following as we can expect the leaves to be `jnp.array`s:
```python
acc_grads = jtu.tree_map(
          lambda upd, acc: self._acc_update(upd.astype(acc.dtype), acc, n_acc=state.mini_step)
          updates,
          state.acc_grads,
      )
```
and the following for line 365 ff.:
```python
final_updates = jtu.tree_map(
           lambda ga, st: emit * ga.astype(st.dtype),
           final_updates,
           state.acc_grads
       )
```
Let me know what you think of the proposed fix in general and I'll updates these.

vroulet:
Thank you very much for tacking a look at this!
1. Could you add a test?
2. Since the goal is to ensure the same types whether the update is actually used or not, one may use `jax.eval_shape`.

Happy to take a closer look once you have a test ready to fiddle with different solutions.

vroulet:
Hello @hlzl,

Thank you again for looking into this! Looking back at your example the only culprit is in the definition of the `zero_updates` in the `_skip_update` function.
Thank you again for the MWE, below is a proposed fix.
Would you be willing to do this and add a test?

Thank you again!
```
import os; os.environ['JAX_PLATFORMS'] = 'cpu'
import jax
import jax.numpy as jnp
import jax.tree_util as jtu
import optax
import optax.tree_utils as otu

@jax.jit
@jax.value_and_grad
def f(params, x, labels):
    logits = params @ x
    loss = optax.softmax_cross_entropy_with_integer_labels(logits, labels)
    return loss.mean()

params = jnp.zeros((5, 18), dtype=jnp.bfloat16)
x = jnp.zeros((18, 4), dtype=jnp.bfloat16)
labels = jnp.zeros((5,), dtype=jnp.uint16)
value, grad = f(params, x, labels)

lr = 0.00005
n_accumulation_steps = 4

optimizer = optax.adafactor(learning_rate=lr)
optimizer = optax.MultiSteps(optimizer, n_accumulation_steps)
state = optimizer.init(params)
# updates, opt_state = optimizer.update(grad, opt_state, params)

extra_args = {}
zero_updates = otu.tree_zeros_like(state.acc_grads)
print(f'zero updates dtype before fix: {jtu.tree_map(jnp.dtype, zero_updates)}' )
update_shape, _ = jax.eval_shape(
    optimizer.inner_opt.update,
    updates=otu.tree_zeros_like(params),
    state=state.inner_opt_state,
    params=otu.tree_zeros_like(params),
    **extra_args
)
zero_updates = jtu.tree_map(lambda x, s: x.astype(s.dtype), zero_updates, update_shape)
print(f'zero updates dtype after fix: {jtu.tree_map(jnp.dtype, zero_updates)}' )
```

hlzl:
Hi @vroulet,
Thank you for looking into the fix and sorry for my late responds!
I'm happy to have a look and add the test. Unfortunately, I'm a bit busy the next 2-3 weeks. I'll try to find some time for this in-between but will definitely have a look after my deadline. Hope this works for you!

vroulet:
No worries, it can wait, thank you again for looking into that!

mar-muel:
Hello - I'm also being blocked by this - Great to see a fix is in the workings 🙏 

In the meantime, is there a workaround or is `optax.MultiSteps` just generally incompatible with non-fp32 params for now?

vroulet:
@mar-muel, as a quick patch you may try adding just before calling the update function.
```
update_dtype = ... # add the expected dtype of the optimizer you are working with
state = optax.tree_utils.tree_set('acc_grads', optax.tree_utils.tree_cast(state.acc_grads, update_dtype)
```
PS: I don't think that optax.MultiSteps is generally incompatible with non-fp32 params. @hlzl singled out well the issue I think (mismatch between type of returned updates of inner optimizer and type of generic gradients (which is normally the same type as the parameters). @mar-muel are you referring to another issue?

mar-muel:
Unfortunately, there seems to be another upcasting happening internally in optax. So even if one manually casts the `inner_opt_state` to bfloat16 before applying the updates the same error will be raised (`DIFFERENT ShapedArray(bfloat16[3072]) vs. ShapedArray(float32[3072])`).

For reference here is what I've tried to do:
```python
inner_opt_state = jax.tree_util.tree_map(lambda s: s.astype(jnp.bfloat16) if s.dtype == jnp.float32 else s, opt_state.inner_opt_state)
opt_state = opt_state._replace(inner_opt_state=inner_opt_state)       

# optimizer.update(...)                                                                                                                               
```

hlzl:
@mar-muel I believe casting the update to be of the opt_state type is more robust. You could try copying my proposed changes from the PR and update your local optax version (it's only a few lines of code). A bit of a hack for right now but would also allow to test my changes ;)

mar-muel:
Hey @hlzl - Yes, I can confirm that the fix works (I've applied your changes to my current `optax==0.2.2`). However, I run into OOM sadly. Is the opt state still allocated in fp32 at one point, yielding a memory spike? It also seems like optax.MultiSteps is not very memory efficient in general, given that it keeps a full duplicate of the opt state :/

hlzl:
@mar-muel Good to know, thanks! 
As far as I can tell, all states within the optimizer (which is `inner_opt_state and `acc_grads`) are of the same types as the params you pass to the optimizer when calling `init`, no other upcasting.

vroulet:
The culprit is not MultiSteps. The culprits are all optimizers that do not preserve the dtype of the parameters. I'm working on a PR to correct that across the code base. The PR should be ready today and pushed next week. I'll ping you then to know if you still encounter issues after the pass.

vroulet:
#1060 solves #1038 and #337, see tests, so I think this PR may be closed once #1060 is merged.
Thank you again @hlzl for investigating this!
The optimizers that posed problems were adafactor, sm3 and novograd in the alias file, and most of the optimizers in the contrib file.
@mar-muel, what optimizer were you considering?
About
> It also seems like optax.MultiSteps is not very memory efficient in general, given that it keeps a full duplicate of the opt state :/
if you have some probing tools to assert that, it would be great. HLO is usually doing a very good job at reusing buffers when possible, but it indeed seems that ` state.inner_opt_state` and  `new_inner_state` may at one point occupy two different buffers in this code. If you don't have such probing tools, posting an issue would be great. Thanks for all!

mar-muel:
@vroulet Indeed the problem seems to only occur with adafactor, not with adam (using optax==0.2.2, without any patches)

vroulet:
@hlzl, @mar-muel, I'm closing this PR since #1060 has been merged and solved aforementioned issues.
Feel free to reopen it if I'm mistaken.

hlzl:
Hi @vroulet , great to see the updates on the optimizer-side in optax! 

I would still propose to additionally introduce the checks and conversions from this PR in `MultiSteps` as there might be differences in the dtypes due to external libraries. E.g. I had this problem originally with the dynamic scaler of flax.
You can adapt the original example to reproduce the problem when using flax' `dynamic_scale` as follows:
```python
import os

os.environ["JAX_PLATFORMS"] = "cpu"
import jax
import jax.numpy as jnp
import optax
from flax.training.dynamic_scale import DynamicScale


@jax.jit
def f(params, x, labels):
    logits = params @ x
    loss = optax.softmax_cross_entropy_with_integer_labels(logits, labels)
    return loss.mean()


dyn_scale = DynamicScale(). # changed
params = jnp.zeros((5, 18), dtype=jnp.bfloat16)
x = jnp.zeros((18, 4), dtype=jnp.bfloat16)
labels = jnp.zeros((5,), dtype=jnp.uint16)
dynamic_scale, is_fin, loss, grad = dyn_scale.value_and_grad(f)(params, x, labels). # changed

lr = 0.01
n_accumulation_steps = 4

optimizer = optax.sgd(learning_rate=lr)
optimizer = optax.MultiSteps(optimizer, n_accumulation_steps)
updates, opt_state = optimizer.update(grad, opt_state, params)
print(updates)
```

Let me know if we should thus reopen this PR or keep as is.

vroulet:
Hello @hlzl !
Beyond the bug in the MultiSteps function, what we want to avoid is having dtype promotions when applying the optimizer. In other words, a code like
```
grads = jax.grad(loss)(params)
updates, _ = opt.update(grads, state, params)
params = optax.apply_update(params, updates)
```
shall not induce in its last line a promotion of params.dtype. To me, any such piece of code that induces a promotion is a bug, and actually MultiSteps is catching somehow this bug.

However, optax should be able to handle mixed precision training. So in the code above, `optax.apply_update` can be modified while keeping dtypes of updates and params different. And all optimizers could also allow for dtypes that could be different of the dtypes of the params but we need to be careful to document such an approach (so that the user knows exactly that there may be a dtype promotion) with an adequate flag.

To conclude, I see what you mean, and we may incorporate such a change. But there is a more important underlying issue that needs to be documented (about mixed precision training). So if you could document clearly such an issue after reviewing some solutions like the ones discussed in #336, it would be great.

hlzl:
@vroulet I'm not sure I understand what you mean. 

Currently, in the case with the dynamic scaler it is not possible to run `optimizer.update(grad, opt_state, params)` even before applying any updates because `MultiSteps` is causing the same error as from #377 and #1038 .

The fix doesn't change the dtype of `params`, instead it adapts `grad` to be of the same dtype as `params` (which is necessary to run  `optimizer.update(...)` when using `MultiSteps`).
`updates` is then already of the same dtype as `params`.

E.g. adding the following loop to end of the previous example would fail in the first line without the fix but consistently returns `bfloat16` when using the proposed fix:
```python
for e in range(n_accumulation_steps * 2):
    updates, opt_state = optimizer.update(grad, opt_state, params)
    print(f"Step {e}: grad {grad.dtype} updates {updates.dtype} - params {params.dtype}") 
```

Note that this is a `MultiSteps` specific behavior. Commenting `MultiSteps` out, the script runs without an issue and returns `grad` and `updates` as `float32` while `params` is returned as `bfloat16`.

hlzl:
Ah, I see now. Is it possible that you mixed up the order of `updates` and `params` in `optax.apply_updates`? I think it should be `optax.apply_updates(params, updates)`, which also results in no promotion.

Other than that, I still think `updates` should be of the same dtype independent of `MultiSteps` being used or not. So my current proposed fix is not clean. I think a better alternative is to change how we initialize `updates` within `MultiSteps`. 

Let me know if changing the order of the arguments inside `apply_updates` solved your issue and I'll look into this.

vroulet:
Ah sorry indeed by setting `optax.apply_updates(params, updates)` it works. 
So yes, MultiSteps still has an issue I agree. It should mimic the behavior of apply_updates.

Sorry for the long discussion but this helps too! I think #1060 should probably be patched in light of that.
Namely, we do not necessarily want the updates of the type of params but we probably want them of the same dtype as the gradients. I'll open an issue to make sure #1060 aligns with this.

Thank you again! So you can reopen this PR or create a new one (I let you choose) with your fix and add corresponding tests.

vroulet:
Thanks @hlzl  for following up on this! First, we could schedule a meeting to chat about this directly, it could speed things up. Send me an email and let's schedule it.

> the update and more importantly the states (e.g. mu and nu) are always promoted to float32.

What are (i) the dtype of params, (ii) the dtype of gradients? Currently, the attributes of the state follow the dtype of the params. If `params_dtype` < `grads_dtype` (with < being defined by dtype promotion see e.g. the `tree_dtype` function for more explanation if it's not clear), the attributes of the state get promoted to the type of the grads after the first update (not ideal because it induces a recompilation). If `grads_dtype` < `params_dtype`, the dtype of the attributes may never match the dtype of the gradients (and the dtype of gradients will always be promoted to the dtype of params). I could make a whole pass on the codebase to prevent that. But so I'd like to know your use case.

Generally, this is a design issue of the init function of the gradient transformations (we should a priori give "fake" gradients rather than params to the init function). Mostly, it may be a simple renaming of the inputs in the init function for the user to understand that the init function should rather be fed gradients-like variables rather than params-like parameters. There may be some cases where parameters should still be fed and in that case I think one may consider modifying the init function to accept some additional arguments (just as GradientTransformationExtraArgs). This would then be a change in API that requires careful handling. But so again chatting together about your use case would be helpful (I have not considered mixed precision training in my work so some inputs from you will be great).

> Do you have an idea how I can change the dtype of states within e.g. ScaleByAdamState without doing a full update and increasing the count @vroulet ?

One could simply use a tree_map to convert dtypes as long as the leaves match the structure of the params. Where would you need that exactly?

> Maybe we can discuss here and I'll open a new one once it works as planned.

Sounds good, again could be good to discuss that "de vive voix"

hlzl:
@vroulet A meeting sounds good, I'll send you an email!

> we should a priori give "fake" gradients rather than params to the init function

This basically summarizes where the problems come from. Dealing with this is quite straight forward for most cases but a bit tricky when the dtypes change in the `inner_opt_state` in one of the two conditions (see promotion in https://github.com/hlzl/optax/blob/73a19a266f6d1aa6a61ff480bd30a2367a39c79a/optax/transforms/_accumulation.py#L344 vs. no change in https://github.com/hlzl/optax/blob/73a19a266f6d1aa6a61ff480bd30a2367a39c79a/optax/transforms/_accumulation.py#L377).

PR #1042, opened 2024-09-02
schedule_free: fix broadcasting of scalar arrays to 1d arrays

n-gao:
Currently, the momentum is stored in a 1D array `[b1]` of shape `(1,)`. We should store it instead in a scalar array `()`to avoid broadcasting scalars to `(1,)` in `schedule_free_eval_params`.

Before:
```python
opt = optax.contrib.schedule_free_adamw()
x = jnp.ones(())
state = opt.init(x)
optax.contrib.schedule_free_eval_params(state, x).shape
# (1,)
```

After:
```python
opt = optax.contrib.schedule_free_adamw()
x = jnp.ones(())
state = opt.init(x)
optax.contrib.schedule_free_eval_params(state, x).shape
# ()
```

### DIFF
--- a/optax/contrib/_schedule_free_test.py
+++ b/optax/contrib/_schedule_free_test.py
@@ -164,5 +164,16 @@ class ScheduleFreeTest(chex.TestCase):
     params_wrapper = run(opt_wrapper)
     chex.assert_trees_all_close(params_shortcut, params_wrapper)
 
+  @parameterized.parameters(*_OPTIMIZERS_UNDER_TEST)
+  def test_scalar_preservance(self, opt_name, opt_kwargs):
+    opt = getattr(alias, opt_name)(learning_rate=0.0, **opt_kwargs)
+    opt = _schedule_free.schedule_free(opt, learning_rate=0.0)
+
+    params = jnp.ones((), dtype=jnp.float32)
+    state = opt.init(params)
+    eval_params = _schedule_free.schedule_free_eval_params(state, params)
+    chex.assert_equal_shape([params, eval_params])
+    chex.assert_trees_all_equal_dtypes(params, eval_params)
+
 if __name__ == '__main__':
   absltest.main()
--- a/optax/contrib/_schedule_free_test.py
+++ b/optax/contrib/_schedule_free_test.py
@@ -166,8 +166,10 @@ class ScheduleFreeTest(chex.TestCase):
 
   @parameterized.parameters(*_OPTIMIZERS_UNDER_TEST)
   def test_scalar_preservance(self, opt_name, opt_kwargs):
-    opt = getattr(alias, opt_name)(learning_rate=0.0, **opt_kwargs)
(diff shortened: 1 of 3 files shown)


fabianp:
your solution seems reasonable to me. However, there are now some doctest errors in `optax/contrib/_schedule_free.py`, probably due to them now being included in the docs.

fabianp:
you might not need to build the docs (although if you wanted to, its described in the README). Just check the errors from the failing CI (https://github.com/google-deepmind/optax/actions/runs/10664454490/job/29555745356?pr=1042). as you can see, the issue seems to be that some examples in the docstrings use `schedule_free_eval_params` instead of the full name `optax.contrib.schedule...`

let me know if this doesn't make sense

n-gao:
I haven't touched any line related to `schedule_free_eval_params`. All the other lines that `docs/api/contrib.rst` also don't use complete paths? Let me check if the tests pass if I remove the lines again.

n-gao:
I don't get why the tests are failing. it works locally and the change seems unrelated. @fabianp do you have another idea?

fabianp:
you can undo the changes in docs/api/contrib.rst if you want since they are orthogonal to this PR

n-gao:
done

fabianp:
can you also add a test showing that the new approach doesn't have the broadcasting problem?

n-gao:
I added a test that fails before and succeeds after the PR

optax/contrib/_schedule_free_test.py
```diff
@@ -164,5 +164,16 @@ def run(opt):
     params_wrapper = run(opt_wrapper)
     chex.assert_trees_all_close(params_shortcut, params_wrapper)
 
+  @parameterized.parameters(*_OPTIMIZERS_UNDER_TEST)
+  def test_scalar_preservance(self, opt_name, opt_kwargs):
+    opt = getattr(alias, opt_name)(learning_rate=0.0, **opt_kwargs)
```
fabianp, line 169: perhaps call this base_opt ? otherwise both the base and the wrapper have the same name which is confusing

optax/contrib/_schedule_free_test.py
```diff
@@ -164,5 +164,16 @@ def run(opt):
     params_wrapper = run(opt_wrapper)
     chex.assert_trees_all_close(params_shortcut, params_wrapper)
 
+  @parameterized.parameters(*_OPTIMIZERS_UNDER_TEST)
+  def test_scalar_preservance(self, opt_name, opt_kwargs):
```
fabianp, line 168: the name `test_scalar_preservance` is not very descriptive. Please either use a more precise name for the test or add a comment below explaining what behavior the function is testing

n-gao:
Added a comment and changed the variable name. Though, this criticism probably applies to all the other tests in that file.

fabianp (approved):

fabianp:
excellent, thanks!

PR #1059, opened 2024-09-13
schedule-free optimier: ensure it's possible to donate both the state and the params

enolan:
Prior to this commit, with a schedule-free optimizer, if you tried to donate an entire train state, including both the parameters and a freshly initialized optimizer state, you'd get an "INVALID_ARGUMENT: Attempt to donate the same buffer twice in Execute()" error, because z was the same array as params. This commit fixes the issue and adds a test.

### DIFF
--- a/optax/contrib/_schedule_free.py
+++ b/optax/contrib/_schedule_free.py
@@ -145,6 +145,9 @@ def schedule_free(
       z = otu.tree_cast(params, dtype=state_dtype)
     else:
       z = params
+    # It's imporant to copy the params here so that z is a distinct array and
+    # we can donate both z and the params to JITted functions.
+    z = jax.tree_util.tree_map(lambda t: t.copy(), z)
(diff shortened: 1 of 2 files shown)


vroulet:
Hello @enolan,

Thank you for catching that!
Could you wait for #1060 to be merged ?
I can ping you back when it's merged into head, you'll merge your PR with head, then I'll review.

enolan:
sure no problem

vroulet:
Hello @enolan, 
Your PR should be good to go. Just sync with head, check if the problem persists without your change (i.e. whether #1060 changed the behavior), then upload, and I'll approve. 
Thanks again for looking into this!

enolan:
We should be good. I rebased and checked, the test still fails without the copy and passes with it.

vroulet (approved):
Thank you very much for looking into that.
The comment and the test are great

PR #1065, opened 2024-09-19
Add missing initializer argument of 0 to tree_reduce in tree_vdot and tree_sum.

carlosgmartin:
Currently, the following commands
```
$ python3 -c "import optax; optax.tree_utils.tree_sum({})"
$ python3 -c "import optax; optax.tree_utils.tree_vdot({}, {})"
```
 yield the following error:
```
TypeError: reduce() of empty iterable with no initial value
```
This PR fixes this bug by setting `initializer=0` in the call to [`tree_reduce`](https://jax.readthedocs.io/en/latest/_autosummary/jax.tree_util.tree_reduce.html#jax.tree_util.tree_reduce) inside each of these functions.

### DIFF
--- a/optax/tree_utils/_tree_math.py
+++ b/optax/tree_utils/_tree_math.py
@@ -151,7 +151,7 @@ def tree_vdot(tree_x: Any, tree_y: Any) -> chex.Numeric:
   numerical issues.
   """
   vdots = jax.tree.map(_vdot_safe, tree_x, tree_y)
-  return jax.tree.reduce(operator.add, vdots)
+  return jax.tree.reduce(operator.add, vdots, initializer=0)
 
 
 def tree_sum(tree: Any) -> chex.Numeric:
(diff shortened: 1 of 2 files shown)


vroulet:
Thanks! Can you add a quick test too? (just to bookkeep such edge cases in our tests in case we make a new similar function)

carlosgmartin:
@vroulet Done.

vroulet (approved):
Thanks!

fabianp:
thanks for the contribution @carlosgmartin !

do you mind to explicitly use the keyword argument ? that is, use 
```
 jtu.tree_reduce(operator.add, vdots, initializer=0)
```
instead of the current ``` jtu.tree_reduce(operator.add, vdots, 0)```. They're semantically equivalent but the first one is easier to read and less error-prone

carlosgmartin:
@fabianp Done.

vroulet (approved):

vroulet (approved):

PR #1068, opened 2024-09-23
Replace deprecated typing.Hashable with collections.abc.Hashable.

carlosgmartin:
According to the Python documentation, [typing.Hashable](https://docs.python.org/3/library/typing.html#typing.Hashable) is deprecated and should be replaced with [collections.abc.Hashable](https://docs.python.org/3/library/collections.abc.html#collections.abc.Hashable). This PR fixes optax accordingly.

### DIFF
--- a/docs/ext/coverage_check.py
+++ b/docs/ext/coverage_check.py
@@ -16,7 +16,8 @@
 
 import inspect
 import types
-from typing import Any, Mapping, Sequence, Tuple
+from collections.abc import Mapping
+from typing import Any, Sequence, Tuple
 
 import optax
 from sphinx import application
--- a/examples/cifar10_resnet.ipynb
+++ b/examples/cifar10_resnet.ipynb
@@ -44,7 +44,8 @@
       ],
       "source": [
         "import functools\n",
-        "from typing import Any, Callable, Sequence, Tuple, Optional, Dict\n",
+        "from collections.abc import Callable\n",
+        "from typing import Any, Sequence, Tuple, Optional, Dict\n",
         "\n",
         "from flax import linen as nn\n",
         "\n",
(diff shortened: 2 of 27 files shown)


vroulet (commented):
Thanks! I didn't know about PEP 585. 
In fact all Callable, Mapping, NamedTuple, etc... should also be imported from collections.abc too, if I understand well the docs https://docs.python.org/3/library/stdtypes.html#types-genericalias.
We can start with Hashable here, and you may add an issue for the other types to be treated later. If that's what you prefer, tell me by answering this comment and I'll approve this PR.
If you are up to make a whole pass on these deprecations in this PR, let me know, I'll let you do it and add the approval once the pass is done.
Thank you again @carlosgmartin !

carlosgmartin:
@vroulet I see a deprecation notice for

- [typing.Callable](https://docs.python.org/3/library/typing.html#typing.Callable)
- [typing.Mapping](https://docs.python.org/3/library/typing.html#typing.Mapping)

but not for

- [typing.NamedTuple](https://docs.python.org/3/library/typing.html#typing.NamedTuple)
- [typing.Union](https://docs.python.org/3/library/typing.html#typing.Union)

[collections.abc](https://docs.python.org/3/library/collections.abc.html) doesn't seem to have the latter two. Is that right? If so, I can edit the first two.

vroulet:
We keep Union as long as we support python 3.9 (we'll drop Union and Optional if we drop 3.9).
For NamedTuple, you're right. The docs https://docs.python.org/3/library/stdtypes.html#types-genericalias are a bit confusing:
> For heterogeneous collections of data where access by name is clearer than access by index, [collections.namedtuple()](https://docs.python.org/3/library/collections.html#collections.namedtuple) may be a more appropriate choice than a simple tuple object.
But we are using NamedTuple as a class, which is not supported by collections.namedtuple. Thanks for checking this.

carlosgmartin:
@vroulet Updated.

vroulet (approved):
Thanks for doing this!

Issue #1067, opened 2024-09-23
Running tests modifies .pylintrc and causes subsequent failures


carlosgmartin:
Running tests modifies the file `.pylintrc`, by appending the line
```
+disable=unnecessary-lambda-assignment,no-value-for-parameter,use-dict-literal
```
Running tests a _second_ time yields numerous failures. An example of this is shown below:

<details>

```zsh
(venv) $ JAX_VERSION= sh ./test.sh
+ rm -rf _testing
+ rm -rf dist/
+ rm -rf '*.whl'
+ rm -rf .pytype
+ mkdir -p _testing
+++ pwd
++ mktemp -d /Users/carlos/desktop/optax/_testing/optax-env.XXXXXXXX
+ readonly VENV_DIR=/Users/carlos/desktop/optax/_testing/optax-env.OHDqCZTC
+ VENV_DIR=/Users/carlos/desktop/optax/_testing/optax-env.OHDqCZTC
+ python3 -m venv /Users/carlos/desktop/optax/_testing/optax-env.OHDqCZTC
+ source /Users/carlos/desktop/optax/_testing/optax-env.OHDqCZTC/bin/activate
++ deactivate nondestructive
++ '[' -n '' ']'
++ '[' -n '' ']'
++ hash -r
++ '[' -n '' ']'
++ unset VIRTUAL_ENV
++ unset VIRTUAL_ENV_PROMPT
++ '[' '!' nondestructive = nondestructive ']'
++ '[' darwin23 = cygwin ']'
++ '[' darwin23 = msys ']'
++ export VIRTUAL_ENV=/Users/carlos/desktop/optax/_testing/optax-env.OHDqCZTC
++ VIRTUAL_ENV=/Users/carlos/desktop/optax/_testing/optax-env.OHDqCZTC
++ _OLD_VIRTUAL_PATH=/Users/carlos/venv/bin:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin
[...]

PR #1071, opened 2024-09-23
Fix test.sh to not modify .pylintrc.

carlosgmartin:
Addresses https://github.com/google-deepmind/optax/pull/1024#issuecomment-2367079688 and https://github.com/google-deepmind/optax/issues/1067 by editing `test.sh` to not modify or delete `.pylintrc`.

Also applies other improvements to `test.sh`:
- Do not enable [xtrace](https://www.gnu.org/software/bash/manual/bash.html#index-set) (`set -x`), to reduce the verbosity of the output
- Replace `pip` commands with `python3 -m pip`
  - https://github.com/python/cpython/pull/24003
  - https://github.com/python/cpython/issues/66491
- Replace `` `command` `` with `$(command)`
  - [Why is $(...) preferred over \`...\` (backticks)?](http://mywiki.wooledge.org/BashFAQ/082)

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -397,3 +397,4 @@ valid-classmethod-first-arg=cls,
 
 # List of valid names for the first argument in a metaclass class method.
 valid-metaclass-classmethod-first-arg=mcs
+disable=unnecessary-lambda-assignment,no-value-for-parameter,use-dict-literal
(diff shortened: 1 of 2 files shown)


fabianp (approved):
thanks!

Issue #954, opened 2024-04-27
Add an assignment problem solver


carlosgmartin:
**Feature request:** Add a GPU/TPU-friendly solver for the [assignment problem](https://en.wikipedia.org/wiki/Assignment_problem). For context, see:

1. [scipy.optimize.linear_sum_assignment](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linear_sum_assignment.html)
2. https://github.com/google/jax/issues/10403
3. https://github.com/google/jax/pull/16974

The last page contains the following comment:

> There is a TPU-friendly implementation of the [Hungarian algorithm](https://en.wikipedia.org/wiki/Hungarian_algorithm) here: https://github.com/google-research/scenic/blob/main/scenic/model_lib/matchers/hungarian_cover.py

Potentially relevant literature:
- [A distributed auction algorithm for the assignment problem](https://ieeexplore-ieee-org.cmu.idm.oclc.org/document/4739098) (2008)
- [Bipartite Graph Matching Computation on GPU](https://link-springer-com.cmu.idm.oclc.org/chapter/10.1007/978-3-642-03641-5_4) (2009)
- [An anytime assignment algorithm: From local task swapping to global optimality](https://link.springer.com/article/10.1007/s10514-013-9351-2) (2013)
- [On implementing 2D rectangular assignment algorithms](https://ieeexplore.ieee.org/document/7738348) (2016)
- [GPU-accelerated Hungarian algorithms for the Linear Assignment Problem](https://www.sciencedirect.com/science/article/abs/pii/S016781911630045X) (2016)
- [Fast block distributed CUDA implementation of the Hungarian algorithm](https://www.sciencedirect.com/science/article/pii/S0743731519302254) (2019)
- [Development of an Accelerating Hungarian Method for Assignment Problems](https://papers.ssrn.com/sol3/papers.cfm?abstract_id=3724157) (2021)
- [Algorithm 1015: A Fast Scalable Solver for the Dense Linear (Sum) Assignment Problem](https://dl.acm.org/doi/abs/10.1145/3442348) (2021)
- [New auction algorithms for the assignment problem and extensions](https://www.sciencedirect.com/science/article/pii/S2666720724000134) (2024)

carlosgmartin:
@vroulet What would be the best place to add a solver function? `optax.contrib.linear_assignment_problem`? I can create a draft PR with my current implementation.

vroulet:
Hello @carlosgmartin, 
[...]

PR #1083, opened 2024-10-02
Add Hungarian algorithm for the linear assignment problem.

carlosgmartin:
#954


fabianp:
thanks for starting this @carlosgmartin !

At a high level this looks good to me, but it would be important to have an example of this method in the gallery (https://optax.readthedocs.io/en/latest/gallery.html) to showcase its usage

vroulet (commented):
Thanks for integrating this
Here are a few comments

optax/assignment/_hungarian_algorithm.py
```diff
@@ -0,0 +1,291 @@
...
+# ==============================================================================
+"""The Hungarian algorithm for the linear assignment problem."""
+
+from functools import partial
+
+import jax
+from jax import lax, numpy as jnp
+
+
+def hungarian_algorithm(costs, /):
```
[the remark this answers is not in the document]
carlosgmartin: There's only 1 argument and I was undecided between calling it `costs`, `cost_matrix`, `matrix`, `cost_array`, `array`, etc. so I figured we can just keep the API name-independent for now.

If you prefer a specific argument name, let me know.

optax/assignment/_hungarian_algorithm.py
```diff
@@ -0,0 +1,291 @@
...
+  )
+
+  if transpose:
+    i = col4row.argsort()
+    return col4row[i], i
+  else:
+    return jnp.arange(costs.shape[0]), col4row
+
+
+def _find_short_augpath_while_body_inner_for(it, val):
```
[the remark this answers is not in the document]
carlosgmartin: Added a reference to the paper with the mathematical expressions: https://ieeexplore.ieee.org/document/7738348

carlosgmartin:
@fabianp @vroulet Done.

optax/assignment/_hungarian_algorithm.py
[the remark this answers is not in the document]
vroulet: I see. Interesting solution, I'll keep it in mind in the future.
I think for now it's ok to use `cost_matrix` (Wikipedia uses such term so we'll consider it standard).
I'd prefer a name (potentially not optimal) than an uncommon signature.

optax/assignment/_hungarian_algorithm_test.py
```diff
@@ -0,0 +1,88 @@
...
+import optax
+
+
+class HungarianAlgorithmTest(parameterized.TestCase):
+
+  @parameterized.product(
+    n=[0, 1, 2, 4, 8, 16],
+    m=[0, 1, 2, 4, 8, 16],
+  )
+  def test_hungarian_algorithm(self, n, m):
```
vroulet, line 30: It would be great to see whether this algorithm can support usual transformations (like jitting or vmap).

Side note beyond this PR:
It could be great to have this building block differentiable. To be exact: there is no way the current outputs can be differentiable (they are ints). But the resulting cost could be differentiable. The advantage is that people may then parameterize the cost matrix, run the algorithm to get the optimal cost and then differentiate through the algorithm to further refine their cost matrix.
This would not be super hard actually to let your function return the optimal cost so that a user could easily make it differentiable. 
Anyway, it's not very important, so you can dismiss this comment

optax/assignment/_hungarian_algorithm.py
```diff
@@ -0,0 +1,350 @@
...
+    v,
+    shortest_path_costs,
+    path,
+    lowest,
+    row4col,
+    index,
+  )
+
+
+def _find_short_augpath_while_body_tail_alt(val):
```
vroulet, line 181: All names of functions are very well chosen, which makes the code nicely readable. An exception is "alt" here: is it for alternative? Either change this word to something readable or add a comment below

vroulet (commented):
Thanks again! A few minor remaining comments

optax/assignment/_hungarian_algorithm.py
carlosgmartin: Ah, that's supposed to be just `_find_short_augpath_while_body_tail`. (I was testing an alternative implementation.)

optax/assignment/_hungarian_algorithm_test.py
carlosgmartin: Currently, the algorithm has 2 occurrences of [lax.while_loop](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.while_loop.html), which is not reverse-mode differentiable.

In theory, these could be replaced with a [bounded while_loop](https://github.com/jax-ml/jax/issues/2469). I don't think that's been ported [from jaxopt](https://github.com/google/jaxopt/blob/main/jaxopt/_src/loop.py) yet, has it?

On the other hand, it seems to me that, for [almost all](https://en.wikipedia.org/wiki/Almost_all) inputs and infinitesimal perturbations, the (integer) solution indices remain unchanged, and thus differentiating the solution cost is equivalent to simply differentiating `cost_matrix[i_solution, j_solution]`.

carlosgmartin:
@vroulet Updated.

optax/assignment/_hungarian_algorithm_test.py
vroulet: > has 2 occurrences of lax.while_loop

Ok, I should have been more careful, this is indeed an impediment for an efficient differentiation.
Anyway, we may think about that later, the current implementation should already be useful for many use cases.
And thanks for checking the interoperability with jit and vmap!

vroulet (approved):
Great work, looks good to me

vroulet:
Yes, that's clearly worth it :)
Ideally we would like your effort to be adopted so such comments are good.
Also, I was wondering, your implementation is neat, but aren't there other implementations potentially more tpu friendly? (not very important, but one may mention it if it's the case? I'm not an expert on that, packages in optimal transport may have some clever tricks).

carlosgmartin:
@vroulet Right, there are faster specialized algorithms for approximating the Wasserstein distance (see [OTT-JAX](https://github.com/ott-jax/ott)).

Perhaps I can append the example above to the [LAP gallery entry](https://optax.readthedocs.io/en/latest/_collections/examples/linear_assignment_problem.html)?

JTT94:
How does this compare to `hungarian_tpu_matcher`? from 
https://github.com/google-research/scenic/blob/main/scenic/model_lib/matchers/hungarian_jax.py

carlosgmartin:
@vroulet Submitted a PR: #1140.

PR #1088, opened 2024-10-03
Fix initial step of scale_by_optimistic_gradient.

### LINKED ISSUE #1082: Possible issue with the initial step of `scale_by_optimistic_gradient`
There's a possible issue with the current implementation of [`scale_by_optimistic_gradient`](https://optax.readthedocs.io/en/latest/api/transformations.html#optax.scale_by_optimistic_gradient). At the initial step, it acts as if the previous gradient was zero. However, the previous gradient doesn't actually exist (zero is just a placeholder there). Not accounting for this causes it to make an abnormal "jump" at the initial step.

The example below illustrates this phenomenon:

<details>

```python3
import argparse
from typing import NamedTuple

import chex
import jax
import optax
from jax import lax, numpy as jnp
from matplotlib import pyplot as plt, rcParams


class ScaleByOptimisticGradientState(NamedTuple):
    is_initial_step: chex.Array  # shape=(), dtype=bool.
    prev_grad: optax.Updates


def scale_by_optimistic_gradient_fixed(alpha=1.0, beta=1.0):
    def init_fn(params):
        return ScaleByOptimisticGradientState(
            is_initial_step=jnp.array(True),
            prev_grad=optax.tree_utils.tree_zeros_like(params),
        )

        del params

        new_updates = jax.tree.map(
            lambda grad_t, grad_tm1: jnp.where(
                state.is_initial_step,
                alpha * grad_t,
                (alpha + beta) * grad_t - beta * grad_tm1,
            ),
            updates,
            state.prev_grad,
        )

        new_state = ScaleByOptimisticGradientState(
            is_initial_step=jnp.array(False),
            prev_grad=updates,
        )

        return new_updates, new_state

    return optax.GradientTransformation(init_fn, update_fn)


def optimistic_sgd(learning_rate, strength):
    return optax.scale_by_optimistic_gradient(-learning_rate, -strength)


def optimistic_sgd_fixed(learning_rate, strength):
    return scale_by_optimistic_gradient_fixed(-learning_rate, -strength)


def parse_args():
    p = argparse.ArgumentParser()
    p.add_argument("--lr", type=float, default=1e-3)
    p.add_argument("--iters", type=int, default=10**3)
    p.add_argument("--strength", type=float, default=1e-1)
    return p.parse_args()


def main():
    args = parse_args()

    def loss_fn(params):
        return jnp.square(params[0]) + jnp.square(params[1] * 2)

    def update(state, _):
        params, opt_state = state
        grads = jax.grad(loss_fn)(params)
        updates, new_opt_state = opt.update(grads, opt_state, params)
        new_params = optax.apply_updates(params, updates)
        return (new_params, new_opt_state), params

    _, ax_distances = plt.subplots()
    _, ax_params = plt.subplots()
    params = jnp.array([2.0, 1.0])
    ax_params.scatter(*params, c="red", label="start")

    for label, opt in [
        ("SGD", optax.sgd(args.lr)),
        ("Optimistic SGD", optimistic_sgd(args.lr, args.strength)),
        (
            "Optimistic SGD (fixed)",
            optimistic_sgd_fixed(args.lr, args.strength),
        ),
    ]:
        opt_state = opt.init(params)
        _, params_hist = lax.scan(
            update, (params, opt_state), length=args.iters
        )
        distances_to_origin = jnp.hypot(*params_hist.T)
        ax_params.plot(*params_hist.T, label=label)
        ax_distances.plot(distances_to_origin, label=label, lw=1)

    ax_params.legend()
    ax_distances.legend()
    ax_params.set(title="parameters", aspect="equal")
    ax_distances.set(
        xlabel="iteration", ylabel="distance to origin", yscale="log"
    )
    rcParams["savefig.dpi"] = 300
    plt.show()


if __name__ == "__main__":
    main()
```

</details>

Output:

Notice how the initial step is abnormally large, and also causes subsequent jitters.

My implementation above, `scale_by_optimistic_gradient_fixed`, fixes this by checking whether we're at the initial step and, if so, acting as if `beta` is zero, i.e., taking an ordinary SGD step instead. (This is also equivalent to letting `grad_tm1 = grad_t` at the initial step.) This seems to fix the initial step and reduce subsequent jittering.

This also makes the algorithm invariant to whatever the initial "previous gradient" (actually a placeholder) is set to, which conceptually feels like the right thing to do.

If this makes sense, I can submit a PR to fix `scale_by_optimistic_gradient`.

carlosgmartin:
Fixes #1082.

### DIFF
--- a/optax/_src/transform.py
+++ b/optax/_src/transform.py
@@ -1183,6 +1183,11 @@ def scale_by_novograd(
   return base.GradientTransformation(init_fn, update_fn)
 
 
+class ScaleByOptimisticGradientState(NamedTuple):
+  is_initial_step: chex.Array
+  previous_gradient: base.Updates
+
+
 def scale_by_optimistic_gradient(
     alpha: float = 1.0,
     beta: float = 1.0
@@ -1201,15 +1206,29 @@ def scale_by_optimistic_gradient(
   """
 
   def init_fn(params):
-    return TraceState(trace=otu.tree_zeros_like(params))
+    return ScaleByOptimisticGradientState(
+        is_initial_step=jnp.array(True),
+        previous_gradient=otu.tree_zeros_like(params),
+    )
 
   def update_fn(updates, state, params=None):
     del params
 
-    new_updates = jax.tree.map(
-        lambda grad_t, grad_tm1: (alpha + beta) * grad_t - beta * grad_tm1,
-        updates, state.trace)
-    return new_updates, TraceState(trace=updates)
+    def f(grad_t, grad_tm1):
+      # At the initial step, the previous gradient doesn't exist, so we use the
+      # current gradient instead.
+      # https://github.com/google-deepmind/optax/issues/1082
+      grad_tm1 = jnp.where(state.is_initial_step, grad_t, grad_tm1)
+      return (alpha + beta) * grad_t - beta * grad_tm1
+
+    new_updates = jax.tree.map(f, updates, state.previous_gradient)
+
+    new_state = ScaleByOptimisticGradientState(
+        is_initial_step=jnp.array(False),
+        previous_gradient=updates,
+    )
+
+    return new_updates, new_state
 
   return base.GradientTransformation(init_fn, update_fn)
 
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -1303,11 +1303,11 @@ def optimistic_gradient_descent(
     ...  updates, opt_state = solver.update(grad, opt_state, params)
     ...  params = optax.apply_updates(params, updates)
     ...  print('Objective function: {:.2E}'.format(f(params)))
+    Objective function: 1.38E+01
     Objective function: 1.37E+01
     Objective function: 1.35E+01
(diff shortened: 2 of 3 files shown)


optax/_src/transform.py
```diff
@@ -1201,15 +1206,30 @@ def scale_by_optimistic_gradient(
...
+    )
 
   def update_fn(updates, state, params=None):
     del params
 
     new_updates = jax.tree.map(
-        lambda grad_t, grad_tm1: (alpha + beta) * grad_t - beta * grad_tm1,
-        updates, state.trace)
-    return new_updates, TraceState(trace=updates)
+        lambda grad_t, grad_tm1: jnp.where(
```
fabianp, line 1218: could you please add a comment explaining the initial step issue and why the jnp.where is necessary ?

carlosgmartin: @fabianp Done.

fabianp:
excellent, thanks!

fabianp (approved):

fabianp (approved):

PR #1094, opened 2024-10-07
update jax.tree.map to comply with jax 0.4.34

### LINKED ISSUE #1093: Jax 0.4.34 raises an error for None as tree-prefix
Hello:)

[The most recent release of jax 0.4.34](https://github.com/jax-ml/jax/releases/tag/jax-v0.4.34) does no longer support anything of the kind `jax.tree.map(f, None, non-None)`. 

This also affects optax, for example, in [add_decayed_weights](https://github.com/google-deepmind/optax/blob/main/optax/transforms/_adding.py#L50). 

The error can be simply prevented by changing the line to 

```
    updates = jtu.tree_map(
        lambda g, p: None if g is None else g + weight_decay * p, updates, params, is_leaf=lambda x: x is None)
```

I'm happy to take a look and see if there are any other parts affected and provide a pull request if desired. 

Best, 
Alex

a1302z:
Pull Request to fix #1093

### DIFF
--- a/optax/_src/update_test.py
+++ b/optax/_src/update_test.py
@@ -79,5 +79,9 @@ class UpdateTest(chex.TestCase):
           params_2, new_params, atol=1e-10, rtol=1e-5)
 
 
+  def test_none_argument(self):
+    update.apply_updates(None, jnp.array([1.,2.,3.]))
+    update.incremental_update(None, jnp.array([1.,2.,3.]), 0.1)
+
 if __name__ == '__main__':
   absltest.main()
--- a/optax/_src/update_test.py
+++ b/optax/_src/update_test.py
@@ -14,7 +14,7 @@
 # ==============================================================================
 """Tests for `update.py`."""
 
-from absl.testing import absltest
+from absl.testing import absltest, parameterized
 
 import chex
 import jax
@@ -78,10 +78,14 @@ class UpdateTest(chex.TestCase):
       chex.assert_trees_all_close(
           params_2, new_params, atol=1e-10, rtol=1e-5)
 
-
-  def test_none_argument(self):
-    update.apply_updates(None, jnp.array([1.,2.,3.]))
-    update.incremental_update(None, jnp.array([1.,2.,3.]), 0.1)
+  @parameterized.named_parameters(
+      dict(testcase_name='apply_updates', operation=update.apply_updates),
+      dict(testcase_name='incremental_update',
+           operation=lambda x, y: update.incremental_update(x,y,1)),
+  )
+  def test_none_argument(self, operation):
+    x = jnp.array([1., 2., 3.])
+    operation(None, x)
 
 if __name__ == '__main__':
   absltest.main()
--- a/optax/transforms/_adding_test.py
+++ b/optax/transforms/_adding_test.py
@@ -91,14 +91,14 @@ class AddingTest(chex.TestCase):
 
       chex.assert_trees_all_close(updates_i, updates_i_rescaled, rtol=1e-4)
 
-def test_none_argument():
-  weights = (
-      jnp.ones((2,), dtype=jnp.float32),
-      dict(
-          a=jnp.ones((2,), dtype=jnp.float32),
-          b=jnp.ones((2,), dtype=jnp.float32),))
-  tf = _adding.add_decayed_weights(0.1, mask=None)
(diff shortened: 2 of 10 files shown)


vroulet (commented):
Thank you for doing that!
Would it be possible to add a generic test for such functions?
(tree_update_moment, tree_update_infinity_moment and tree_update_moment_per_elem_norm follow the same fix).
Thanks again!

a1302z:
Happy to do that. How would you imagine this test to be fully generic? A simple test case would be something like this (extended to all corresponding functions):
```
  def test_none_arguments(self):
    m = jnp.logspace(-10, 10, num=21, dtype=jnp.bfloat16)  # 1e-10 ... 1e10
    tu.tree_update_moment(None, m, 0.9, 1)
```
Let me know what you think.

a1302z:
I have added test cases for `tree_add_scalar_mul`, `tree_update_moment`, `tree_update_infinity_moment`, `tree_update_moment_per_elem_norm`, `apply_updates`, `incremental_update`, `keep_params_nonnegative`, and `add_decayed_weights`. I have omitted `scale_by_adam` as mu would need to be None, which is not an input argument to the function.

optax/_src/update_test.py
```diff
@@ -79,5 +79,9 @@ def test_periodic_update(self):
           params_2, new_params, atol=1e-10, rtol=1e-5)
 
 
+  def test_none_argument(self):
```
vroulet, line 82: Make it a parameterized test, that is, add `from absl.testing import parameterized` at the top and then write the test as
```
  @parameterized.named_parameters(
      dict(testcase_name='apply_updates', operation=update.apply_updates),
      dict(testcase_name='incremental_update', operation=update.incremental_update),
  )
    def test_none_argument(self, operation)
      x = jnp.array([1., 2., 3.])
      operation(None, x)
  ```

File: optax/transforms/_adding_test.py [hunk 2]
```diff
@@ -91,6 +91,14 @@ def test_add_noise_has_correct_variance_scaling(self):
 
       chex.assert_trees_all_close(updates_i, updates_i_rescaled, rtol=1e-4)
 
+def test_none_argument():
```
vroulet, line 94: Add an indent (the test must be part of the class AddingTest) and add a self argument to the function.

optax/transforms/_constraining_test.py
```diff
@@ -114,6 +114,15 @@ def test_zero_nans(self):
             (jnp.array(False), jnp.array(False), jnp.array(False))))
     chex.assert_trees_all_close(updates, grads)
 
+  def check_none_arguments(self):
```
vroulet, line 117: call it `test_none_arguments`

vroulet: Also why are you testing for this to raise an error?

optax/tree_utils/_tree_math_test.py
```diff
@@ -236,6 +236,13 @@ def test_empty_tree_reduce(self):
       self.assertEqual(tu.tree_sum(tree), 0)
       self.assertEqual(tu.tree_vdot(tree, tree), 0)
 
+  def test_none_arguments(self):
```
vroulet, line 239: Make it a parameterized test with the different functions

vroulet (commented):
Thank you very much, I left you some additional comments

optax/transforms/_constraining_test.py
a1302z: Done. In this function is a check that raises a ValueError if None is passed as params. Hence, the tree error can only occur if this check is, for some reason, disabled. If we rely on this check not being modified, we can also remove this test case entirely.

optax/_src/update_test.py
a1302z: Done. However, the functions take a different number of arguments. To make it compatible I've used a lambda expression. Also below for tree_math_test. Let me know if this is fine or if there is a better way to deal with that.

vroulet: That's perfect, thank you!

optax/transforms/_constraining_test.py
vroulet: Right, perfect, thank you for the explanation!

vroulet (approved):
That's perfect, thank you again @a1302z !

PR #1095, opened 2024-10-09
Fix formatting issues with gallery entry for linear assignment problem.

carlosgmartin:
Looks like https://optax.readthedocs.io/en/latest/_collections/examples/linear_assignment_problem.html has some formatting issues. This PR should fix it.


vroulet (approved):
Thank you for fixing that!

PR #1096, opened 2024-10-09
use sphinx references instead of hardcoded links.

fabianp:
several advantages:

  * looks nicer
  * doesn't break when the links change (and warns if the endpoint changes)
  * references the current version of the documentation (e.g., local links when building locally, and the latest version when building on readthedocs etc.)


fabianp:
FYI @carlosgmartin  since I tweaked a bit your last example 😄

vroulet (approved):
Thanks!

PR #1097, opened 2024-10-09
cosmetic improvements perturbations module

fabianp:
* reference example from api reference
  * add title and colab link to example notebook


q-berthet (approved):
Thanks!

PR #1108, opened 2024-10-17
Avoid re-compilation and allow non-float32 dtypes in the zoom_linesearch

mirkobunse:
# Problems

I experienced two problems that the zoom_linesearch exhibits when used within a JIT-compiled training step:

1) the training step is compiled twice because some dtypes of the linesearch's state change after the first iteration.
2) the compilation errors for dtypes other than float32.

I want to give a minimal example to reproduce both problems and I want to provide a work-around for the first one. The second problem cannot be worked around; hence, I propose this PR to ultimately solve both problems.

# Minimal working example for the re-compilation issue

I sample some random data and create an LBGFS optimizer with zoom linesearch. The training step is JIT-compiled and a `print` statement informs us about the triggering of a compilation (because this statement is silent in the compiled variant of the function).

```py
import jax.numpy as jnp
import jax
import optax
from time import time

key = jax.random.PRNGKey(42) # similar to the getting started guide
x = jax.random.normal(key, (16, 2))
y = jnp.sum(x * 0.5, axis=-1)
optimizer = optax.lbfgs()

@jax.jit
def training_step(params, state):
  print("I'm being compiled for", state[-1].info)
  def loss_fn(params):
    y_pred = (params @ x.T).T # a linear model
    loss = jnp.mean(optax.l2_loss(y_pred, y))
    return loss
  loss, grad = jax.value_and_grad(loss_fn)(params)
  updates, state = optimizer.update(
    grad,
    state,
    params,
    value = loss,
    grad = grad,
    value_fn = loss_fn,
  )
  return params, state

params = jnp.array([0.0, 0.0])
state = optimizer.init(params)

t0 = time()
for _ in range(10):
  params, state = training_step(params, state)
print(f"Took {time() - t0} sec")
```

The output of this script is the following. We recognize that the function is compiled twice and that, between iterations, dtypes have changed from weak types to strong types.

```
I'm being compiled for ZoomLinesearchInfo(num_linesearch_steps=Traced<ShapedArray(int32[], weak_type=True)>with<DynamicJaxprTrace(level=1/0)>, decrease_error=Traced<ShapedArray(float32[], weak_type=True)>with<DynamicJaxprTrace(level=1/0)>, curvature_error=Traced<ShapedArray(float32[], weak_type=True)>with<DynamicJaxprTrace(level=1/0)>)
I'm being compiled for ZoomLinesearchInfo(num_linesearch_steps=Traced<ShapedArray(int32[])>with<DynamicJaxprTrace(level=1/0)>, decrease_error=Traced<ShapedArray(float32[])>with<DynamicJaxprTrace(level=1/0)>, curvature_error=Traced<ShapedArray(float32[])>with<DynamicJaxprTrace(level=1/0)>)
Took 0.7838089466094971 sec
```

# Workaround for the re-compilation issue

We can change the initial state to have strong dtypes right from the beginning:

```py
state = optimizer.init(params) # as before

state = ( # replace occurences of weak_type=True with strongly typed traces.
  *state[:-1], # all but the last item of this state are good
  optax._src.linesearch.ScaleByZoomLinesearchState(
    learning_rate = state[-1].learning_rate, # will not become strong -> can stay weak
    value = jnp.asarray(state[-1].value, dtype=state[-1].value.dtype), # infer dtype
    grad = state[-1].grad, # is strong already
    info = optax._src.linesearch.ZoomLinesearchInfo(
        num_linesearch_steps = jnp.asarray(
          state[-1].info.num_linesearch_steps,
          dtype = jnp.int32 # the original dtype is not the final one -> fix beforehand
        ),
        decrease_error = jnp.asarray(
          state[-1].info.decrease_error,
          dtype = jnp.float32
        ),
        curvature_error = jnp.asarray(
          state[-1].info.curvature_error,
          dtype = jnp.float32
        ),
    ),
  )
)
```

As desired, the training step is now compiled only once.

```
I'm being compiled for ZoomLinesearchInfo(num_linesearch_steps=Traced<ShapedArray(int32[])>with<DynamicJaxprTrace(level=1/0)>, decrease_error=Traced<ShapedArray(float32[])>with<DynamicJaxprTrace(level=1/0)>, curvature_error=Traced<ShapedArray(float32[])>with<DynamicJaxprTrace(level=1/0)>)
Took 0.4254724979400635 sec
```

# Minimal working example for non-default dtypes

In the above example, we make two slight changes,

```diff
- x = jax.random.normal(key, (16, 2))
+ x = jax.random.normal(key, (16, 2), dtype=jnp.bfloat16)
```

and

```diff
- params = jnp.array([0.0, 0.0])
+ params = jnp.array([0.0, 0.0], dtype=jnp.bfloat16)
```

Unfortunately, these changes break the compilation entirely. Since this error, shown in the following, appears in the internals of the zoom linesearch, I do not see an easy work-around for this problem.

```
...

  File "/home/bunse/Repos/optax/optax/_src/linesearch.py", line 1167, in step_fn
    new_state = jax.lax.cond(
                ^^^^^^^^^^^^^
TypeError: true_fun and false_fun output must have identical types, got
ZoomLinesearchState(count='ShapedArray(int32[])', params='ShapedArray(bfloat16[2])', updates='ShapedArray(bfloat16[2])', stepsize_guess='ShapedArray(float32[], weak_type=True)', stepsize='ShapedArray(float32[])', value='ShapedArray(bfloat16[])', grad='ShapedArray(bfloat16[2])', slope='ShapedArray(bfloat16[])', value_init='ShapedArray(bfloat16[])', slope_init='ShapedArray(bfloat16[])', decrease_error='DIFFERENT ShapedArray(float32[]) vs. ShapedArray(bfloat16[])', curvature_error='ShapedArray(bfloat16[])', error='DIFFERENT ShapedArray(float32[]) vs. ShapedArray(bfloat16[])', interval_found='ShapedArray(bool[])', done='ShapedArray(bool[])', failed='ShapedArray(bool[])', low='ShapedArray(float32[])', value_low='ShapedArray(bfloat16[])', slope_low='ShapedArray(bfloat16[])', high='ShapedArray(float32[])', value_high='ShapedArray(bfloat16[])', slope_high='ShapedArray(bfloat16[])', cubic_ref='ShapedArray(float32[], weak_type=True)', value_cubic_ref='ShapedArray(bfloat16[])', safe_stepsize='ShapedArray(float32[])', safe_value='ShapedArray(bfloat16[])', safe_grad='ShapedArray(bfloat16[2])')
```

# Solution: this PR

I suggest the following improvements:

- arbitrary dtypes can be used as parameters and data within JIT-compiled zoom linesearches (solves problem 2)
- users can optionally specify an expected dtype returned by the value function, to prevent a function from being re-compiled (solves problem 1)

The specification of a value dtype works as follows:

```diff
- optimizer = optax.lbfgs()
+   optimizer = optax.lbfgs(
+   linesearch = optax.scale_by_zoom_linesearch(15, value_dtype=jnp.bfloat16)
+ )
```

### DIFF
--- a/docs/api/projections.rst
+++ b/docs/api/projections.rst
@@ -37,6 +37,7 @@ Available projections
     projection_l1_sphere
     projection_l2_ball
     projection_l2_sphere
+    projection_linf_ball
     projection_non_negative
     projection_simplex
 
@@ -63,6 +64,10 @@ Projection onto the L2 ball
 Projection onto the L2 sphere
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 .. autofunction:: projection_l2_sphere
+
+Projection onto the L-infinity ball
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. autofunction:: projection_linf_ball
 
 Projection onto the non-negative orthant
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/optax/projections/__init__.py
+++ b/optax/projections/__init__.py
@@ -21,5 +21,6 @@ from optax.projections._projections import projection_l1_ball
 from optax.projections._projections import projection_l1_sphere
 from optax.projections._projections import projection_l2_ball
 from optax.projections._projections import projection_l2_sphere
+from optax.projections._projections import projection_linf_ball
 from optax.projections._projections import projection_non_negative
 from optax.projections._projections import projection_simplex
--- a/optax/projections/_projections.py
+++ b/optax/projections/_projections.py
@@ -277,3 +277,26 @@ def projection_l2_ball(pytree: Any, scale: float = 1.0) -> Any:
                       lambda pytree: pytree,
                       lambda pytree: otu.tree_scalar_mul(factor, pytree),
                       operand=pytree)
+
+
+def projection_linf_ball(pytree: Any, scale: float = 1.0) -> Any:
+  r"""Projection onto the l-infinity ball.
+
+  This function solves the following constrained optimization problem,
+  where ``x`` is the input pytree.
+
+  .. math::
+
+    \underset{y}{\text{argmin}} ~ ||x - y||_2^2 \quad \textrm{subject to} \quad
+    ||y||_{\infty} \le \text{scale}
+
+  Args:
+    pytree: pytree to project.
+    scale: radius of the ball.
+
+  Returns:
+    projected pytree, with the same structure as ``pytree``.
+  """
+  lower_tree = otu.tree_full_like(pytree, -scale)
+  upper_tree = otu.tree_full_like(pytree, scale)
+  return projection_box(pytree, lower=lower_tree, upper=upper_tree)
--- a/optax/projections/_projections_test.py
+++ b/optax/projections/_projections_test.py
@@ -43,6 +43,7 @@ class ProjectionsTest(parameterized.TestCase):
     self.fns = dict(
         l1=(proj.projection_l1_ball, otu.tree_l1_norm),
         l2=(proj.projection_l2_ball, otu.tree_l2_norm),
+        linf=(proj.projection_linf_ball, otu.tree_linf_norm),
     )
 
   def test_projection_non_negative(self):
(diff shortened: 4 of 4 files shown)


vroulet:
Hello @mirkobunse ,

Thanks for the fix! There is a larger issue in optax: any optimizer with mixed precision (different precision with gradients and parameters) may compile twice. A fix would probably be to change the signature of the init functions rather than adding arguments to the definition of the optimizer. We will discuss that internally soon and I'll keep you posted.

If this is blocking your research you may also consider [optimistix](https://github.com/patrick-kidger/optimistix) which carefully took care of some recompilation issues

PR #1122, opened 2024-10-28
fix: Coherent dtypes of updates with and without MultiSteps

hlzl:
Following #1039 and discussions with @vroulet a proposed fix such that updates with and without `MultiSteps` have the same dtypes independent of how `acc_grads` and the `inner_opt_state` were initialized.


vroulet (approved):
Very nice, thank you @hlzl !
By curiosity did you check the speed before and after your change?
I think it should be the same but would be good to check just to be sure.

In any case, I'll approve this, thank you for all the effort!

PR #1126, opened 2024-11-02
Add Muon Optimizer to `contrib`

leloykun:
Adds support for @KellerJordan's Muon optimizer as detailed here: https://kellerjordan.github.io/posts/muon/

This optimizer allowed us to reduce the training time of a GPT-2 level model down to just ~3.5 minutes on 8xH100s. See: https://github.com/KellerJordan/modded-nanogpt

image

This implementation also supports the adaptive version of Muon that can adapt to the scale of the gradients as they change during training. To enable this feature, simply set `muon(adaptive=True, ...)`. See more at: https://github.com/leloykun/adaptive-muon

---

Muon is equivalent to Shampoo without accumulation. I.e., take Shampoo's update rule, discard the left and right preconditioners, then simplify the matrix operations. Alternatively, Muon can also be thought of as steepest descent under Spectral norm. For more details, please refer to @jxbz's paper https://arxiv.org/pdf/2409.20325.

More generally, steepest descent under Schatten-p norm, for some large p, can be thought of as variants of Muon (note that Schatten-infty norm = Spectral norm). A (rough) proof of this claim can be found here: https://x.com/leloykun/status/1846842887839125941. This implementation supports these variants by allowing the user to pass in custom coefficients for the newton-schulz iterations.

### DIFF
--- a/optax/contrib/_muon.py
+++ b/optax/contrib/_muon.py
@@ -1,0 +1,254 @@
+"""Muon.
+
+Implementation of the
+[Muon optimizer](https://github.com/KellerJordan/modded-nanogpt)
+by Keller Jordan
+"""
+
+
+from typing import NamedTuple, Optional, Union
+
+import chex
+import jax
+import jax.numpy as jnp
+
+from optax import tree_utils as otu
+from optax._src import alias
+from optax._src import base
+from optax._src import combine
+from optax._src import numerics
+from optax._src import transform
+from optax._src import utils
+
+
+def orthogonalize_via_newton_schulz(
+  x: jax.Array,
+  ns_coeffs: jax.Array,
+  ns_steps: int = 5,
+  eps: float = 1e-8,
+) -> jax.Array:
+  r"""Orthogonalize via Newton-Schulz iteration.
+
+  We opt to use a quintic iteration whose coefficients are selected to maximize
+  the slope at zero. For the purpose of minimizing steps, it turns out to be
+  empirically effective to keep increasing the slope at zero even beyond the
+  point where the iteration no longer converges all the way to one everywhere
+  on the interval. This iteration therefore does not produce UV^T but rather
+  something like US'V^T where S' is diagonal with S_{ii}' ~ Uniform(0.5, 1.5),
+  which turns out not to hurt model performance at all relative to UV^T, where
+  USV^T = G is the SVD.
+
+  Args:
+    x: A matrix to orthogonalize.
+    ns_coeffs: Coefficients for the Newton-schulz iterators.
+      Must have shape (n, 3) where n is the number of iterations.
+    ns_steps: Number of Newton-schulz iterations.
+      Ignored if `ns_coeffs` is a 2D array.
+    eps: Term added to denominators to improve numerical stability.
+
+  Returns:
+    The orthogonalized matrix.
+  """
+  if x.ndim != 2:
+    raise ValueError(f'Input must have shape (m, n), got {x.shape}')
+  if ns_coeffs.ndim > 2 or ns_coeffs.shape[-1] != 3:
+    raise ValueError(
+      'Newton-Schulz coefficients must have shape (3,) or (n, 3), '
+      f'got {ns_coeffs.shape}'
+    )
+  def newton_schulz_iterator(x: jax.Array, coeffs: jax.Array) -> jax.Array:
+    a = x @ x.T
+    b = coeffs[1] * a + coeffs[2] * a @ a
+    return coeffs[0] * x + b @ x
+  transposed = False
+  if x.shape[0] > x.shape[1]:
+    x = x.T
+    transposed = True
+  x /= jnp.linalg.norm(x) + eps  # Ensure spectral norm is at most 1
+  ns_coeffs = ns_coeffs.astype(x.dtype)
+  if ns_coeffs.ndim == 1:
+    x = jax.lax.fori_loop(
+      0, ns_steps, lambda _, x: newton_schulz_iterator(x, ns_coeffs), x
+    )
+  else:
+    x, _ = jax.lax.scan(
+      lambda x, abc: (newton_schulz_iterator(x, abc), None), x, ns_coeffs
+    )
+  if transposed:
+    x = x.T
+  return x
+
+
+class MuonState(NamedTuple):
+  """State for the Adam algorithm."""
+  count: chex.Array  # shape=(), dtype=jnp.int32.
+  mu: base.Updates
+
+
+def scale_by_muon(
+    ns_coeffs: Union[
+      tuple[float, float, float],
+      tuple[tuple[float, float, float], ...],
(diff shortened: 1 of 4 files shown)


leloykun:
Hi all!

How do I restrict the tests to exclude vectors as weights?

vroulet:
Hello @leloykun,

Looks like an interesting contribution thank you!

> How do I restrict the tests to exclude vectors as weights?

I don't fully understand your question. Could you give more context?

Other questions/remarks:

How does this optimizer treat vector-shaped values? The"muon_iterator" could run on vectors but not return what you want, so how do you make the distinction? Should it take a mask to only apply on matrices? Should it raise value error?

Also could you add the mathematical description in the doc string? That would greatly help.

Finally, put references at the end of the docstring (we'll adopt this format with #1129)

Thank you!

leloykun:
Hi @vroulet,

Muon is only defined for matrix-shaped values.

I'm thinking of raising an error when the input is vector-shaped, but where's the best place to put it? If there are other optimizers here that does this, can you point me to them?

vroulet:
Hello @leloykun 

- I would put a test in the init function. 
- What is the algorithm for? Most architectures have non-matrix shaped parameters (for example just the scalar parameters of a layer norm). If the algorithm is specific to some problem, could you make a quick notebook explaining the setup? (Is it for example for finetuning with LoRa?)
- In any case, some doctest would be good.
- For the tests, you may add a synthetic linear classification problem with centered data to test the algorithm for example (again if you have a specific setup let me know).

Thanks again and sorry for the delay!

vroulet:
Hey @leloykun,

Looking at the original repository, it seems that the best course of action is to:
- keep the scale_by_muon as you did
- make muon as a [partition](https://github.com/google-deepmind/optax/blob/63cdeb4ada95498626a52d209a029210ac066aa1/optax/transforms/_combining.py#L155#L267) instance where the partition is inferred from the dimensions of the leaves in the optimizer
- add in the tests a simple mlp and make all optimizers pass it

Please let me know if you are willing to pursue the PR :)

leloykun:
Hi @rdyro! Thank you so much for the help. This PR should now be ready for review. I've also already addressed @vroulet's concerns. Please let me know if there is anything else I need to do on my end.

optax/contrib/_muon.py
```diff
@@ -0,0 +1,254 @@
+"""Muon.
+
+Implementation of the
+[Muon optimizer](https://github.com/KellerJordan/modded-nanogpt)
+by Keller Jordan
+"""
+
+
+from typing import Any, NamedTuple, Optional, Tuple, Union
```
rdyro, line 9: nit: could you remove `Optional, Tuple, Union` and use python 3.9+ annotations, e.g., Union[A, B] === A | B, Optional[A] === A | None

optax/contrib/_muon.py
```diff
@@ -0,0 +1,254 @@
...
+
+class MuonState(NamedTuple):
+  """State for the Adam algorithm."""
+  count: chex.Array  # shape=(), dtype=jnp.int32.
+  mu: base.Updates
+
+
+def scale_by_muon(
+    ns_coeffs: Union[
+      Tuple[float, float, float],
```
rdyro, line 90: nit: could we change Union to the | operator and Tuple to tuple

optax/contrib/_muon.py
```diff
@@ -0,0 +1,254 @@
...
+        lambda x, y: jnp.einsum('ij,ij,ab->ab', x, y, y), mu_hat, updates
+      )
+    mu = otu.tree_cast(mu, mu_dtype)
+    return updates, MuonState(count=count_inc, mu=mu)
+  return base.GradientTransformation(init_fn, update_fn)
+
+
+def muon(
+    learning_rate: base.ScalarOrSchedule,
+    ns_coeffs: Union[
```
rdyro, line 171: nit: | and tuple

optax/contrib/_muon.py
```diff
@@ -0,0 +1,254 @@
...
+from optax._src import alias
+from optax._src import base
+from optax._src import combine
+from optax._src import numerics
+from optax._src import transform
+from optax._src import utils
+
+
+def orthogonalize_via_newton_schulz(
+  x: jnp.ndarray,
```
rdyro, line 25: nit: could we use jax.Array instead of jnp.ndarray

optax/contrib/_muon.py
```diff
@@ -0,0 +1,254 @@
...
+    The orthogonalized matrix.
+  """
+  if x.ndim != 2:
+    raise ValueError(f'Input must have shape (m, n), got {x.shape}')
+  if ns_coeffs.ndim > 2 or ns_coeffs.shape[-1] != 3:
+    raise ValueError(
+      'Newton-Schulz coefficients must have shape (3,) or (n, 3), '
+      f'got {ns_coeffs.shape}'
+    )
+  def newton_schulz_iterator(x: jnp.ndarray, coeffs: jnp.ndarray):
```
rdyro, line 59: nit: could we use jax.Array instead of jnp.ndarray

rdyro (commented):
Can you introduce these cosmetic changes? The PR looks really good!

rdyro:
> Hi @rdyro! Thank you so much for the help. This PR should now be ready for review. I've also already addressed @vroulet's concerns. Please let me know if there is anything else I need to do on my end.

Hey, I just reviewed the PR, it looks great, thanks for addressing @vroulet's comments. I left some cosmetic comments.

Once you get those in, can you squash your commits please?

rdyro:
I'm sorry, I mislead you. Because we're supporting Python 3.9 Union and Tuple are necessary, exactly like you had it (having Optional would be ok too, but you can leave it as Union[A, None]). I'll start the merge once you can revert those Union/Tuple changes.

leloykun:
@rdyro, I've just addressed your nits and squashed the commits.

I kept the `Union` & `Optional` type hints tho as support for `|` was only added in v3.10 instead of v3.9... See: https://peps.python.org/pep-0604/

rdyro (approved):

jkyl:
Thank you for this contribution, I think the way that you partitioned the matrix params from the non-matrix ones is totally reasonable. I did just want to point out an important edge case which is not handled here yet. [When using `vmap` to create a stack of layers](https://flax.readthedocs.io/en/latest/nnx_basics.html#scan-over-layers), the layer stack's weight matrices will be stacked along the zeroth axis, creating a 3-dimensional array. This kind of param won't get picked up by Muon in this implementation. It's not obvious to me how this should be handled -- maybe its best to expose the partitioning condition to the user somehow?

stephen-huan:
Also convolution filters, I think? cf. [Muon](https://github.com/KellerJordan/Muon/blob/12dc5336852ec10fbe1a1119ad09c92d2d81505e/muon.py#L97-L98)

```python
                    if g.ndim == 4: # for the case of conv filters
                        g = g.view(len(g), -1)
```

PR #1127, opened 2024-11-02
Add Nesterov momentum to AdaBelief optimizer.

carlosgmartin:
Add Nesterov momentum to the AdaBelief optimizer. This commit adds the same code that [scale_by_adam](https://github.com/google-deepmind/optax/blob/d4592a61e353128714e12a975902e4ed39b74846/optax/_src/transform.py#L310) uses to compute `mu_hat` when `nesterov=True` to [scale_by_belief](https://github.com/google-deepmind/optax/blob/d4592a61e353128714e12a975902e4ed39b74846/optax/_src/transform.py#L720).

### DIFF
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -36,7 +36,10 @@ def adabelief(
     b1: float = 0.9,
     b2: float = 0.999,
     eps: float = 1e-16,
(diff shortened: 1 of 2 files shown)


fabianp (approved):
looks good to me, thanks!

PR #1130, opened 2024-11-09
Fix AdaBelief implementation.

carlosgmartin:
As shown [here](https://arxiv.org/pdf/2010.07468#page=2) in Algorithm 2, the prediction error should use the newly-calculated $m_t$, not the old $m_{t - 1}$.

This seems to yield an improvement:


vroulet (approved):
Good catch, thank you!

PR #1138, opened 2024-11-13
Replaced seed with key in add_noise and noisy_sgd

Tomas542:
Replaced `seed` in `add_noise` with `key` in favor of `jax.random`-like style.
Added (duplicated) example with `add_noise` from `noisy_sgd`.
Changed `seed` to `key` in `noisy_sgd`. Imported `chex` for annotation purpose in `_alias.py`

### DIFF
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -28,6 +28,7 @@ from optax._src import transform
 from optax._src import wrappers
 import chex
 
+
 MaskOrFn = Optional[Union[Any, Callable[[base.Params], Any]]]
 
 
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -1319,7 +1319,7 @@ def noisy_sgd(
     Networks <https://arxiv.org/abs/1511.06807>`_, 2015
   """
   return combine.chain(
-      transform.add_noise(eta, gamma, key),
+      transform.add_noise(key, eta, gamma),
       transform.scale_by_learning_rate(learning_rate),
   )
 
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -26,7 +26,7 @@ from optax._src import factorized
 from optax._src import linesearch as _linesearch
 from optax._src import transform
 from optax._src import wrappers
-
+import chex
 
 MaskOrFn = Optional[Union[Any, Callable[[base.Params], Any]]]
 
@@ -1253,10 +1253,10 @@ def lamb(
 
 
 def noisy_sgd(
+    key: chex.PRNGKey, 
     learning_rate: base.ScalarOrSchedule,
     eta: float = 0.01,
     gamma: float = 0.55,
-    seed: int = 0,
 ) -> base.GradientTransformation:
   r"""A variant of SGD with added noise.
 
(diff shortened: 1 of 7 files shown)


rdyro:
These changes look good! Can you sign the google CLA to be able to run unit tests?

I like the change of placing the key first in the argument list. Can you take a look at these files as well:
* https://github.com/google-deepmind/optax/blob/63cdeb4ada95498626a52d209a029210ac066aa1/optax/_src/utils.py#L112
* https://github.com/google-deepmind/optax/blob/63cdeb4ada95498626a52d209a029210ac066aa1/optax/perturbations/_make_pert.py#L38
* https://github.com/google-deepmind/optax/blob/63cdeb4ada95498626a52d209a029210ac066aa1/optax/contrib/_privacy.py#L36

It'd be good to rename the argument to `key` from `seed` in all these places and move the `key` argument to the beginning of the argument list - this might require touching up a couple of tests.

Tomas542:
I signed the CLA yesterday and went to bed.
It's a great idea, but I think I'll do it this weekend. 
I also have a question (sorry if it's stupid, I've never done Pull Requests before): if I make a new commit with these changes to my branch - do they automatically go into this PR? Or is there some button I have to press?

Tomas542:
Ok, and one more question - there are a lot of things that will change. Maybe we'll add `key` as an optional argument with an `DeprecationWarning` if `seed` is set, like `Warning: argument seed will be removed (or replaced in cases where seed already PRNGKey) in the next release. Use key instead`. But in this case key will move from the first argument, and we will have to make it of type `Optional` for now. And in the next-next major release I will remove/replace seed with key. 
Or is it better to change everything now? What do you think?

vroulet (commented):
Just a small comment about the signature of the optimizer. Otherwise I fully agree with Robert that a full pass would be great. Could be in a separate PR if needed.

Tomas542:
Hi, haven't had time to make changes.
I agree with the positions of the arguments.
But I don't agree that the key should be set to the default value for two reasons:
- At this point, this parameter does not have a default value yet. So not having a value will not “slow down” optimizer testing.
- JAX (as far as I know) never sets default values for keys, this is more specific to libraries such as datasets.
So perhaps “don't set a default value” is a better option?

rdyro:
A default key value might be confusing giving determinism where people don't expect it, but `noisy_sgd` is a first-order optimizer and argument order consistency is definitely a value. Maybe we can keep key last, but **required**.

Something like:
```python
def noisy_sgd(
    learning_rate: base.ScalarOrSchedule,
    eta: float = 0.01,
    gamma: float = 0.55,
    key: chex.PRNGKey | None = None,
) -> base.GradientTransformation:
  if key is None:
    raise ValueError("noisy_sgd optimizer requires specifying random key: noisy_sgd(..., key=random.key(0))")
```

@Tomas542 @vroulet what do you think?

vroulet:
I like this idea. It may also smooth out backward compatibility issues with clear raised errors.

rdyro:
> A default key value might be confusing giving determinism where people don't expect it, but `noisy_sgd` is a first-order optimizer and argument order consistency is definitely a value. Maybe we can keep key last, but **required**.
> 
> Something like:
> 
> ```python
> def noisy_sgd(
>     learning_rate: base.ScalarOrSchedule,
>     eta: float = 0.01,
>     gamma: float = 0.55,
>     key: chex.PRNGKey | None = None,
> ) -> base.GradientTransformation:
>   if key is None:
>     raise ValueError("noisy_sgd optimizer requires specifying random key: noisy_sgd(..., key=random.key(0))")
> ```
> 
> @Tomas542 @vroulet what do you think?

@Tomas542 could you add this change?

Tomas542:
Yeap, I can do this. Also I will try to change tests.

To summarize, we decided to replace all `seed` with `key`, make `key` the default value `None` and raise an error if it is `None`.  As for the order of function arguments, we put `key` in the last position for optimizers and in the first position in some other functions, such as `add_noise`?

vroulet:
Yes, good point for the annotation. And yes for the summary!

Tomas542:
Now we can move to #1145

PR #1140, opened 2024-11-20
Add improved version of Hungarian algorithm.

carlosgmartin:
Follow-up to https://github.com/google-deepmind/optax/pull/1083#issuecomment-2487415673.

### DIFF
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -2481,12 +2481,12 @@ def lbfgs(
     ...      grad, opt_state, params, value=value, grad=grad, value_fn=f
     ...   )
     ...   params = optax.apply_updates(params, updates)
-    ...   print('Objective function: ', f(params))
-    Objective function:  7.5166864
-    Objective function:  7.460699e-14
-    Objective function:  2.6505726e-28
-    Objective function:  0.0
-    Objective function:  0.0
+    ...   print('Objective function: {:.2E}'.format(f(params)))
+    Objective function: 7.52E+00
+    Objective function: 7.46E-14
+    Objective function: 2.65E-28
+    Objective function: 0.00E+00
+    Objective function: 0.00E+00
 
   References:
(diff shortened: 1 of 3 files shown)


vroulet (commented):
Thanks again @carlosgmartin!
Some suggestions:
- keep a single file, and name the first algorithm you did as "base_hungarian_algorithm"
- add in e.g. the new algorithm's docstring a sentence that differentiates the two implementations.
- Add a test that compares the results between the two implementations maybe

Otherwise it's great

carlosgmartin:
@vroulet Seems to me it’s cleaner to keep each implementation in a separate file since

1. It keeps their respective helper functions separate and prevents confusion.
2. More implementations might be added in the future, and this results in less crowded files.

WDYT?

vroulet:
> It keeps the implementations less entangled and keeps their respective helper functions separate, avoiding potential confusion.

In any case we should add in the docstring why they differ.

> It keeps the modules shorter/less crowded, especially if more implementations are added in the future.

I would actually agree with you, but typically the alias file in optax has an ever increasing list of optimizers. So let's keep the same logic here.

carlosgmartin:
@vroulet Done.

vroulet (approved):
Thanks again @carlosgmartin !

optax/_src/alias.py
```diff
@@ -2482,7 +2482,7 @@ def lbfgs(
     ...   )
     ...   params = optax.apply_updates(params, updates)
     ...   print('Objective function: ', f(params))
-    Objective function:  7.5166864
+    Objective function:  7.516686...
```
fabianp, line 2485: A better way to consider only a few decimals is to print only those decimals. See for example the snippet `print('Objective function: {:.2E}'.format(f(params)))` used by adam

carlosgmartin: Updated.

Issue #1141, opened 2024-11-21
L-BFGS optimizer with complex inputs


gautierronan:
I am adapting the docstring example of `optax.lbfgs()` to use complex-valued parameters instead of real-valued ones. The output function to optimize is still real-valued. But this raises a `TypeError: true_fun and false_fun output must have identical types` when calling `solver.update`, along with a warning `ComplexWarning: Casting complex values to real discards the imaginary part`. I am guessing that this warning is the source of the TypeError.

MWE:
```python
import optax
import jax.numpy as jnp

def f(x):
    return jnp.sum(jnp.abs(x**2))

solver = optax.lbfgs()
params = jnp.array([1.0 + 1j, 2.0 + 2j, 3.0 + 3j])
print("Objective function: ", f(params))

opt_state = solver.init(params)
value_and_grad = optax.value_and_grad_from_state(f)

for _ in range(5):
    value, grad = value_and_grad(params, state=opt_state)
    updates, opt_state = solver.update(
        grad, opt_state, params, value=value, grad=jnp.conj(grad), value_fn=f
    )
    params = optax.apply_updates(params, updates)
    print("Objective function: ", f(params))
```

Stack trace:
```python
Objective function:  28.0
[/python3.11/site-packages/jax/_src/ops/scatter.py:92](/python3.11/site-packages/jax/_src/ops/scatter.py:92): FutureWarning: scatter inputs have incompatible types: cannot safely cast value from dtype=complex64 to dtype=float32 with jax_numpy_dtype_promotion='standard'. In future JAX releases this will result in an error.
  warnings.warn(
[/python3.11/site-packages/jax/_src/ops/scatter.py:134](/python3.11/site-packages/jax/_src/ops/scatter.py:134): ComplexWarning: Casting complex values to real discards the imaginary part
  return lax_internal._convert_element_type(out, dtype, weak_type)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[1], line 16
     14 for _ in range(5):
     15     value, grad = value_and_grad(params, state=opt_state)
---> 16     updates, opt_state = solver.update(
     17         grad, opt_state, params, value=value, grad=jnp.conj(grad), value_fn=f
     18     )
     19     params = optax.apply_updates(params, updates)
     20     print("Objective function: ", f(params))

[...]

PR #1142, opened 2024-11-22
Fix complex support for L-BFGS

gautierronan:
Closes https://github.com/google-deepmind/optax/issues/1141.

Not 100% sure that this doesn't break other things or fully works, but at least the MWE below seems to work fine.
```python
import optax
import jax.numpy as jnp

def f(x):
    return jnp.sum(jnp.abs(x**2))

solver = optax.lbfgs()
params = jnp.array([1.0 + 1.0j, 2.0 + 2.0j, 3.0 + 3.0j])
print("Objective function: ", f(params))

opt_state = solver.init(params)
value_and_grad = optax.value_and_grad_from_state(f)

for _ in range(5):
    value, grad = value_and_grad(params, state=opt_state)
    updates, opt_state = solver.update(
        jnp.conj(grad), opt_state, params, value=value, grad=jnp.conj(grad), value_fn=f
    )
    print("Objective function: ", f(params))
```

Notice the `solve.update` call which requires a `jnp.conj(grad)` twice. I believe this is correct and aligned with other optax solvers, but not sure either.

### DIFF
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -2514,6 +2514,8 @@ def lbfgs(
     constrain the trust-region of the first step to an Euclidean ball of radius
     1 at the first iteration. The choice of :math:`\gamma_0` is not detailed in
     the references above, so this is a heuristic choice.
+
+  .. note:: The algorithm can support complex inputs.
   """
   if learning_rate is None:
     base_scaling = transform.scale(-1.0)
--- a/optax/_src/alias_test.py
+++ b/optax/_src/alias_test.py
@@ -869,14 +869,13 @@ class LBFGSTest(chex.TestCase):
 
   @parameterized.product(
     linesearch=[
-      # _linesearch.zoom_linesearch(max_linesearch_steps=20),
       _linesearch.scale_by_backtracking_linesearch(max_backtracking_steps=20),
       _linesearch.scale_by_zoom_linesearch(
         max_linesearch_steps=20, initial_guess_strategy='one'
       )
     ],
   )
-  def test_complex(self, linesearch):
+  def test_lbfgs_complex(self, linesearch):
     # Test that optimization over complex variable matches equivalent real case
 
     tol = 1e-5
--- a/optax/_src/alias_test.py
+++ b/optax/_src/alias_test.py
@@ -910,5 +910,33 @@ class LBFGSTest(chex.TestCase):
         sol_complex, to_complex(sol_real), atol=tol, rtol=tol
     )
 
+  @parameterized.product(
+    linesearch=[
+      _linesearch.scale_by_backtracking_linesearch(max_backtracking_steps=20),
+      _linesearch.scale_by_zoom_linesearch(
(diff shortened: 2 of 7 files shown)


vroulet:
Hey @gautierronan,
Thanks for the PR! We'll need a test. Take look at this PR: https://github.com/google/jaxopt/pull/468 that added support for complex parameters for the lbfgs of jaxopt. I think you'll find all that you'll need in that PR.
Thanks again!

gautierronan:
@vroulet Should be good for review.

vroulet (commented):
That looks great, thank you @gautierronan ! I left you some cosmetic comments.
In addition, can you add the following note in the docstrings of `lbfgs`, `scale_by_zoom_linesearch`, `scale_by_backtracking_linesearch` (please compile the docs to verify that it's well formatted)?
```
  ..note:: The algorithm can support complex inputs.
```
Ideally, one would add a simple example in the lbfgs notebook to link this note to an implementation (as you pointed out the gradients need to pass through a conjugate). Let me know if you could add that too.

Also could you add the test on the complex rosenbrock that was in [jaxopt](https://github.com/google/jaxopt/blob/9552f3923d487d2edca3a2c95196b9e77d84dadc/tests/lbfgs_test.py#L571C1-L589C44). It should amount to add exactly
```
  def test_lbfgs_complex_rosenbrock(self):
    # Taken from previous jax tests
    tol = 1e-5
    complex_dim = 5

    fun_real = _get_problem('rosenbrock')['fun']
    init_real = jnp.zeros((2 * complex_dim,), dtype=complex)
    expected_real = jnp.ones((2 * complex_dim,), dtype=complex)

    def fun(z):
      x_real = jnp.concatenate([jnp.real(z), jnp.imag(z)])
      return fun_real(x_real)

    init = init_real[:complex_dim] + 1.j * init_real[complex_dim:]
    expected = expected_real[:complex_dim] + 1.j * expected_real[complex_dim:]

    opt = alias.lbfgs()
    got, _ = _run_opt(opt, fun, init, maxiter=500, tol=tol)
    chex.assert_trees_all_close(got, expected)
 ```
If it does not pass, let me know. I can work on that in a separate PR once yours is merged.

gautierronan:
Thanks for the review. All comments should be addressed in the last commits. Let me know if that works.

> Ideally, one would add a simple example in the lbfgs notebook to link this note to an implementation (as you pointed out the gradients need to pass through a conjugate). Let me know if you could add that too.

I don't think I will have time for this. Also, it's IMO expected that gradients need to pass through a conjugate (nothing specific to L-BFGS), and is already the behavior in the other optax functions I believe. So a tutorial dedicated to complex-valued gradient descent might be more relevant than a specific addition to the L-BFGS tutorial?

vroulet:
> So a tutorial dedicated to complex-valued gradient descent might be more relevant than a specific addition to the L-BFGS tutorial?

Good point. We can make an issue for that. Thanks for all!

vroulet (approved):

PR #1145, opened 2024-11-30
Replace `seed` with `key`

Tomas542:
Replace `seed` with key in all files. Add `key=None` error for 2 optimizers

UPD: Forgot, added example in `add_noise` docstring

### DIFF
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -1321,7 +1321,7 @@ def noisy_sgd(
   """
   if key is None:
     raise ValueError(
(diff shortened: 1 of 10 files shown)


Tomas542:
It seems, I've failed in some tests (maybe in rewritten there is some error), need time to set up Linux VM to run bash scripts...

Tomas542:
messed with commit text... removed `random` word from error

Tomas542:
this test will fail due to assertions about tree not equal in key values. I will fix it later. Now I just reducing errors in logs

PR #1154, opened 2024-12-09
Add linear program solver based on the restarted Halpern primal-dual hybrid gradient (rHPDHG) algorithm.

carlosgmartin:
Reference:

- Haihao Lu, Jinwen Yang. [Restarted Halpern PDHG for Linear Programming](https://arxiv.org/abs/2407.16144). 2024.

See also https://github.com/MIT-Lu-Lab/MPAX.


vroulet:
Thank you again @carlosgmartin
Why not adding this algorithm to mpax?
Optax is more oriented deep learning, you would have much less visibility here than in mpax I think.

Issue #1152, opened 2024-12-08
Add `nnls` (non-negative least squares)


carlosgmartin:
**Feature request:** Add a function called `nnls` that solves a [non-negative least squares problem](https://en.wikipedia.org/wiki/Non-negative_least_squares). For an example, see SciPy's [`scipy.optimize.nnls`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.nnls.html).

I've opened a PR here: https://github.com/google-deepmind/optax/pull/1155.

PR #1155, opened 2024-12-09
Add non-negative least squares (NNLS) solver.

carlosgmartin:
Fixes https://github.com/google-deepmind/optax/issues/1152.

### DIFF
--- a/optax/_src/linear_algebra.py
+++ b/optax/_src/linear_algebra.py
@@ -265,3 +265,79 @@ def matrix_inverse_pth_root(
     resultant_mat_h = is_converged * mat_h + (1 - is_converged) * old_mat_h
     resultant_mat_h = jnp.asarray(resultant_mat_h, matrix.dtype)
   return resultant_mat_h, error
+
+
+def get_spectral_radius_upper_bound(matrix):
+  # Get an upper bound on the spectral radius of a matrix.
+  a = jnp.linalg.matrix_norm(matrix, ord="fro")
+  if matrix.size == 0:
+    b = 0  # TODO: remove this branch after jax/issues/26555 is fixed
+  else:
+    b = jnp.linalg.matrix_norm(matrix, ord=1)
+  return jnp.minimum(a, b)
+
+
+def nnls(
+    A: jax.Array,
+    b: jax.Array,
+    iters: int,
+    unroll: Union[int, bool] = 1,
+    L: Union[jax.Array, float, None] = None,
+) -> jax.Array:
+  r"""Solves the non-negative least squares problem.
+
+  Minimizes :math:`\|A x - b\|_2` subject to :math:`x \geq 0`.
+
+  Uses the fast projected gradient (FPG) algorithm of Polyak 2015.
+
+  Args:
+    A: Input matrix.
+    b: Input vector.
+    iters: Number of iterations to run the algorithm for.
+    unroll: Unroll parameter passed to `lax.scan`.
+    L: An upper bound on the spectral radius of `A.T @ A` (optional).
+
+  Returns:
+    The solution vector.
+
+  Examples:
+    >>> from jax import numpy as jnp
+    >>> import optax
+    >>> A = jnp.array([[1., 2.], [3., 4.]])
+    >>> b = jnp.array([5., 6.])
+    >>> x = optax.nnls(A, b, 10**3)
+    >>> print(f"{x[0]:.2f}")
+    0.00
+    >>> print(f"{x[1]:.2f}")
+    1.70
+
+  References:
+    Roman A. Polyak, `Projected gradient method for non-negative least square
+    <http://www.ams.org/books/conm/636/>`_, 2015
+  """
+  Q = A.T @ A
+  q = A.T @ b
+
+  if L is None:
+    L = get_spectral_radius_upper_bound(Q)
+
+  L = jnp.where(L == 0, 1, L)  # avoid division by zero below
+
+  def f(x_p_c, _):
+    x, p, c = x_p_c
+
+    cn = (1 + jnp.sqrt(1 + 4 * c ** 2)) / 2
+    s = (c - 1) / cn
+
+    xn = (p - (Q @ p - q) / L).clip(0)
+    pn = xn + s * (xn - x)
+
+    return (xn, pn, cn), None
+
+  x = jnp.zeros(A.shape[1])
+  p = x
+  c = 0.
+
+  (x, p, c), _ = lax.scan(f, (x, p, c), length=iters, unroll=unroll)
+
+  return x
--- a/docs/api/utilities.rst
+++ b/docs/api/utilities.rst
@@ -50,6 +50,7 @@ Linear Algebra Operators
 .. autosummary::
     matrix_inverse_pth_root
     power_iteration
+    nnls
 
 Matrix inverse pth root
 ~~~~~~~~~~~~~~~~~~~~~~~~
@@ -58,6 +59,10 @@ Matrix inverse pth root
 Power iteration
 ~~~~~~~~~~~~~~~
 .. autofunction:: power_iteration
+
+Non-negative least squares
+~~~~~~~~~~~~~~~
+.. autofunction:: nnls
 
 
 Second Order Optimization
--- a/optax/__init__.py
+++ b/optax/__init__.py
@@ -80,6 +80,7 @@ from optax._src.factorized import scale_by_factored_rms
 from optax._src.linear_algebra import global_norm
 from optax._src.linear_algebra import matrix_inverse_pth_root
 from optax._src.linear_algebra import power_iteration
+from optax._src.linear_algebra import nnls
 from optax._src.linesearch import scale_by_backtracking_linesearch
 from optax._src.linesearch import scale_by_zoom_linesearch
 from optax._src.linesearch import ScaleByBacktrackingLinesearchState
@@ -378,6 +379,7 @@ __all__ = (
     "MultiTransformState",
     "nadam",
     "nadamw",
+    "nnls",
     "noisy_sgd",
     "novograd",
     "NonNegativeParamsState",
--- a/optax/_src/linear_algebra_test.py
+++ b/optax/_src/linear_algebra_test.py
@@ -209,6 +209,39 @@ class LinearAlgebraTest(chex.TestCase):
         # No guarantee of success after e >= 7
         pass
 
+  @parameterized.parameters(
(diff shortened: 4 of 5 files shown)


carlosgmartin:
@rdyro How does this look?

rdyro:
@carlosgmartin This looks great, I left one comment!

carlosgmartin:
@rdyro Where is this comment?

rdyro:
> @rdyro Where is this comment?

Oops, should be up now!

optax/_src/linear_algebra.py
```diff
@@ -265,3 +265,100 @@ def _iter_body(state):
...
+      s = jnp.linalg.lstsq(AtA * p[:, None] * p[None, :], Atb * p)[0]
+
+      return x, p, s, it
+
+    x, p, w, it = carry
+
+    j = masked_argmax(w, ~p)
+    p = p.at[j].set(True)
+
+    # s = jnp.linalg.lstsq(A * p, b)[0]
```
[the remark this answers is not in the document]
carlosgmartin: Done.

rdyro:
>     # We use lstsq with a pre-computed AtA to reduce computation time.
>     s = jnp.linalg.lstsq(AtA * p[:, None] * p[None, :], Atb * p)[0]

If speed is a consideration, perhaps we should use the `jnp.linlag.lsqt(A * p, b)` directly, letting XLA optimize as necessary?

I'd be curious if we could introduce a once-factorized cholesky version of this algorithm where we repeatedly apply cho_solve to a masked L or U factor of A^T A?

@carlosgmartin @fabianp ?

carlosgmartin:
@rdyro Edited.

optax/_src/linear_algebra.py
```diff
@@ -265,3 +265,98 @@ def _iter_body(state):
...
+
+    return x, p, w, it
+
+  _, n = A.shape
+
+  if n == 0:
+    return jnp.zeros(0)
+
+  Atb = b @ A
+  AtA = A.T @ A
```
rdyro, line 353: For readability can we move closure variables (`Atb, AtA`) above the function enclosing them?

carlosgmartin: @rdyro Done.

rdyro (approved):
Thank you for this PR!

rdyro:
I ran the tests internally and I can't pass the tolerance checks for this algorithm for larger dimensions (e.g., `n=24 d=64`). Could you take a look?

optax/_src/linear_algebra.py
```diff
@@ -265,3 +265,98 @@ def _iter_body(state):
...
+      return x, p, s, it
+
+    x, p, w, it = carry
+
+    j = masked_argmax(w, ~p)
+    p = p.at[j].set(True)
+
+    s = jnp.linalg.lstsq(A * p, b)[0]
+
+    x, p, s, it = lax.while_loop(in_cond_fn, in_body_fn, (x, p, s, it))
```
fabianp, line 343: x is overwritten in the next line. is this potentially a bug?

carlosgmartin: Don't think so. That seems to match the [Lawson-Hanson algorithm](https://en.wikipedia.org/wiki/Non-negative_least_squares#Algorithms).

carlosgmartin:
Possibly related: [`jax.numpy.linalg.lstsq`](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.linalg.lstsq.html) seems to be [buggy](https://github.com/jax-ml/jax/issues/26468).

rdyro:
NaNs propagate so we'd see NaNs in the output, but the algorithm seems to not be converging instead.

carlosgmartin:
True. It might be that the operations of [the algorithm](https://en.wikipedia.org/wiki/Non-negative_least_squares#Algorithms) accumulate numerical error, yielding incorrect outcomes when the inequality comparison operators (which are discontinuous) are applied. I'm taking a closer look.

carlosgmartin:
Due to the above difficulties, I switched to the fast projected gradient (FPG) algorithm of [Polyak 2015](http://www.ams.org/books/conm/636/).

This has some advantages that are described in the [that paper's introduction](https://mason.gmu.edu/~rpolyak/Publications/ProjectGradientMethod_least_square.pdf), including not having to solve a linear system on each iteration (instead, just doing matrix-vector multiplications).

rdyro:
Nice! Maybe we can explore also adding an ADMM version in a future CL [https://stanford.edu/class/ee364b/lectures/admm_slides.pdf](https://stanford.edu/class/ee364b/lectures/admm_slides.pdf) since it's also promising a $\mathcal{O}(1/k)$ convergence (requires resolves, but with only a single factorization at the beginning).

optax/_src/linear_algebra.py
```diff
@@ -265,3 +265,72 @@ def _iter_body(state):
...
+  References:
+    Roman A. Polyak, `Projected gradient method for non-negative least square
+    <http://www.ams.org/books/conm/636/>`_, 2015
+  """
+  if A.shape[0] == 0 or A.shape[1] == 0:
+    return jnp.zeros(A.shape[1])
+
+  Q = A.T @ A
+  q = A.T @ b
+  L = jnp.linalg.matrix_norm(Q, ord=2)  # upper bound on spectral radius of Q
```
rdyro, line 317: This is a very expensive operation, can we upper bound it with the Frobenius norm? [https://www.math.uwaterloo.ca/~hwolkowi/matrixcookbook.pdf](https://www.math.uwaterloo.ca/~hwolkowi/matrixcookbook.pdf) - seems to imply $||A||_2 \leq ||A||_F$

carlosgmartin: What about using [power iteration](https://optax.readthedocs.io/en/latest/api/utilities.html#optax.power_iteration) to get an upper bound on the spectral radius, as Polyak's paper suggests at the bottom of [page 2](https://mason.gmu.edu/~rpolyak/Publications/ProjectGradientMethod_least_square.pdf#page=2)?

rdyro: [power_iteration](https://optax.readthedocs.io/en/latest/api/utilities.html#optax.power_iteration) sounds great!

carlosgmartin:
@rdyro Done.

optax/_src/linear_algebra.py
```diff
@@ -265,3 +265,78 @@ def _iter_body(state):
...
+  """
+  if A.shape[0] == 0 or A.shape[1] == 0:
+    return jnp.zeros(A.shape[1])
+
+  Q = A.T @ A
+  q = A.T @ b
+
+  if L is None:
+    # Compute an upper bound on the spectral radius of Q.
+    L, _ = power_iteration(Q.astype(float))
```
rdyro, line 323: Default settings with power_iteration are aimed at a high accuracy computation (`error_tolerance = 1e-6`), but here we just need an approximation, perhaps something like `1.5 * power_iteration(..., num_iters=10)` would work better? (where 1.5 is a correction factor)

rdyro: I'm curious why the cast to `float`?

carlosgmartin: Without the cast, I get the following error:
```
venv $ py -c "from jax import numpy as jnp; import optax; A = jnp.array([[1, 2], [3, 4]]); b = jnp.array([5, 6]); optax.nnls(A, b, 10**3)"
  File "<string>", line 1, in <module>
  File "/Users/carlos/Desktop/optax/optax/_src/linear_algebra.py", line 323, in nnls
    L, _ = power_iteration(Q)
           ^^^^^^^^^^^^^^^^^^
  File "/Users/carlos/Desktop/optax/optax/_src/linear_algebra.py", line 110, in power_iteration
    v0 = jax.random.uniform(
         ^^^^^^^^^^^^^^^^^^^
  File "/Users/carlos/venv/lib/python3.12/site-packages/jax/_src/random.py", line 402, in uniform
    raise ValueError(f"dtype argument to `uniform` must be a float dtype, "
ValueError: dtype argument to `uniform` must be a float dtype, got int32
```

carlosgmartin: How did you get the approximate correction factor of 1.5?

rdyro: `.astype(float)` will default to the highest float precision available, we probably want to leave the dtype the same as the one passed by the user. I'd recommend updating the test case to construct `float32` matrices instead of `int32`

carlosgmartin: I removed the cast.

rdyro: > How did you get the approximate correction factor of 1.5?

It's just a guess, Polyak needs an over-approximation of the largest eigenvalue, so 10 power-iterations plus an overcorrection should do the trick without affecting convergence too much.

Another cheap approximation is just the Frobenius norm of Q.

carlosgmartin: You're right, $`\rho(A) \leq \|A\|_\text{F}`$. Apparently it's also true that $`\rho(A) \leq \|A\|`$ for any consistent norm $`\|\cdot\|`$. This includes $`\|\cdot\|_1`$ and $`\|\cdot\|_\infty`$, which are [easily computable](https://en.wikipedia.org/wiki/Matrix_norm#p_=_1_or_%E2%88%9E), as well as $`\|\cdot\|_{2,\infty}`$ and $`\|\cdot\|_{1,2}`$, which are also [easily computable](https://en.wikipedia.org/wiki/Matrix_norm#Matrix_norms_induced_by_vector_%CE%B1-_and_%CE%B2-norms). I have to investigate which of these 5 bounds is tightest.

carlosgmartin: Empirically, it seems that the matrix 1-norm is tighter than the Frobenius norm. A few examples:
```
F-norm: 133566
1-norm: 360.021
2-norm: 138.76

F-norm: 7424.79
1-norm: 66.2015
2-norm: 49.7296

F-norm: 11.6024
1-norm: 3.40622
2-norm: 3.40622

F-norm: 184.197
1-norm: 13.0405
2-norm: 12.835

F-norm: 129.118
1-norm: 10.9546
2-norm: 9.4542

F-norm: 582.806
1-norm: 21.3551
2-norm: 18.3858

F-norm: 496.488
1-norm: 23.5996
2-norm: 18.6503

F-norm: 718.635
1-norm: 27.821
2-norm: 21.9953
```

Should I use the matrix 1-norm as the upper bound?

carlosgmartin:
@rdyro Made some minor changes to handle zero matrices without producing nans.

PR #1156, opened 2024-12-10
Split github workflows for lower latency, add ruff

rdyro:
This PR 
* splits linting and doctests from the combined `test.sh`  to provide a faster workflow iteration
  - without removing that action from `test.sh`
  - in the future this duplication can be removed
* adds `ruff check .` for fast linting check (currently in addition to pylint)
* `build-and-test` and `doctests` workflows now also depend on the linting success


fabianp (approved):
thanks Robert!

PR #1165, opened 2025-01-02
Support tuple of axis in `softmax_cross_entropy_with_integer_labels`

### LINKED ISSUE #1162: Axis tuple is not handled properly in `softmax_cross_entropy_with_integer_labels` loss
According to the docstring of `softmax_cross_entropy_with_integer_labels`, `axis` admits `None`, `int`, or `tuple[int, ...]`. However, snippet below demonstrates that `axis` of `tuple[int, int]` causes an exception.

```python
xs = jnp.ones((1, 2, 3, 4))
ys = jnp.zeros(xs.shape[:-2], dtype=jnp.int32)
mask = jnp.ones_like(xs, dtype=jnp.bool)
optax.softmax_cross_entropy_with_integer_labels(xs, ys, (-2, -1), mask)  # FAIL
# TypeError: 'tuple' object cannot be interpreted as an integer
```

The reason is that implementation exploits `take_along_axis` which admits scalar `axis` but not `tuple`.

https://github.com/google-deepmind/optax/blob/1e08bccf195ac54e7d9d766eb5e69345bf0e3230/optax/losses/_classification.py#L335-L343

There are two options from my perspective.

1. Parameter `axis` of `softmax_cross_entropy_with_integer_labels` should be either `None` or `int`.
2. Fix a bug and implement "vector" `axis` parameter.
   However, semantic of the `labels` becomes unclear.
   Specifically, if `axis` is `(-2, -1)` as in example above than what shape of `labels` and how to specify a label of 2-dimensional slice of `logits`? Obvious solution is append another dimension to `labels` with `len(axis)` elements (i.e. `(1, 2, 2)` in example above). Another solution is to assume that elements `labels` are flat indices in 2-dimensional slices of `logits` in this example.

daskol:
Fix #1162 (see #1164 for concurrent fix).

### DIFF
--- a/optax/losses/_classification.py
+++ b/optax/losses/_classification.py
@@ -15,18 +15,31 @@
 """Classification losses."""
 
 import functools
+import operator
 from typing import Optional, Union
 
 import chex
 import jax
 import jax.numpy as jnp
-import numpy as np
 from optax import projections
 
-if np.__version__.startswith('1.'):
-  from numpy.core.numeric import normalize_axis_index, normalize_axis_tuple
-else:
-  from numpy.lib.array_utils import normalize_axis_index, normalize_axis_tuple
+
+def canonicalize_axis(axis, ndim):
+  """Vendored version of :func:`numpy.lib.array_utils.normalize_axis_index`.
+  """
+  if 0 <= (axis := operator.index(axis)) < ndim:
+    return axis
+  elif -ndim <= axis < 0:
+    return axis + ndim
+  else:
+    raise ValueError(f'axis {axis} is out of bounds for array of '
+                     f'dimension {ndim}')
+
+
+def canonicalize_axes(axes, ndim) -> tuple[int, ...]:
+  """Vendored version of :func:`numpy.lib.array_utils.normalize_axis_tuple`.
+  """
+  return tuple(canonicalize_axis(x, ndim) for x in axes)
 
 
 def sigmoid_binary_cross_entropy(
@@ -354,11 +367,11 @@ def softmax_cross_entropy_with_integer_labels(
   chex.assert_type([logits], float)
   chex.assert_type([labels], int)
   if isinstance(axis, int):
-    axis = normalize_axis_index(axis, logits.ndim)
+    axis = canonicalize_axis(axis, logits.ndim)
   elif isinstance(axis, tuple):
     # Move all "feature" dimensions to the end preserving axis ordering and
     # subsequent flattening "feature" dimensions to a single one.
-    logit_axis = normalize_axis_tuple(axis, logits.ndim, argname='logits')
+    logit_axis = canonicalize_axes(axis, logits.ndim)
     batch_axis = tuple(x for x in range(logits.ndim) if x not in logit_axis)
     axis = len(batch_axis)
     logits = logits.transpose(batch_axis + logit_axis)
--- a/optax/losses/_classification.py
+++ b/optax/losses/_classification.py
@@ -20,7 +20,13 @@ from typing import Optional, Union
 import chex
 import jax
 import jax.numpy as jnp
+import numpy as np
 from optax import projections
+
+if np.__version__.startswith('1.'):
+  from numpy.core.numeric import normalize_axis_index, normalize_axis_tuple
+else:
+  from numpy.lib.array_utils import normalize_axis_index, normalize_axis_tuple
 
 
 def sigmoid_binary_cross_entropy(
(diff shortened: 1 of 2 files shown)


optax/losses/_classification.py
```diff
@@ -20,8 +20,14 @@
 import chex
 import jax
 import jax.numpy as jnp
+import numpy as np
 from optax import projections
 
+if np.__version__.startswith('1.'):
+  from numpy.core.numeric import normalize_axis_index, normalize_axis_tuple
```
rdyro, line 27: could we vendor a simplified version of `normalize_axis_index` so as to not depend on numpy here?

optax/losses/_classification.py
```diff
@@ -329,6 +353,22 @@ def softmax_cross_entropy_with_integer_labels(
   """
   chex.assert_type([logits], float)
   chex.assert_type([labels], int)
+  if isinstance(axis, int):
```
rdyro, line 356: can we unify these cases by converting axis to tuple if it's an int first?

optax/losses/_classification_test.py
```diff
@@ -247,6 +247,40 @@ def test_axis(self, shape, axis):
     )
     np.testing.assert_allclose(x, y, atol=1e-4)
 
+  @parameterized.parameters(
+      {'axis': (1, 3), 'shape': (2, 3, 4, 5)},
+      {'axis': (3, 2), 'shape': (2, 3, 4, 5)},
```
rdyro, line 252: can we have tests with more than 2 logit axis?

rdyro:
Hey, thanks! These PR looks great, I left some comments!

rdyro:
JAX has an example of [canonicalize_axis](https://github.com/jax-ml/jax/blob/0f4677bcf68f92364d7e8bf9299ee659e2745c20/jax/_src/util.py#L385), we could vendor it from there (but it's a private API, so let's not import it from JAX directly)

optax/losses/_classification.py
daskol: Yes, we definitely could but then we would get a regression in performance of greedy and jit-compiled execution. Resulting resulting `jaxpr`s are almost identical for `axis=-1` and `axis=(-1, )`). The only difference is a transposition of "feature" axis/axes to the end.

```python
xs = jnp.ones((2, 3, 4, 5))
ys = jnp.ones((2, 3, 5), jnp.int32)
fn = optax.optax.softmax_cross_entropy_with_integer_labels
jaxpr_of = jax.make_jaxpr(fn, static_argnums=2)

jaxpr_int = jaxpr_of(xs, ys, -2)
# { lambda ; a:f32[2,3,4,5] b:i32[2,3,5]. let
#     c:f32[2,3,5] = reduce_max[axes=(2,)] a
#     ... }

jaxpr_tuple = jaxpr_of(xs, ys, (-2, ))
# { lambda ; a:f32[2,3,4,5] b:i32[2,3,5]. let
#     c:f32[2,3,5,4] = transpose[permutation=(0, 1, 3, 2)] a
#     d:f32[2,3,5] = reduce_max[axes=(3,)] c
#     ... }
```

Maybe we should do the opposite and covert 1-tuples to plain `int`?

daskol:
Yeah, I know of `canonicalize_axis` and decided to get `normalize_axis_*` from origin for multiple reasons. Firstly, `normilize_axis_*` are proper implementation. Secondly, it basically consists of list/tuple transformation and jit-compilable by design. Thirdly, NumPy is already among dependencies and there is an explicit notice about deprecation in NumPy 2. But vendoring these routines is nice too.

optax/losses/_classification.py
rdyro: That's a fair point about greedy, let's keep your current version

rdyro (approved):

rdyro:
> Yeah, I know of `canonicalize_axis` and decided to get `normalize_axis_*` from origin for multiple reasons. Firstly, `normilize_axis_*` are proper implementation. Secondly, it basically consists of list/tuple transformation and jit-compilable by design. Thirdly, NumPy is already among dependencies and there is an explicit notice about deprecation in NumPy 2. But vendoring these routines is nice too.

Makes sense, numpy's canonicalize are nice, but the import discrepancy between versions is a bit of a downside given potential API instability. Thanks for vendoring them!

daskol:
Should I go ahead and close this pull request? I noticed a merge conflict and rebased the branch on top of main. However, it seems that the concurrent fix #1164 has already been merged, making this PR redundant.

daskol:
Great! Thank you for your time.

PR #1169, opened 2025-01-09
removing `scalar_type_of` - random.key breaks it

rdyro:


fabianp (approved):
looks good to me, would it be possible to add a test?

PR #1170, opened 2025-01-09
Fix Muon implementation

leloykun:
Apply orthogonalization on momentum term instead of the original updates.

This should now match the original implementation by @KellerJordan

---

Postmortem: There was an operation before this line that takes in `mu_hat` and outputs `updates`. Thus, this line made sense then. I have since removed that operation. And thus, this line should now take in `mu_hat` instead. My fault for not double-checking.

What we should do so this never happens again: we should add harder toy benchmarks so mistakes like this would be spotted more quickly.

Tagging @rdyro for review.


rdyro:
No worries, thanks for the quick fix!

rdyro (approved):

PR #1173, opened 2025-01-13
CI: test against nightly JAX

jakevdp:
This will catch potential incompatibilities introduced in JAX before they make it into a release.

Also update parts of the workflow definition to reflect current best practices.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,13 +8,21 @@ on:
   schedule:
     - cron: '0 3 * * *'
 
+permissions:
+  contents: read  # to fetch code
+  actions: write  # to cancel previous workflows
+
+concurrency:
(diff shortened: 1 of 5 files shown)


jakevdp:
Test failures point to code added earlier today in https://github.com/google-deepmind/optax/pull/1169

jakevdp:
Looks like #1172 should fix the failures; I'll rebase on that once it's merged.

jakevdp:
Chatting with @rdyro it seems the test failure is probably related to JAX enabling partitionable threefry by default (https://github.com/jax-ml/jax/pull/25363). This is a new RNG scheme that produces different draws from the same distribution; that change was enough to just exceed the test tolerance with the new random draws.

The best fix is probably to bump the test tolerance.

rdyro (approved):

Issue #1171, opened 2025-01-10
backtracking linesearch huge slow down  -- recompiling functions every run of `update_fn` ?


bafflingbits:
When trying to optimize a large function, `scale_by_backtracking_linesearch` will slow to a crawl as if it is trying to recompile my function every single time.

I eventually tracked it down to this:

`_src/linesearch.py : scale_by_backtracking_linesearch : update_fn`
```
    search_state = jax.lax.while_loop(cond_fn, body_fn, search_state)
```

If I change it to:
```   
    while cond_fn(search_state):
        search_state = body_fn(search_state)
```
I go from minutes per step to many steps per second.  (I am just running on a cpu, nothing fancy here.)

I'm not well versed in the jax stuff, so I'm not actually sure what is going on. However, my `value_fn` takes a long time to jit (minutes), and the docs say `jax.lax.while_loop` will compile the functions. So as a guess, it appears that compiling `body_fn` (which doesn't take `value_fn` as an argument, but gets it from the surrounding `update_fn` scope) somehow leads to `value_fn` being re-analyzed / re-compiled eacn time update is called?

To be honest, this structure looks a bit strange in the first place, and `cond_fn` and `body_fn` are defined locally in `update_fn`, so even if this wasn't hitting `value_fn`, it still seems weird to redefine and recompile `cond_fn` and `body_fn` each time. It looks like it was done this way mostly to get values from the surrounding scope instead of by function params.  Again, I'm not vell versed in the jax stuff, so maybe I'm misunderstanding how stuff is scheduled and passed around, but rearranging the code may alleviate that as well.

rdyro:
while_loop should definitely not recompile the functions every time

Can you try the context manager: [https://jax.readthedocs.io/en/latest/_autosummary/jax.log_compiles.html](https://jax.readthedocs.io/en/latest/_autosummary/jax.log_compiles.html) to see if you're hitting recompilation many times?

bafflingbits:
Okay, I added `jax.config.update('jax_log_compiles', True)` after importing jax.

I see a lot of logs during startup, and then the following when I compile my function.
```
WARNING:2025-01-10 02:53:58,889:jax._src.interpreters.pxla:1906: Compiling _f with global shapes and types [ShapedArray(float64[120])]. Argument mapping: (UnspecifiedValue,).
[...]

PR #1175, opened 2025-01-14
minor change so backtracking init and update return the same state types

bafflingbits:
Currently if update is jitted, then first call (using state from init) will see slightly different types than the second call (using state from update) leading to an unnecessary recompilation. The difference is just in weak_type on some values, which can be controlled by explicitly setting the dtype when initializing.

(Discovered in discussion of issue #1171 )

### DIFF
--- a/optax/_src/linesearch.py
+++ b/optax/_src/linesearch.py
@@ -253,13 +253,15 @@ def scale_by_backtracking_linesearch(
       grad = otu.tree_zeros_like(params)
     else:
       grad = None
+    # base output type on params type, except only real part if complex
+    val_dtype = jnp.real(jax.tree.leaves(params)[0]).dtype
     return ScaleByBacktrackingLinesearchState(
         learning_rate=jnp.array(1.0),
-        value=jnp.array(jnp.inf),
+        value=jnp.array(jnp.inf, dtype=val_dtype),
         grad=grad,
         info=BacktrackingLinesearchInfo(
             num_linesearch_steps=0,
-            decrease_error=jnp.array(jnp.inf),
+            decrease_error=jnp.array(jnp.inf, dtype=val_dtype),
         ),
     )
 


bafflingbits:
It appears optimization with complex values may have been using that type ambiguity?
```
=========================== short test summary info ============================
FAILED test_venv/lib/python3.11/site-packages/optax/_src/alias_test.py::LBFGSTest::test_lbfgs_complex0 - TypeError: while_loop body function carry input and carry output must have equal types (e.g. shapes and dtypes of arrays), but they differ:

  * the input carry component carry[1][2].value has type complex64[] but the corresponding output carry component has type float32[], so the dtypes do not match;
  * the input carry component carry[1][2].info.decrease_error has type complex64[] but the corresponding output carry component has type float32[], so the dtypes do not match.
```

Fundamentally, it appears the issue is that init needs to know the output type of the function to be optimized, but it isn't given this information. So it guesses with a value + type that can be promoted to the actual type later? Is that what is going on?

But then on the second call to update, the type has 'narrowed', and so this triggers a recompilation.

What is the "best" way to fix this?  
Add a new optional argument to init, that allows specifying the function output type?

rdyro:
Hey, it's a very interesting contribution! Also, your analysis is correct I think

This test combines L-BFGS and the backtracing linesearch and so one of their `update_fn` is not respecting the dtype, I haven't manage to test which one (you could try digging deeper into this failing test to find a proper fix).

Overall it might be hard to ensure that all optax gradient transforms output the same type as input (it'd need a more comprehensive review). And because `init_fn` and `update_fn` are generally generated via closures it might be difficult to ensure that JAX gets cache hits on these functions every time for all transforms.

I'd be happy to merge your PR if you can find the root issue in that failing test, but overall jitting the top level might be always necessary.

bafflingbits:
Oh, since it doesn't make sense to say one complex number is < than other, even if the function to be minimized may take complex parameters, its output should still be real valued if we are going to find a (local) minimum.  So init_fn should reasonably be able to figure out the function output dtype based on the params dtype (unless someone writes a weird function specifically casting the types to break this, in which case they can pay the extra time for confusing the jit analysis).

So I think this should pass the tests now (it at least passed the bfgs tests in a local run).
If there is some helper function that finds the equivalent real dtype for a complex dtype, the code could be written a bit nicer. But hopefully the intent is clear with the comment.

optax/_src/linesearch.py
```diff
@@ -253,13 +253,15 @@ def init_fn(params: base.Params) -> ScaleByBacktrackingLinesearchState:
       grad = otu.tree_zeros_like(params)
     else:
       grad = None
+    # base output type on params type, except only real part if complex
+    val_dtype = jax.dtypes.canonicalize_dtype(jnp.real(params[0]).dtype)
```
rdyro, line 257: Can we replace this line with:
`jnp.real(jax.tree.leaves(params)[0]).dtype`

bafflingbits: I'm still learning, so I don't know what `jax.tree.leaves(params)[0]` vs `params[0]` does. So I'll defer to you. Is it for possibly handling matrices? Type wise it seems reasonable, and it avoids recompiling in my example (and still passes the bfgs tests locally). So looks good to me.

rdyro: The PR is awesome, thanks for submitting it!

`jax.tree.leaves(params)[0]` will first flatten whatever arbitrary pytree the input is, then take the first element, `params[0]` only works the same if `params` is a single array or a flat list of arrays

I'll double check it's safe to cast the linesearch type to a real (your explanation makes sense)

rdyro (approved):

PR #1177, opened 2025-01-16
Add support for loss functions with auxiliary data to linesearch

ro0mquy:
### Summary
This change adds support for loss functions that return auxiliary data alongside their primary value, like `(loss_value, extra_data)`. This pattern is commonly used with `jax.value_and_grad(fn, has_aux=True)`.

The approach:
1. Added `value_fn_has_aux` flag to `zoom_linesearch` and `scale_by_zoom_linesearch`
2. Modified value handling to properly unpack auxiliary data when needed using a new `_unpack_value` helper that extracts just the loss value
3. Updated value storage in state to keep the full value+aux tuple when needed
4. Added `has_aux` parameter to `value_and_grad_from_state` to properly handle auxiliary data when reusing cached values

This allows the linesearch algorithms to work with loss functions that return auxiliary data while maintaining the optimization over just the primary loss value.

### Input needed: How to initialize `opt_state`?
The linesearch algorithm stores `value` and `grad` in the optimizer state to enable reuse of function evaluations. When using auxiliary data, JAX compilation needs to know the structure of this data upfront.

Currently, I'm initializing it like this:
```python
# Run loss function once to get auxiliary data structure
_, aux = loss(params)
# Set value to infinity (to force recalculation) but keep aux structure
value = (jnp.asarray(jnp.inf), aux)
opt_state = optax.tree_utils.tree_set(opt_state, value=value)
```
This feels a bit hacky since it requires an extra function evaluation just to get the structure. Is there a better way to handle this initialization?

The challenge is that the auxiliary data structure is determined by the loss function and could be arbitrary (e.g., dictionaries, nested structures, etc.).

### ToDos
- [ ] Add support to backtracking linesearch
- [ ] Add documentation and doc strings
- [ ] Add tests
- [ ] Improve handling of initial `opt_state`

### DIFF
--- a/optax/_src/linesearch.py
+++ b/optax/_src/linesearch.py
@@ -579,6 +579,7 @@ def zoom_linesearch(
     curv_rtol: float = 0.9,
     approx_dec_rtol: Optional[float] = 1e-6,
     interval_threshold: float = 1e-5,
+    value_fn_has_aux: bool = False,
     verbose: bool = False,
 ) -> tuple[
     Callable[..., ZoomLinesearchState],
@@ -652,6 +653,9 @@ def zoom_linesearch(
     with Guaranteed Descent
     <https://doi.org/10.1145/1132973.1132979>`_, 2006
   """
+
+  def _unpack_value(value):
+    return value[0] if value_fn_has_aux else value
 
   def _value_and_slope_on_line(
       value_and_grad_fn: Callable[..., tuple[chex.Numeric, base.Updates]],
@@ -825,11 +829,11 @@ def zoom_linesearch(
     slope_init = state.slope_init
 
     prev_stepsize = state.stepsize
-    prev_value_step = state.value
+    prev_value_step = _unpack_value(state.value)
     prev_slope_step = state.slope
 
     safe_stepsize = state.safe_stepsize
-    safe_value = state.safe_value
+    safe_value_packed = state.safe_value
     safe_grad = state.safe_grad
 
     # Choose new point, larger than previous one or set to initial guess
@@ -842,9 +846,13 @@ def zoom_linesearch(
     else:
       max_stepsize_reached = jnp.asarray(False)
 
-    _, new_value_step, new_grad_step, new_slope_step = _value_and_slope_on_line(
-        value_and_grad_fn, params, new_stepsize, updates, fn_kwargs
+    _, new_value_step_packed, new_grad_step, new_slope_step = (
+        _value_and_slope_on_line(
+            value_and_grad_fn, params, new_stepsize, updates, fn_kwargs
+        )
     )
+
+    new_value_step = _unpack_value(new_value_step_packed)
 
     decrease_error = _compute_decrease_error(
         new_stepsize, new_value_step, new_slope_step, value_init, slope_init
@@ -855,10 +863,10 @@ def zoom_linesearch(
     # If the new point satisfies at least the decrease error we keep it
     # in case the curvature error cannot be satisfied.
     safe_decrease = decrease_error <= tol
-    new_safe_stepsize, new_safe_value, new_safe_grad = otu.tree_where(
+    new_safe_stepsize, new_safe_value_packed, new_safe_grad = otu.tree_where(
         safe_decrease,
-        [new_stepsize, new_value_step, new_grad_step],
-        [safe_stepsize, safe_value, safe_grad],
+        [new_stepsize, new_value_step_packed, new_grad_step],
+        [safe_stepsize, safe_value_packed, safe_grad],
     )
 
     # If the new point is not good, set high and low values according to
@@ -934,7 +942,7 @@ def zoom_linesearch(
         stepsize_guess=stepsize_guess,
         #
         stepsize=new_stepsize,
-        value=new_value_step,
+        value=new_value_step_packed,
         grad=new_grad_step,
         slope=new_slope_step,
         #
@@ -959,7 +967,7 @@ def zoom_linesearch(
         value_cubic_ref=value_low,
         #
         safe_stepsize=new_safe_stepsize,
-        safe_value=new_safe_value,
+        safe_value=new_safe_value_packed,
         safe_grad=new_safe_grad,
     )
     return new_state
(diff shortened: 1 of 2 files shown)


rdyro:
Hey, thanks for continuing to work on this! Can you sign the CLA please?

theo-brown:
@ro0mquy thanks for this PR, I'm keen to see this feature added!

ro0mquy:
I'm still figuring out the CLA thing with my company. But it's just a matter of time. I also accidentally pushed another fix (the "fix slope calculation in zoom_linesearch") to the same branch. I can revert that again if needed.

ro0mquy:
I signed the CLA and updated my email. Is there a way to retrigger the check?

vroulet:
Hello @ro0mquy,
Any commit will retrigger the tests, so you may merge with main for example.

By curiosity
1. why did you need the aux value returned ? (use cases would be great to understand) 

2. why not just wrapping the function that returned an aux into a function that does not return an aux? The linesearch only needs the value. At the end of the linesearch, the value and grad computed at the accepted point may get recycled by the ``value_and_grad_from_state`` function but it does not have to. You can always just recompute the value, grad (and potentially aux) by yourself. True, that would not be optimal (because you would be computing twice the value, grad) but that would work. 
Namely, when wrapping LBFGS into a solver (see the [notebook](https://optax.readthedocs.io/en/latest/_collections/examples/lbfgs.html)) consider simply defining 
```
def run_opt(init_params, fun_with_aux, opt, max_iter, tol):
  fun_without_aux = lambda *a, **kw: fun_with_aux(*a, **kw)[0]

  def step(carry):
    params, state = carry
    (value, aux), grad = jax.value_and_grad(fun_with_aux, has_aux=True)
    updates, state = opt.update(
        grad, state, params, value=value, grad=grad, value_fn=fun_without_aux
    )
    return params, state, aux

  def continuing_criterion(carry):
    _, state = carry
    iter_num = otu.tree_get(state, 'count')
    grad = otu.tree_get(state, 'grad')
    err = otu.tree_l2_norm(grad)
    return (iter_num == 0) | ((iter_num < max_iter) & (err >= tol))

  (_, init_aux), _ =  jax.value_and_grad(fun_with_aux, has_aux=True)(init_params)
  init_carry = (init_params, opt.init(init_params), init_aux)
  final_params, final_state, final_aux = jax.lax.while_loop(
      continuing_criterion, step, init_carry
  )
  return final_params, final_state, final_aux
  ```
  
But your fix seems pretty good ! 
  1. I would like the "has_aux" argument to rather be in the update function too if possible (not clear that it's possible). The reason is to keep all the logic pertaining to the actual function considered outside the signature of the method.
  2. One needs to be careful at initialization: the shape of the aux may not be known by the init function of the optimizer as you noticed. Your hack is quite ok though as long as it is documented (ideally we would change the base api so that the init function could accept additional arguments but that's a much deeper revamp).
  
Thanks for looking into this!

**Comment by theo-brown** [comment]
@vroulet Just so I understand, does
```python
    fun_without_aux = lambda *a, **kw: fun_with_aux(*a, **kw)[0]
    (value, aux), grad = jax.value_and_grad(fun_with_aux, has_aux=True)
    updates, state = opt.update(
        grad, state, params, value=value, grad=grad, value_fn=fun_without_aux
    )
 ```
 end up calling `fun_with_aux` twice for optimizers with linesearch, due to
 https://github.com/google-deepmind/optax/blob/2d1cd85bd370d7d4b4132b66dbbac9cf0f098b43/optax/_src/linesearch.py#L354

**Comment by vroulet** [comment]
@theo-brown 

A linesearch needs to evaluate the function at the candidate point to know whether it will be accepted or not. 
Ideally, if a point is accepted one would keep the evaluation of the function (and potentially the gradient) that we just did at the end of the linesearch and avoid calling a `jax.value_and_grad` when  passing back into the lines of code you present. If we have aux values currently this additional evaluation is not avoided.
I think @ro0mquy's PR was good but it's still in draft mode. @ro0mquy did you want to just present what you had and let someone else finish? (I'm ok with that, just want to clarify)

**Comment by ro0mquy** [comment]
Yes I would prefer if someone else could finish. In https://github.com/google-deepmind/optax/issues/1053 I was told to create a merge request.

**Comment by MouXuanKang** [comment]
Hi @vroulet, thanks for your question earlier! I wanted to chime in with a concrete use case that illustrates why we need `has_aux` support for linesearch.

I work with **Physics-Informed Neural Networks (PINNs)**, and here's why this feature matters:

### 1. Why we need the aux value returned?
The aux data contains *individual loss components* (PDE residuals, boundary condition losses, initial condition losses). Tracking these separately is critical for:
- Diagnosing training issues (e.g., "Is the PDE residual not decreasing, or just the boundary loss?")
- Balancing loss weights dynamically during training.

### 2. Why not just wrap the function to drop aux?
The `PINNs.PDE` function uses **second-order derivatives** (via `jax.grad`/`jax.hessian`), which are extremely expensive to compute. Wrapping the function would force us to recompute these derivatives *twice*: once for the linesearch, and once for logging the aux data. This is a major computational bottleneck.

With this PR, we could use `optax.value_and_grad_from_state(..., has_aux=True)` to **reuse the value+grad already computed during linesearch**, eliminating all duplicate computation.

Here's a simplified snippet of my setup:
```python
class PINNs:
    @staticmethod
    def loss(parameters, net, x, ...):
        # Expensive 2nd-order derivatives inside PINNs.PDE
        Ju1, Ju2 = jax.vmap(PINNs.PDE, ...)(parameters, net, ...)
        Jbc1, Jbc2 = jax.vmap(PINNs.Boundary, ...)(parameters, net, ...)
        Jic1, Jic2 = jax.vmap(PINNs.Initial, ...)(parameters, net, ...)

        # Total loss + individual components for monitoring
        Jm = a1*(Ju1+Ju2) + a2*(Jbc1+Jbc2) + a3*(Jic1+Jic2)
        return Jm, (Jm, Ju1, Ju2, Jbc1, Jbc2, Jic1, Jic2)

    @staticmethod
    def lbfgs_step(optimiser, opt_state, parameters, net, x, ...):
        def loss_fn(params): return PINNs.loss(params, net, x, ...)[0]
        
        # Ideal: Reuse value+grad from linesearch (needs this PR!)
        # (Jm, loss_components), grads = optax.value_and_grad_from_state(loss_fn, has_aux=True)(parameters, state=opt_state)
        
        # Workaround: Duplicates expensive 2nd-order derivatives
        (Jm, loss_components), grads = jax.value_and_grad(PINNs.loss, argnums=0, has_aux=True)(parameters, net, x, ...)
        
        updates, opt_state = optimiser.update(grads, opt_state, parameters, value=Jm, grad=grads, value_fn=loss_fn)
        return loss_components, opt_state, optax.apply_updates(parameters, updates)
```

Thanks again to @ro0mquy for the PR! This would be a huge quality-of-life improvement for PINNs and other computationally heavy optimization workflows.

PR #1180, opened 2025-01-22
Add 'plus' option to polyak_sgd.

### LINKED ISSUE #1178: Add an option to Polyak SGD to take the positive part of the suboptimality
**Feature request:** Add an option to the [Polyak SGD](https://optax.readthedocs.io/en/latest/api/optimizers.html#optax.polyak_sgd) optimizer that uses only the positive part of the suboptimality gap. That is, it replaces $`f(x) - f^*`$ with $`(f(x) - f^*)_+`$, where $`a_+ = \max \{a, 0\}`$ is the [ramp function](https://en.wikipedia.org/wiki/Ramp_function).

This is the variant of Stochastic Polyak Stepsize (SPS) that [Garrigos et al. 2023](https://arxiv.org/abs/2307.14528) call SPS$`_+`$.

Possible names for this option are
- `clip`
- `clamp`
- `plus`

I can submit a PR for this.

carlosgmartin:
Fixes #1178.

### DIFF
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -2314,6 +2314,7 @@ def polyak_sgd(
     scaling: base.ScalarOrSchedule = 1.0,
     f_min: float = 0.0,
     eps: float = 0.0,
+    variant: str = 'sps',
 ) -> base.GradientTransformationExtraArgs:
   r"""SGD with Polyak step-size.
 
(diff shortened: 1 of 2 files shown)


optax/_src/alias.py
```diff
@@ -2314,6 +2314,7 @@ def polyak_sgd(
     scaling: base.ScalarOrSchedule = 1.0,
     f_min: float = 0.0,
     eps: float = 0.0,
+    plus: bool = False,
```
fabianp, line 2317: I would instead have a `variant` keyword that can take (say) a string as argument, where "plus" is one of the options. This way we can accommodate other variants that we might want to implement in the future. what do you think?

carlosgmartin: Sounds good. What should the current variant be named?

carlosgmartin: Updated.

fabianp (approved):
thanks!

PR #1160, opened 2024-12-19
Move configuration for Pylint from .pylintrc to pyproject.toml.

carlosgmartin:
This is a follow-up to
- https://github.com/google-deepmind/optax/issues/1091
- https://github.com/google-deepmind/optax/pull/1110

To prevent confusion and duplication of work, I will try to keep further improvements as atomic as possible.

The sole purpose of this PR is to move the configuration for [Pylint](https://pylint.readthedocs.io/en/stable/) from `.pylintrc` to the standardized place for configuration of Python projects, namely `pyproject.toml`.

Future PRs will re-enable some of the disabled Pylint warnings after editing the code that triggers them.

PR #1188, opened 2025-02-05
Re-enable some pylint warnings.

carlosgmartin:
Follow-up to https://github.com/google-deepmind/optax/pull/1160.

Re-enables some pylint warnings, after fixing the code that triggers them.


rdyro:
Thanks for the PR! Long-term we want to move towards [the ruff linter](https://docs.astral.sh/ruff/linter/). We'd also like to keep allowing `lambda`s instead of `def`s.

Could you explore enabling any additional checks via ruff and keep allowing lambdas?

carlosgmartin:
@rdyro Done.

rdyro:
Unfortunately, these changes fail our internal linter, so we can't merge this PR

carlosgmartin:
@rdyro Can you give more detail? Is there something I can adjust?

rdyro:
We'd like to keep 
* lambdas
* keep `;` for proper jupyter notebook output suppression
* the internal linter fails importing `from mod import dog as dog`, we'll stick with `from mod import dog`
* keeping current test case base class

Closing this for now

carlosgmartin:
@rdyro I can undo those. What about the other stuff in this commit?

rdyro:
Linter changes (`pyproject.toml` changes in general) require some iteration to get to agree with the internal linter.

Can you point me to a specific issue you found with the current linting configuration?

rdyro:
Disabling the ignore in `ruff.tool.linter`: `"E402",  # allow modules not at top of file` is a good idea.

`l` might be a discouraged name by the linter, but with optax's focus on mathematical optimization, disallowing `l` is not ideal.

As far as unused imports, I'd like this to work under `ruff check`, disabling the current `ruff.tool.lint`  `"F401",  # allow unused imports`, but that requires modifying both the example notebooks and alias files (`__init__.py` files).

I'm afraid I don't understand changes to `.github/workflows/tests.yml` and `test.sh` - could you explain?

Overall, if you'd like to focus on:

> Disabling the ignore in `ruff.tool.linter`: `"E402"`

I'd be really happy to review it!

PR #1192, opened 2025-02-10
Update pyproject.toml to reflect etils version depedency.

alexalemi:
Optax tries to load the lazy_imports functionality that wasn't introduced until later version of etils.


rdyro:
This version of etils does not seem to be released for Python 3.9.

Can you point me to where the lazy imports fails?

rdyro:
The lazy_import has been removed in commit 40d886b

Closing this, thanks for finding this!

Issue #1186, opened 2025-01-30
Unexpected tracer in muon optimizer when sharding


hlzl:
Converting the tuple to an array `ns_coeffs_ = jnp.asarray(ns_coeffs)` in `scale_by_muon` can lead to an `UnexpectedTracerError` in some cases when using sharding.

This can be solved on the user's side - so feel free to close this issue if this is generally fine.

A potential fix is moving this conversion into `init_fn` and then accessing the coefficients as part of the `MuonState`. As we never update the coefficients, this is not super clean though.

To reproduce:
```python
import itertools
import jax
import jax.numpy as jnp
import numpy as np
import flax
import optax
from functools import partial
from flax.training import train_state
from jax.sharding import Mesh, PartitionSpec as P
from jax.experimental.shard_map import shard_map


def f(state, x, y):
    def loss_fn(params, x, y):
        logits = x @ params.T
        loss = optax.softmax_cross_entropy_with_integer_labels(logits, y)
        return loss.mean()

    _, grads = jax.value_and_grad(loss_fn)(state.params, jnp.squeeze(x), jnp.squeeze(y))
    grads = jax.lax.pmean(grads, axis_name="data")

    updates, new_opt_state = state.tx.update(grads, state.opt_state, state.params)
    new_params = optax.apply_updates(state.params, updates)
    return state.replace(params=new_params, opt_state=new_opt_state)


def create_train_state():
    params = jnp.zeros((5, 18))
    tx = optax.contrib.muon(learning_rate=0.05)
    return train_state.TrainState.create(apply_fn=None, params=params, tx=tx)


mesh = Mesh(np.array(jax.devices()), ("data",))
sharded_state = shard_map(create_train_state, mesh=mesh, in_specs=P(), out_specs=P())()
sharded_f = shard_map(f, mesh=mesh, in_specs=(P(), P("data"), P("data")), out_specs=P())

x = itertools.cycle([jnp.zeros((2, 4, 18))])
y = itertools.cycle([jnp.zeros((2, 4), jnp.int32)])

sharded_state = sharded_f(sharded_state, next(x), next(y))

print("success")
```
will throw:
```python
UnexpectedTracerError: Encountered an unexpected tracer. A function transformed by JAX had a side effect, allowing for a reference to an intermediate value with type float32[5,5] wrapped in a DynamicJaxprTracer to escape the scope of the transformation.
JAX transformations require that functions explicitly return their outputs, and disallow saving intermediate values to global state.
```

[...]

PR #1193, opened 2025-02-13
fix: Unexpected tracer in muon optimizer when sharding.

hlzl:
Fix for #1186


rdyro:
Hey, I think the CI was broken last time you pushed this commit (sorry!). Could you retry pushing?

hlzl:
@rdyro Seems like it went through now.

rdyro (approved):

PR #1200, opened 2025-02-17
Fix nnls to handle batch shapes correctly.

carlosgmartin:
Fixes [`nnls`](https://optax.readthedocs.io/en/latest/api/utilities.html#optax.nnls) to handle batch shapes correctly.

### DIFF
--- a/optax/_src/linear_algebra.py
+++ b/optax/_src/linear_algebra.py
@@ -294,14 +294,15 @@ def nnls(
   Uses the fast projected gradient (FPG) algorithm of Polyak 2015.
 
   Args:
-    A: Input matrix.
-    b: Input vector.
+    A: Input matrix of shape `(..., M, N)`.
+    b: Input vector of shape `(..., M)`. Leading dimensions must be broadcast-compatible
+      with the leading dimensions of `A`.
     iters: Number of iterations to run the algorithm for.
     unroll: Unroll parameter passed to `lax.scan`.
     L: An upper bound on the spectral radius of `A.T @ A` (optional).
 
   Returns:
-    The solution vector.
+    A solution vector of shape `(..., N)`.
 
   Examples:
     >>> from jax import numpy as jnp
@@ -318,8 +319,12 @@ def nnls(
     Roman A. Polyak, `Projected gradient method for non-negative least square
     <http://www.ams.org/books/conm/636/>`_, 2015
   """
-  Q = A.T @ A
-  q = A.T @ b
+  def matvec(A, x):
+    # Replace with jax.numpy.matvec once Python 3.9 support is dropped.
+    return jnp.einsum("...ij,...j", A, x)
+
+  Q = A.mT @ A
+  q = matvec(A.mT, b)
 
   if L is None:
     L = get_spectral_radius_upper_bound(Q)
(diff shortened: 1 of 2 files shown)


carlosgmartin:
@rdyro Why does the failed test say [`matvec`](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.matvec.html) doesn't exist?

rdyro:
> @rdyro Why does the failed test say [`matvec`](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.matvec.html) doesn't exist?

We're supporting Python 3.9 for which the latest JAX is 0.4.30 which doesn't yet include the matvec fn, could you try to work around it?

rdyro:
Is there a reason we're not allowing multiple rhs?

carlosgmartin:
That's just `jax.vmap(optax.nnls, [None, 1])(A, B)`, isn't it?

Most versions I see online use a vector RHS.

Replacing the RHS vector with a matrix would raise the question: Why not a rank-3 tensor? Why not a rank-4 tensor? And so on.

It seems most natural to stick with the vector and note that the others are just `vmap`s of that.

carlosgmartin:
@rdyro Thoughts on the above?

rdyro:
I'd really prefer to rely on `@` for matmul directly so both rhs vector and rhs batch of vectors cases are supported—having explicit support for 2D A and 1D and 2D b (this is not currently supported).

Batching on the left seems like the best fit for vmap.

Could you try to lower the total test cases from current 36 to something less than 8 please?

optax/_src/linear_algebra_test.py
```diff
@@ -247,6 +247,19 @@ def test_nnls(self, n, d, zero_lhs, seed, dtype, atol=1e-5):
       jnp.allclose((l100 - l10).clip(max=0), 0, atol=atol)
       jnp.allclose((l1000 - l100).clip(max=0), 0, atol=atol)
 
+  @parameterized.product(
```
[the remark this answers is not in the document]
carlosgmartin: @rdyro I've lowered the number of test cases.

carlosgmartin:
@rdyro I'm not sure I understand your comment. Currently, the interface takes

1. an input matrix `A` of shape `(..., M, N)`
2. an input vector `b` of shape `(..., M)`

[`Matvec`](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.matvec.html) has an analogous interface. This allows both `A` and `b` to have arbitrary leading batch axes, as long as they're compatible.

What would be the exact interface for your proposal?

Or do you think my version is fine?

carlosgmartin:
I see. Do you intend to support multiple axes on the RHS? (Not just one or two.)

rdyro:
I was thinking just either one vector or matrix RHS, to conform to typical linear algebra notations:

optax/_src/linear_algebra_test.py
```diff
@@ -209,43 +210,42 @@ def _gen_symmetrix_matrix(dim, condition_number):
...
-    else:
-      A = jax.random.normal(keys[0], (n, d), dtype=dtype)  # pylint: disable=invalid-name
-    b = jax.random.normal(keys[1], (n,), dtype=dtype)
 
-    x10 = linear_algebra.nnls(A, b, iters=10)
-    x100 = linear_algebra.nnls(A, b, iters=100)
-    x1000 = linear_algebra.nnls(A, b, iters=100000)
+    keys = random.split(random.key(seed), 2)
+    A = random.normal(keys[0], (m, n), dtype=dtype)  # pylint: disable=invalid-name
+    b = random.normal(keys[1], (m,), dtype=dtype)
```
rdyro, line 224: Can we add a test with $b \in R^{m \times k}$ 2 dimensional?

optax/_src/linear_algebra.py
```diff
@@ -337,12 +337,11 @@ def f(x_p_c, _):
 
     return (xn, pn, cn), None
 
-  x = jnp.zeros_like(b, shape=b.shape[:-1] + A.shape[-1:])
```
rdyro, line 340: can we add assertions

`assert A.ndim == 2 and b.ndim in (1, 2) and b.shape[0] == A.shape[-1]`

rdyro (commented):
Thanks!

optax/_src/linear_algebra.py
carlosgmartin: Did you mean `b.shape[0] == A.shape[0]`?

rdyro (approved):
Approved, thank you for this PR!

PR #1205, opened 2025-02-23
Update _combining.py changing examples to use optax.multi_transform, not optax.partition

ywrt:
Change mention of optax.partition to optax.multi_transform as optax.partition no longer exists.

### DIFF
--- a/optax/transforms/__init__.py
+++ b/optax/transforms/__init__.py
@@ -38,8 +38,8 @@ from optax.transforms._clipping import unitwise_clip
 from optax.transforms._clipping import unitwise_norm
 from optax.transforms._combining import chain
 from optax.transforms._combining import named_chain
-from optax.transforms._combining import partition
-from optax.transforms._combining import PartitionState
+from optax.transforms._combining import multi_transform
+from optax.transforms._combining import MultiTransformState
 from optax.transforms._conditionality import apply_if_finite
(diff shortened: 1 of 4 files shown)


rdyro:
This looks great!

Could you also update:
* `optax/transforms/_combining_test.py` to use multi_transform
* `optax/transforms/__init__.py` to export `multi_transform` in addition to the old `partition` (essentially adding `multi_transform = partition` and adding `multi_transform` to `__all__`)

Thank you!

ywrt:
I fully renamed partition to multi_transform, and added partition/PartitionState aliases for backward compatibility.

optax/transforms/__init__.py
```diff
@@ -80,9 +83,11 @@
     "MaskedState",
     "MultiSteps",
     "MultiStepsState",
+    "multi_transform",
+    "MultiTransformState",
     "named_chain",
     "NonNegativeParamsState",
-    "partition",
+    "partition",
```
rdyro, line 90: Could you delete the empty trailing whitespace here?

rdyro:
It looks great! Just a minor comment and let's merge this, thank you!

optax/transforms/__init__.py
ywrt: done :)

vroulet:
It was the other way around, from multi_transform to partition. It has been fixed by #1216. Thanks agian

PR #1213, opened 2025-03-01
Add more perturbation distributions to optax.perturbations.

carlosgmartin:
This PR adds more perturbation distributions to [optax.perturbations](https://optax.readthedocs.io/en/latest/api/perturbations.html).


carlosgmartin:
@vroulet Given the discussion in #1214 about using a function-based API, this PR is probably no longer necessary.

PR #1222, opened 2025-03-06
Update _schedule.py

dpoteryayev:
There was a wrong LaTeX formula shown in the cosine decay schedule description. I adjusted it accordingly of what it actually calculates in the code.


vroulet (approved):
Thanks for catching that!

PR #1233, opened 2025-03-09
Add tree_utils functions to docs that are missing.

carlosgmartin:


vroulet (approved):
Thank you for doing that!

PR #1234, opened 2025-03-09
Add defaults for min_value and max_value to tree_utils.tree_clip.

carlosgmartin:
This matches the defaults for [jax.numpy.clip](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.clip.html).

### DIFF
--- a/optax/tree_utils/_tree_math.py
+++ b/optax/tree_utils/_tree_math.py
@@ -323,18 +323,20 @@ def tree_full_like(
 
 def tree_clip(
     tree: Any,
-    min_value: Optional[jax.typing.ArrayLike],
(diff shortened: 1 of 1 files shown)


vroulet (commented):
Thank you!
As you are on it, can you modify the docs (just copy paste the following in place of Args and Returns sections)
```
  Args:
    tree: pytree.
    min_value: optional minimal value to clip all tensors to. If ``None``
      (default) then result will not be clipped to any minimum value.
    max_value: optional maximal value to clip all tensors to. If ``None``
      (default) then result will not be clipped to any maximum value.

  Returns:
    a tree with the same structure as ``tree``.
```

vroulet (approved):

Issue #1196, opened 2025-02-15
Rename `tree_scalar_mul` to `tree_scale`


carlosgmartin:
**Minor suggestion:** For convenience, rename
- [`tree_scalar_mul`](https://optax.readthedocs.io/en/latest/api/utilities.html#optax.tree_utils.tree_scalar_mul) to `tree_scale`
- [`tree_add_scalar_mul`](https://optax.readthedocs.io/en/latest/api/utilities.html#optax.tree_utils.tree_add_scalar_mul) to `tree_add_scale`

which are more concise.

If adopted, this could be done as follows:
1. Rename the current functions in the code.
2. Add back the old names as aliases. Make them emit deprecation warnings.
3. After a deprecation period, delete the old names.

I can submit a PR for this.

aryanmahawar205:
hey @carlosgmartin can you review my PR and let me know what is possibly going wrong in it?

aryanmahawar205:
hey @carlosgmartin can you have a look at my PR and help figuring out what's possibly going wrong?

rdyro:
Hey @aryanmahawar205 thanks for the contribution! I followed up on your PR #1198

shreyans413:
Hey @rdyro, if this issue isn't closed can i work on it?

rdyro:
The PR is in progress here: https://github.com/google-deepmind/optax/pull/1198 by @aryanmahawar205 but in case work have stalled, it'd be great if you could take a look!

shreyans413:
> The PR is in progress here: [#1198](https://github.com/google-deepmind/optax/pull/1198) by [@aryanmahawar205](https://github.com/aryanmahawar205) but in case work have stalled, it'd be great if you could take a look!

Thanks for pointing this out! Since the previous PR (#1198) seemed inactive, I have submitted a new PR (#1223 ) to ensure progress on this. Let me know if any further changes are needed!

aryanmahawar205:
@rdyro have updated the names across all files you had grepped throughout the repository...could you please have a look at the PR now!

PR #1246, opened 2025-03-27
[#1196] renaming tree_scalar_mul to tree_scale and tree_add_scalar_mul to tree_add_scale

aryanmahawar205:
What changes were proposed in this pull request?

This PR renames the functions tree_scalar_mul to tree_scale and tree_add_scalar_mul to tree_add_scale for improved clarity and conciseness. The old function names are kept as aliases, but now emit deprecation warnings.

Why are the changes needed?
The new names provide a more intuitive and concise representation of the functions’ purpose. The deprecation warnings alert users about the future removal of the old names after a suitable deprecation period.

Fix: https://github.com/google-deepmind/optax/issues/1196

Does this PR introduce any user-facing change?
No, this change is backward-compatible. Users will receive deprecation warnings when using the old function names, but no immediate functionality change occurs.

How was this patch tested?
The changes were verified by ensuring that the old function names now trigger deprecation warnings and that the renamed functions behave as expected.

### DIFF
--- a/docs/api/utilities.rst
+++ b/docs/api/utilities.rst
@@ -96,7 +96,7 @@ Tree
 .. autosummary::
     NamedTupleKey
     tree_add
-    tree_add_scalar_mul
+    tree_add_scale
     tree_batch_shape
     tree_cast
     tree_clip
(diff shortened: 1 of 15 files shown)


rdyro (approved):
Thanks!

Issue #296, opened 2022-02-03
Create an example for documentation for freezing parameters


rosshemsley:
Users often ask how to freeze parameters (https://github.com/deepmind/optax/issues/290).

It would be nice to add a runnable example to the documentation under Examples (next to the meta-learning example) that shows how to do this.

rosshemsley:
I suspect there's more than one valid way to freeze params, and we could potentially show each option in a notebook.

@mkunesch also gave some ideas over in https://github.com/deepmind/optax/issues/290.

pharringtonp19:
@rosshemsley As a starting point, would it be helpful to extend the meta-learning example to show case how parameters can be frozen during the inner loop? I would be happy to write something up this weekend.

rosshemsley:
Hey @pharringtonp19, that would be welcome!

Sounds good! We could use the "meta-learning trick" that means the example can be a notebook - but for the actual code it might be good to use the most simple possible model, so that users who are just learning about deep learning can follow the example, without having to worry about the meta learning part.

(maybe one of the examples from https://github.com/deepmind/optax/tree/master/examples could be made to work?)

I don't know if anyone has tried building the docs outside of the main contributors team, so do let us know if you encounter any difficulties!

pharringtonp19:
@rosshemsley I admit, I am not entirely sure what you mean by 

> building the docs outside of the main contributors team

I have never contributed to a repository before, so I am eager to see how this works. I'll start (on Thursday) with a small colab example, and then go from there (if that makes sense) -- apologies for the delay

pharringtonp19:
@rosshemsley - If you think the following is suitable, I would be happy to write up a more detailed tutorial.

Here is a [colab notebook](https://colab.research.google.com/drive/1-qLk5l09bq1NxZwwbu_yDk4W7da5TnFx?usp=sharing) where we fit a linear model via maximum likelihood. In this notebook, we show how to freeze parameters by applying the `set_to_zero` gradient transformation to the nuisance parameter `sigma`. As is expected, we recover the least squares estimate for `beta`, showcasing that MLE coincides with the Least Squares Estimate. 

-- The inspiration for this notebook was this very nice example of [Jax + MLE](https://github.com/google/jax/discussions/7385)

rosshemsley:
Hey @pharringtonp19 , sorry for dropping your previous message!

> @rosshemsley I admit, I am not entirely sure what you mean

[...]

PR #1255, opened 2025-04-02
Improve docstrings for schedule functions in optax/schedules/_schedule.py

DakshBegani:
This PR improves the docstrings for the following schedule functions as part of issue #296:

- `piecewise_constant_schedule`: added mathematical explanation and usage example
- `piecewise_interpolate_schedule`: clarified both interpolation types, added example
- `_linear_interpolate` and `_cosine_interpolate`: added short internal descriptions

These enhancements aim to make the scheduling utilities more beginner-friendly and self-explanatory.

Fixes: #296

### DIFF
--- a/optax/schedules/_schedule.py
+++ b/optax/schedules/_schedule.py
@@ -205,17 +205,30 @@ def linear_schedule(
 def piecewise_constant_schedule(
     init_value: float, boundaries_and_scales: Optional[dict[int, float]] = None
 ) -> base.Schedule:
-  """Returns a function which implements a piecewise constant schedule.
+  """Piecewise constant schedule with scaled jumps at specific boundaries.
+
+  At each step `t`, this schedule returns `init_value` scaled by the product
+  of all factors `f_i` such that `t >= b_i`, where `(b_i, f_i)` are the
+  entries in `boundaries_and_scales`.
 
   Args:
-    init_value: An initial value ``init_v``.
-    boundaries_and_scales: A map from boundaries ``b_i`` to non-negative scaling
-      factors ``f_i``. For any step count `s`, the schedule returns ``init_v``
-      scaled by the product of all factors ``f_i`` such that ``b_i < s``.
+    init_value: The starting value of the schedule.
+    boundaries_and_scales: Dictionary of `{step: scale}` where `scale` is
+      multiplied into the schedule value at the given `step`. All `scale` values
+      must be non-negative.
 
   Returns:
-    schedule
-      A function that maps step counts to values.
+    A function that maps step index to schedule value.
+
+  Example:
+    >>> sched = optax.piecewise_constant_schedule(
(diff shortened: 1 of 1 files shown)


vroulet (commented):
Looks very nice to me. Can you fix the doctests?
Replacing the calls `sched(...)` by `print(sched(...))` should work

DakshBegani:
@vroulet Fixed it by wrapping schedule calls in print(...). Let me know if anything else needs updating!

optax/schedules/_schedule.py
```diff
@@ -390,27 +405,42 @@ def piecewise_interpolate_schedule(
...
+    >>> sched = optax.piecewise_interpolate_schedule(
+    ...     interpolate_type='linear',
+    ...     init_value=1.0,
+    ...     boundaries_and_scales={100: 0.5, 200: 0.1})
+    >>> print(sched(0))
+    1.0
+    >>> print(sched(100))
+    0.5
+    >>> print(sched(150))
+    0.3
```
vroulet, line 434: 0.275
(see doctest errors)

optax/schedules/_schedule.py
```diff
@@ -390,27 +405,42 @@ def piecewise_interpolate_schedule(
...
+    >>> print(sched(200))
+    0.1
```
vroulet, line 436: 0.05
(see doctest errors)

vroulet (commented):
The behavior of this function is actually counter-intuitive.
Thanks again for actually making the examples.
Unfortunately, your doc does not reflect well the actual usage (see doctest errors).
This does not just interpolate, say a linear function, at the points (x=boundary, y=scales). It accumulates the scaling at the boundary values and then interpolates...
Yes, I agree, this makes little sense... Anyway, for now, the previous docs were describing this weird behavior well. So maybe keep it and rather just add examples.

DakshBegani:
I see now that the behavior isn't as straightforward as I initially thought. Just to clarify, the function accumulates scaling at the boundary values first and then interpolates, rather than performing simple linear interpolation at those points? Should I keep the original documentation as it was, with the explanation of the behavior, and just add the examples that reflect this more complex behavior?

I want to make sure I am on the right track

vroulet:
> Just to clarify, the function accumulates scaling at the boundary values first and then interpolates, rather than performing simple linear interpolation at those points?

Yes, it's counter-intuitive I agree. So your effort of documenting better with examples will really help!

> Should I keep the original documentation as it was, with the explanation of the behavior, and just add the examples that reflect this more complex behavior?

Yes, exactly. Keep the previous doc but add the examples.

Thank you for doing this @DakshBegani !

DakshBegani:
Glad it could be helpful - let me know if there are any further issues i can work on!
Thanks for helping out @vroulet

vroulet (approved):

PR #1260, opened 2025-04-09
Minor improvements to tree_batch_shape.

carlosgmartin:
Minor improvements to [tree_batch_shape](https://optax.readthedocs.io/en/latest/api/utilities.html#optax.tree_utils.tree_batch_shape) (follow-up to https://github.com/google-deepmind/optax/pull/1161):
- Change `x.shape` to `jnp.shape(x)`, in case `x` is an [ArrayLike](https://docs.jax.dev/en/latest/_autosummary/jax.typing.ArrayLike.html) but not an [Array](https://docs.jax.dev/en/latest/_autosummary/jax.Array.html) (e.g. a `float` or `int`).
- Change tuple concatenation to tuple unpacking, in case `shape` is a Sequence but not a tuple (e.g. a list).

### DIFF
--- a/optax/tree_utils/_tree_math.py
+++ b/optax/tree_utils/_tree_math.py
@@ -268,7 +268,7 @@ def tree_batch_shape(
   Returns:
     a pytree with the leading batch dimensions added.
   """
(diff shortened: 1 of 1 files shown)


vroulet (approved):
Thanks !

PR #1261, opened 2025-04-09
Edit sigmoid_binary_cross_entropy and sigmoid_focal_loss to work correctly with non-array label arguments.

carlosgmartin:
Edit [sigmoid_binary_cross_entropy](https://optax.readthedocs.io/en/latest/api/losses.html#optax.losses.sigmoid_binary_cross_entropy) and [sigmoid_focal_loss](https://optax.readthedocs.io/en/latest/api/losses.html#optax.losses.sigmoid_focal_loss) to work correctly with non-array `label` arguments by replacing `x.astype(dt)` with `jnp.astype(x, dt)`.


vroulet (approved):
Good catch, thank you

PR #1262, opened 2025-04-09
Fix docs for optax.contrib.dog.

carlosgmartin:
Fix docs for [optax.contrib.dog](https://optax.readthedocs.io/en/latest/api/contrib.html#optax.contrib.dog).

### DIFF
--- a/optax/contrib/_dog.py
+++ b/optax/contrib/_dog.py
@@ -139,8 +139,12 @@ def dog(
   .. math::
 
     \begin{align*}
-      \eta_t &= \frac{\max\{r_\epsilon, \max_{i\le t}{\|x_i-x_0\|}\}}{
+      \eta_t &= \frac{\bar r_t}{
         \sqrt{\sum_{i\le t}{\|g_i\|^2+\epsilon}}}\\
+      \bar r_t & = \begin{cases}
+        \max_{i\le t}{\|x_i-x_0\|} & \text{if } t \ge 1 \\
+        r_\epsilon & \text{if } t = 0
+      \end{cases} \\
       x_{t+1} & = x_{t} - \eta_t\, g_t
     \end{align*}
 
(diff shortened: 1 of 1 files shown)


optax/contrib/_dog.py
```diff
@@ -139,15 +139,15 @@ def dog(
   .. math::
 
     \begin{align*}
-      \eta_t &= \frac{\max_{i\le t}{\|x_i-x_0\|}}{
-        \sqrt{\sum_{i\le t}{\|g_i\|^2+eps}}}\\
-      x_{t+1} & = x_{t} - \eta_t\, g_t,
+      \eta_t &= \frac{\max\{r_\epsilon, \max_{i\le t}{\|x_i-x_0\|}\}}{
```
vroulet, line 142: max is strange. The right equations should be
```
\begin{align*}
  \eta_t &= \frac{\tilde r_t}{
    \sqrt{\sum_{i\le t}{\|g_i\|^2+\epsilon}}}\\
  \tilde r_t & = \begin{cases}
    \max_{i\le t}{\|x_i-x_0\|} & \text{if } t \ge 1 \\
    r_\epsilon & \text{if } t = 0
  \end{cases} \\
  x_{t+1} & = x_{t} - \eta_t\, g_t
\end{align*}
```

vroulet (commented):
Thank you agian @carlosgmartin 
You may add the following warning

```
..warning:: The authors recommend using model averaging with this optimizer.
```

Either in this PR or in another one, we should move [scale_by_distance_over_gradients](https://github.com/google-deepmind/optax/blob/89c660614a1ae91437fa636489a01e344daa1c19/optax/_src/transform.py#L1331C5-L1331C37) into this file. Let me know if you want to do it or if I shall do it.

optax/contrib/_dog.py
carlosgmartin: @vroulet I was following the formula at the top of page 5 in https://arxiv.org/pdf/2302.12022#page=5.

vroulet: The code is doing the formula above. I took it from http://github.com/formll/dog. If you want to use your formula, the code should be changed too (I'm ok with that too)

vroulet (approved):
Thanks

PR #1276, opened 2025-04-23
Fix MuonState docstring

EIFY:
MuonState is for Muon, not Adam. This is probably copy-pasta.


vroulet (approved):
Thanks for catching this

PR #1284, opened 2025-04-29
Add Freezing Utilities: `freeze`, `selective_optimizer`, and Prefix-Mask Support

pranavagrawaI:
### 1. New Utilities

- **`freeze(mask)`**  
  Zeroes updates for parameters marked `True` in a (prefix-compatible) boolean mask.

- **`selective_optimizer(optimizer, freeze_mask)`**  
  Applies `optimizer` to parameters where `freeze_mask` is `False`, and freezes the rest via `set_to_zero`.

**Mask semantics:**
- `True`  ⇒ freeze
- `False` ⇒ train
- Missing entries ⇒ default to train

### 2. Prefix-Compatible Mask Expansion

- Implements a helper `_broadcast_mask(params, mask)`:
  - Expands scalar and prefix masks to match full PyTree structure.
  - Supports `dict`, `list`, `tuple`, and array leaves.
  - Differentiates containers vs. scalars safely.
  - Defaults unlabeled branches to `False` (i.e., train).

### 3. Docstrings

- All functions have full docstrings with behavior notes.

## Test Coverage

- Three new test suites `BroadcastMaskTest`, `FreezeTest` and `SelectiveOptimizerTest`.


rdyro:
Hey, thanks for the PR! optax should already support [prefix trees](https://docs.jax.dev/en/latest/_autosummary/jax.tree.map.html) for mask (e.g.,` mask = [True, False, True]` and `params = [model1, model2, another_state]` where `model1, model2, another_state` can be nested, so I think no need for custom traversal.

 Do you have an application in mind where the freeze and selective_optimizer utilities could come in handy?

optax/_src/utils.py
```diff
@@ -25,12 +25,12 @@
 import jax
 import jax.numpy as jnp
 import jax.scipy.stats.norm as multivariate_normal
+from jax import tree_util
```
rdyro, line 28: jax prefers the full `jax.tree.map` in place of `tree_util.map` - could you please update uses of `tree_util` and remove the import

optax/_src/utils.py
```diff
@@ -333,6 +333,178 @@ def _value_and_grad(
...
+    if isinstance(params, tuple):
+        mask_seq = list(mask) if isinstance(mask, (list, tuple)) else []
+        mask_seq = (mask_seq + [False] * len(params))[: len(params)]
+        return tuple(_broadcast_mask(p, m, _root=False)
+                     for p, m in zip(params, mask_seq))
+
+    return False
+
+
+def freeze(mask: base.PyTree) -> base.GradientTransformation:
```
rdyro, line 418: let's rely on jax's prefix tree behavior for jax.tree.map and require explicity prefix/leaf False - for more predictable behavior

optax/_src/utils.py
```diff
@@ -333,6 +333,178 @@ def _value_and_grad(
...
+
+    Args:
+      mask: Boolean PyTree (potentially prefix) indicating which parameters to
+        freeze (`True` = freeze, `False` = train).
+
+    Returns:
+      An ``optax.GradientTransformation`` that zeroes out gradient updates based
+      on the dynamically broadcasted mask.
+    """
+    from optax.transforms import masked
```
rdyro, line 442: what's the reason for imports not being at the top?

optax/_src/utils.py
```diff
@@ -333,6 +333,178 @@ def _value_and_grad(
...
+
+    Args:
+      optimizer: The Optax ``GradientTransformation`` to apply to trainable params.
+      freeze_mask: Boolean PyTree (potentially prefix) indicating which parameters
+        to freeze (`True` = freeze, `False` = train).
+
+    Returns:
+      An ``optax.GradientTransformation`` that selectively applies the `optimizer`.
+    """
+    from optax.transforms import partition
```
rdyro, line 487: what's the reason for imports not being at the top?

rdyro:
I'm worried about maintaining custom traversal logic, so I'd like to avoid it if possible unfortunately. Also, in the spirit of simplicity, we should probably require an explicit prefix `False` for not-/freezing behavior.

Overall great contribution, thanks so much!

optax/_src/utils.py
pranavagrawaI: I was facing an issue with cyclic imports when importing at the top, so I moved the imports inside the functions.

Let me know if there is a better way to handle this and I will do that instead.

pranavagrawaI:
I have moved the shortcuts to a `optax/transforms/_freezing.py` and the tests to `optax/transforms/_freezing_test.py`.

Also, do you want me to update the freezing parameters example to also include the new shortcuts or perhaps give the examples in the docstrings?

optax/transforms/_freezing.py
```diff
@@ -0,0 +1,93 @@
...
+
+    Returns:
+        An Optax `GradientTransformation` which applies `set_to_zero()` wherever
+        `mask==True`, and leaves other gradients intact.
+    """
+    def mask_fn(params: base.PyTree) -> base.PyTree:
+        if _is_bool_scalar(mask):
+            return jax.tree.map(lambda _: bool(mask), params)
+
+        chex.assert_trees_all_equal_structs(params, mask)
```
rdyro, line 56: This will likely fail if we're working with a partial prefix mask, right? Can we remove this line?

optax/transforms/_freezing_test.py
```diff
@@ -0,0 +1,283 @@
...
+GRAD_NESTED = {
+    'layer1': [jnp.full((2, 3), 1.0), {'bias': jnp.full(3, 2.0)}],
+    'layer2': {'w': jnp.full((4, 2), 3.0), 'b': jnp.full(4, 4.0)},
+    'misc': (jnp.array(1.0),)
+}
+
+
+class FreezeTest(parameterized.TestCase):
+
+  @parameterized.named_parameters([
```
rdyro, line 45: Can you add a partial prefix mask as a test case?

The tests have a full prefix mask and a full mask, but we're missing a partial prefix mask like:
```
params = [a, [b, c]]
mask = [False, True]
```

optax/transforms/_freezing_test.py
```diff
@@ -0,0 +1,283 @@
...
+    updates, new_state = optimizer.update(grads, state, params)
+
+    chex.assert_trees_all_close(updates, expected_updates, atol=0)
+    chex.assert_trees_all_equal(state, new_state)
+    chex.assert_trees_all_equal(state.inner_state, base.EmptyState())
+
+
+class SelectiveOptimizerTest(parameterized.TestCase):
+
+  @parameterized.named_parameters([
```
rdyro, line 118: Like above, we need a partial prefix mask test

optax/transforms/_freezing_test.py
```diff
@@ -0,0 +1,283 @@
...
+    expected = (
+      jax.tree.map(jnp.zeros_like, PARAMS_FLAT)
+      if bool(scalar_mask) else GRAD_FLAT
+    )
+    chex.assert_trees_all_close(updates, expected, atol=0)
+
+  def test_bad_structure_raises(self):
+    bad_mask = {'layer1': [True]}  # missing the bias leaf
+    opt = _freezing.freeze(bad_mask)
+    with self.assertRaises(Exception):
```
rdyro, line 247: Can you make the exception more specific and check for a specific message using the more specific: `self.assertRaisesRegex` please?

[https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex)

optax/transforms/_freezing_test.py
```diff
@@ -0,0 +1,283 @@
...
+  def test_scalar_freeze_all(self, scalar_mask):
+    opt = _freezing.selective_optimizer(alias.sgd(1.0), scalar_mask)
+    updates, _ = opt.update(GRAD_FLAT, opt.init(PARAMS_FLAT), PARAMS_FLAT)
+    new_params = update.apply_updates(PARAMS_FLAT, updates)
+    chex.assert_trees_all_close(new_params, PARAMS_FLAT, atol=1e-6)
+
+  def test_selective_bad_structure(self):
+    bad_mask = {'a': True}  # missing 'b'
+    opt = _freezing.selective_optimizer(alias.sgd(1.0), bad_mask)
+    with self.assertRaises(Exception):
```
rdyro, line 278: Can you make the exception more specific, see above

optax/transforms/_freezing.py
```diff
@@ -0,0 +1,93 @@
...
+        each leaf indicates whether that specific parameter leaf should be
+        frozen (True) or left unchanged (False).
+
+    Returns:
+        An Optax `GradientTransformation` which applies `set_to_zero()` wherever
+        `mask==True`, and leaves other gradients intact.
+    """
+    def mask_fn(params: base.PyTree) -> base.PyTree:
+        if _is_bool_scalar(mask):
+            return jax.tree.map(lambda _: bool(mask), params)
```
rdyro, line 54: This will likely throw a concretization error if mask is a tracer.

Because tree.map supports a prefix map, you can rely on jax.tree.map directly in both scalar and nested case.

Does `return masked(base.set_to_zero(), mask)` work in all your test cases? (Instead of creating a `mask_fn`)

rdyro:
> I have moved the shortcuts to a `optax/transforms/_freezing.py` and the tests to `optax/transforms/_freezing_test.py`.
> 
> Also, do you want me to update the freezing parameters example to also include the new shortcuts or perhaps give the examples in the docstrings?

Thank you!

Yes, please! I also left some additional review comments, this is looking great!

pranavagrawaI:
### The following changes have been made:

- Simplified the implementations of `freeze `and `selective_optimizer`.
- Removed the internal helper function `_is_bool_scalar`.
- Added a new test case specifically for handling partial prefix masks and combine them to form a single test class for each function.
- Tests now use `self.assertRaisesRegex` 
- Updated the documentation for `freeze` and `selective_optimizer` to explicitly state the requirement for static masks (masks not changing during training).
- Included small usage example in the docstrings for both functions.
- Added freeze and selective_optimizer to optax/transforms/__init__.py, to make them accessible directly under the `optax.transforms` namespace.

I thought of adding specific example uses of these util in [examples/freezing_parameters.ipynb](https://github.com/google-deepmind/optax/blob/main/examples/freezing_parameters.ipynb); I have the notebook ready, but decided not to commit it here, as it maybe better to do that in a separate PR, let me know if you want it here instead.

optax/transforms/_freezing.py
```diff
@@ -0,0 +1,86 @@
...
+        essentially freezing (i.e. holding constant) those parameters.
+
+        The mask must be static (i.e., not dependent on runtime values or updated
+        during training) and can be:
+          • a single boolean (or 0-d JAX bool array), causing every parameter to be
+            either all-frozen (True) or all-trainable (False), or
+          • a PyTree of booleans matching the structure of the parameters, where
+            each leaf indicates whether that specific parameter leaf should be
+            frozen (True) or left unchanged (False).
+
```
rdyro, line 40: Can you add a `.. seealso::` linking to the other function? (an example here: [https://github.com/google-deepmind/optax/blob/5d86211f737a9fe7882b6e22266e9910f3112185/optax/_src/linesearch.py#L245](https://github.com/google-deepmind/optax/blob/5d86211f737a9fe7882b6e22266e9910f3112185/optax/_src/linesearch.py#L245))

optax/transforms/_freezing.py
```diff
@@ -0,0 +1,86 @@
...
+    return masked(base.set_to_zero(), mask)
+
+
+def selective_optimizer(
+    optimizer: "base.GradientTransformation",
+    freeze_mask: Union[bool, chex.ArrayTree],
+) -> "base.GradientTransformation":
+    """
+        Partition updates so that only un-frozen parameters are optimized.
+
```
rdyro, line 59: Can you add a .. seealso:: linking to the other function?

optax/transforms/_freezing.py
```diff
@@ -0,0 +1,86 @@
...
+            >>> freezer = freeze(mask)
+
+        Returns:
+            An Optax `GradientTransformation` which applies `set_to_zero()` wherever
+            `mask==True`, and leaves other gradients intact.
+    """
+    return masked(base.set_to_zero(), mask)
+
+
+def selective_optimizer(
```
rdyro, line 53: Can we rename to `selective_transform` instead of optimizer, please?

optax/transforms/_freezing.py
pranavagrawaI: a .. seealso:: has been added to both freeze and selectve_transform (previously selective_optimizer)

optax/transforms/_freezing.py
pranavagrawaI: a .. seealso:: has been added to both freeze and selectve_transform (previously selective_optimizer)

optax/transforms/_freezing.py
pranavagrawaI: named changed to selective_transform

rdyro (approved):
Great contribution, thanks!

pranavagrawaI:
Happy to Contribute! Just one more thing, it says that one internal check has failed, but I seem to be unable to see the details, let me know if there is a problem and I can fix it.

rdyro:
> Happy to Contribute! Just one more thing, it says that one internal check has failed, but I seem to be unable to see the details, let me know if there is a problem and I can fix it.

We have some internal CI that's somewhat difficult to mirror in the open-source github, I have to resolve these internally. It's usually small changes.

rdyro:
Can you squash your commits please? We're having some trouble merging the chain internally @pranavagrawaI

PR #1292, opened 2025-05-02
Clean up and fix errors in DoG implementation and documentation.

carlosgmartin:
Cleans up and fixes errors in the implementation and documentation for Distance over Gradients (DoG): [optax.contrib.dog](https://optax.readthedocs.io/en/latest/api/contrib.html#optax.contrib.dog).

### DIFF
--- a/optax/contrib/_dog.py
+++ b/optax/contrib/_dog.py
@@ -23,7 +23,7 @@ References:
 """
 
 from collections.abc import Callable
-from typing import Any, NamedTuple, Optional, Union
+from typing import Any, NamedTuple, Optional, Union, Literal
 
 import chex
 import jax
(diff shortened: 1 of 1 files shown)


optax/contrib/_dog.py
```diff
@@ -81,41 +75,28 @@ def init_fn(params: base.Params) -> DoGState:
...
+        estim_dist=jnp.asarray(init_step_length, dtype=params_dtype),
         sum_sq_norm_grads=jnp.asarray(0.0, dtype=params_dtype),
     )
 
   def update_fn(
       updates: base.Updates, state: DoGState, params: base.Params
   ) -> tuple[base.Updates, DoGState]:
-
-    # Reduces to norm of init_params for first step
     curr_distance = otu.tree_l2_norm(otu.tree_sub(state.init_params, params))
```
vroulet, line 87: Why not keeping this heuristic? This was taken from https://github.com/formll/dog/blob/main/src/dog/dog.py which is the official repo of the paper.

vroulet (commented):
Thanks! Could you explain why you wanted to remove the heuristic? 
Also if you are on it, one needs to add notes for each of those algorithms saying that they should be averaged. 
And finally one needs to remove or merge with the duplicate algorithm that is [here](https://github.com/google-deepmind/optax/blob/c6516f7429282a516112088c7d0d1b493c3d4952/optax/_src/transform.py#L1331). That's on me because I'm the one who created the duplicate in the contrib folder (the one in transforms was never documented so I had not found it).
Thank you again for all

optax/contrib/_dog.py
carlosgmartin: Could you clarify what you mean by heuristic? This PR still allows for an optional `init_learning_rate`.

vroulet: From the previous docs:
```
the first step size is given by:
(reps_rel * (1+\|x_0\|)) / (\|g_0\|^2 + eps)^{1/2}
```
I am not sure that that heuristic for the first step has been kept in your modifications but correct me if I'm wrong

carlosgmartin: My implementation matches the paper's (see the top of [page 5](https://arxiv.org/pdf/2302.12022#page=5)), as well as the algorithm presented in the [readme](https://github.com/formll/dog/tree/main?tab=readme-ov-file#algorithm) of the repo you linked to, so it should reduce to the same thing (just expressed as a learning rate vs. as a step length).

vroulet: Ok, but typically the $r_\epsilon$ depends on the parameters https://github.com/formll/dog/tree/main?tab=readme-ov-file#choosing-reps_rel
I agree that it could be coded in a more readable way (maybe "use_heuristic" option? and maybe it could be integrated in the initial stepsize too). And happy to even remove that heuristic if you prefer but it should be mentioned somewhere.

carlosgmartin: Oh, I see what you're saying. The heuristic appears at the bottom of [page 10](https://arxiv.org/pdf/2302.12022#page=10) ("Setting $r_\epsilon$").

The interface is further complicated by the (optional) `init_learning_rate` parameter.

One possibility that unifies all 3 is to have a single argument of the following form:

```python3
initial_step: tuple[Literal["length", "learning_rate", "heuristic"], float]
```
It accepts a tag-float pair and, for the initial step, sets either the length, learning rate, or heuristic parameter $\alpha$ to the given float. This is essentially a sum type.

WDYT? So far, this seems like the tidiest approach to me (it cleanly ensures disjointness). But also open to other ideas.

vroulet: That sounds pretty good to me

carlosgmartin: @vroulet Done.

vroulet (commented):
Looks great to me, thanks again @carlosgmartin.
Can you merge with head so that we can merge it?

optax/contrib/_dog.py
```diff
@@ -37,86 +37,72 @@
...
   def init_fn(params: base.Params) -> DoGState:
     # Define state parameters with the lowest dtype of the parameters to avoid
     # dtype promotion of parameters resulting in a dtype mismatch between
     # parameters and updates.
     params_dtype = optax.tree.dtype(params, 'lowest')
+
+    if init_step_type == "distance":
+      r_epsilon = init_step_value
+    elif init_step_type == "heuristic":
+      r_epsilon = init_step_value * (1 + optax.tree.norm(params))
```
vroulet, line 81: Need to add a warning in the docstring. 
Many init functions do not really care about the actual value of the params. This is not the case here. Feeding this optimizer with a "`dummy_params`" would lead to errors.
(That's the reason why it was strangely coded before).

carlosgmartin: @vroulet Done.

vroulet (approved):
Thanks!

Issue #1309, opened 2025-06-03
make_perturbed_fun gradient can throw TypeError


adconner:
make_perturbed_function can give a TypeError in the gradient computation. Here is a minimal bad example. The problem seems to be related to the shape of the argument, in this case x.

```
import jax
import optax

rng = jax.random.key(1)

x = jax.numpy.ones((1,2))
def f(x):
    return x[0,0]

fp = optax.perturbations.make_perturbed_fun(f, 10)

print(f(x)) # works
print(fp(x,rng)) # works
print(jax.grad(f)(x)) # works
print(jax.grad(fp)(x, rng)) 
# TypeError: dot_general requires contracting dimensions to have the same shape, got (2,) and (1,).
```

carlosgmartin:
Since https://github.com/google-deepmind/optax/pull/1310 has been merged, I think this issue can be closed.

rdyro:
Fixed by [https://github.com/google-deepmind/optax/pull/1310](https://github.com/google-deepmind/optax/pull/1310)

Issue #1308, opened 2025-06-03
perturbations.make_perturbed_fun has incorrectly scaled gradients


adconner:
My first experiments with make_perturbed_fun reveal that the gradient calculation is incorrect. With the identity function and normal noise, the perturbed function is mathematically the identity function (so we get an unbiased estimator). Its Jacobian then, should be an unbiased estimator of the identity matrix. What we get seems to be off by a factor of sigma.

```
In[73]: f=optax.perturbations.make_perturbed_fun(lambda x: x, sigma=0.1, noise=optax.perturbations.Normal())

In [75]: f(jnp.array(1.0), jax.random.PRNGKey(1))
Out[75]: Array(0.998676, dtype=float32)

In [76]: jax.grad(f)(jnp.array(1.0), jax.random.PRNGKey(1))
Out[76]: Array(0.08708793, dtype=float32)

In [77]: jax.grad(f)(jnp.array(0.0), jax.random.PRNGKey(1))
Out[77]: Array(0.10032693, dtype=float32)

In [78]: f2=optax.perturbations.make_perturbed_fun(lambda x: x, sigma=0.5, noise=optax.perturbations.Normal())

In [79]: f2(jnp.array(1.0), jax.random.PRNGKey(1))
Out[79]: Array(0.9933809, dtype=float32)

In [80]: jax.grad(f2)(jnp.array(1.0), jax.random.PRNGKey(1))
Out[80]: Array(0.48839524, dtype=float32)

In [81]: jax.grad(f2)(jnp.array(0.0), jax.random.PRNGKey(1))
Out[81]: Array(0.50163394, dtype=float32)

In [83]: f2(jnp.ones(4), jax.random.PRNGKey(1))
Out[83]: Array([0.9933809, 1.0099303, 1.0249068, 1.0073029], dtype=float32)

In [84]: jax.jacobian(f2)(jnp.ones(4), jax.random.PRNGKey(1))
Out[84]:
Array([[ 4.8839524e-01,  1.3756592e-02,  3.2798059e-02,  2.7877303e-02],
       [-1.9342428e-02,  5.2607626e-01,  5.0115172e-02,  2.5842471e-02],
       [-3.0254519e-02,  2.0161595e-02,  5.3830910e-01,  2.4624784e-02],
       [ 3.2641597e-05,  3.1096857e-02,  5.9832685e-02,  5.2487993e-01]],      dtype=float32)

```

carlosgmartin:
[...]

Issue #1214, opened 2025-03-01
Rename `optax.perturbations` distribution method `log_prob` to `unnormalized_log_prob`


carlosgmartin:
Each distribution under [optax.perturbations](https://optax.readthedocs.io/en/latest/api/perturbations.html) has a method called `log_prob` that returns the logarithm of the _unnormalized_ probability density at a given point.

I propose renaming this method to `unnormalized_log_prob`, like in [TensorFlow Probability](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Distribution#unnormalized_log_prob), so users don't confuse it with the actual (normalized) log probability.

If this is acceptable, I can submit a PR for it.

vroulet:
I think the most important point is that the method ``make_perturbed_fun`` does not require normalized log probabilities hence the advantage of highlighting that with the ``unnormalized_log_prob`` name. This can at least be highlighted in the docs of ``make_perturbed_fun``.

The main issue is that there is no class for "noises" like Gumbel or Normal. I believe the underlying idea (pinging the author of the code @q-berthet) was that the ``make_perturbed_fun`` could be used with standard sampling libraries like distrax which share similar attributes (``sample``, ``log_prob`` for their distributions). 

I think it would be best to consolidate this situation either 
- by asking for two arguments (noise_sampler, noise_unnormalized_log_prob) in the signature of ``make_perturbed_fun`` 
- or by making a proper ``Noise`` class (which can then have sampler, log, log_prob etc...)
I would be more in favor of the first solution, which will require some future deprecated behavior. Ideally we would like not to recode a library of distributions like distrax so making our own class may clash with well thought existing libraries.

Thanks for looking into that

vroulet:
Sorry I had not seen #1213. So before considering merging #1213, we should create a proper class or simply make these distributions some simple NamedTuples with sample, and unnormalized_log_prob as attributes.

carlosgmartin:
You raise an excellent point. I actually had the same thought: It would be better to make the interface function-based rather than object-based, since functions are more "nimble" (and let us easily reuse existing JAX code like [jax.random.normal](https://docs.jax.dev/en/latest/_autosummary/jax.random.normal.html) and [jax.scipy.stats.norm.logpdf](https://docs.jax.dev/en/latest/_autosummary/jax.scipy.stats.norm.logpdf.html)).

[...]

PR #1310, opened 2025-06-04
Fix many issues with make_perturbed_fun

adconner:
Issues fixed:
1) Handle arbitrarily shaped arrays in inputs.
   See #1309 
2) Scale the jvp correctly. The previous implementation was missing a
   factor of 1/sigma (see Prop 3.1 of https://arxiv.org/abs/2002.08676
   for correct formula).
   See #1308 

Enhancements:
1) Vmap over samples instead of using unrolled loops. This greatly
   improves the previously terrible compile times and should also scale
   better on hardware.
2) Use custom_jvp.defjvp instead of custom_jvp.defjvps. This allows us
   to be explicit that the rng input is not differentiable, and also
   avoids calculating _compute_residuals twice during jvp.
3) Use jax.jvp instead of jax.grad for the computation of the
   grad log_prob inner product. This is minor, but potentially valuable
   for more complicated probability distributions.

I also considered making the design change discussed in #1214, where it was proposed to use the interface sample_and_log_prob, but I now believe this is impossible. We don't actually care about the log_prob for our samples, but its gradient. The data dependence through the differentiation of log_prob applied to the result of sample requires two separate forward passes, even if we have an efficient function which can compute the log_prob value along with sample in only one pass.

### DIFF
--- a/optax/perturbations/_make_pert.py
+++ b/optax/perturbations/_make_pert.py
@@ -24,6 +24,7 @@ import jax
 import jax.numpy as jnp
 from optax._src import base
 import optax.tree
+import functools
 
 
 class Normal:
@@ -56,37 +57,6 @@ class Gumbel:
     return -inputs - jnp.exp(-inputs)
 
 
-def _tree_mean_across(trees: Sequence[chex.ArrayTree]) -> chex.ArrayTree:
-  """Mean across a list of trees.
-
-  Args:
-    trees: List or tuple of pytrees with the same structure.
-
-  Returns:
-    a pytree with the same structure as each tree in ``trees`` with leaves being
-    the mean across the trees.
-
-  Examples:
-    >>> optax.tree.reduce_mean_across(
-    ...   [{'first': [1, 2], 'last': 3},
-    ...    {'first': [5, 6], 'last': 7}]
-    ...   )
-    {'first': [3, 4], 'last': 5}
-  """
-  mean_fun = lambda x: sum(x) / len(trees)
-  return jax.tree.map(lambda *leaves: mean_fun(leaves), *trees)
-
-
-def _tree_vmap(
-    fun: Callable[[chex.ArrayTree], chex.ArrayTree],
-    trees: Sequence[chex.ArrayTree],
-) -> chex.ArrayTree:
-  """Applies a function to a list of trees, akin to a vmap."""
-  tree_def_in = jax.tree.structure(trees[0])
-  has_in_structure = lambda x: jax.tree.structure(x) == tree_def_in
-  return jax.tree.map(fun, trees, is_leaf=has_in_structure)
-
-
 def make_perturbed_fun(
     fun: Callable[[chex.ArrayTree], chex.ArrayTree],
     num_samples: int = 1000,
@@ -116,7 +86,8 @@ def make_perturbed_fun(
 
   Returns:
     A function with the same signature (plus an additional rng in the input)
-    that can be automatically differentiated.
+    that can be automatically differentiated. Second order differentiation is
+    currently not implemented.
 
   References:
     Berthet et al., `Learning with Differentiable Perturbed Optimizers
@@ -130,32 +101,31 @@ def make_perturbed_fun(
       inputs: chex.ArrayTree, rng: chex.PRNGKey
   ) -> tuple[chex.ArrayTree, chex.ArrayTree]:
     # random noise Zs to be added to inputs
-    samples = [
-        optax.tree.random_like(rng_, inputs, sampler=noise.sample)
-        for rng_ in jax.random.split(rng, num_samples)
-    ]
+    samples = jax.vmap(lambda rng_: optax.tree.random_like(rng_, inputs, sampler=noise.sample))(
+        jax.random.split(rng, num_samples))
+
     # creates [inputs + Z_1, ..., inputs + Z_num_samples]
-    inputs_pert = _tree_vmap(
-        lambda z: optax.tree.add_scale(inputs, sigma, z), samples
-    )
+    inputs_pert = jax.vmap(lambda z: optax.tree.add_scale(inputs, sigma, z))(samples)
+
     # applies fun: [fun(inputs + Z_1), ..., fun(inputs + Z_num_samples)]
-    outputs_pert = _tree_vmap(fun, inputs_pert)
+    outputs_pert = jax.vmap(fun)(inputs_pert)
     return outputs_pert, samples
 
-  @jax.custom_jvp
-  def fun_perturb(inputs: chex.ArrayTree, rng: chex.PRNGKey) -> chex.ArrayTree:
-    outputs_pert, _ = _compute_residuals(inputs, rng)
+  @functools.partial(jax.custom_jvp,nondiff_argnums=(1,))
+  def fun_perturb(primal_in: chex.ArrayTree, rng: chex.PRNGKey) -> chex.ArrayTree:
+    outputs_pert, _ = _compute_residuals(primal_in, rng)
     # averages the perturbed outputs
-    return _tree_mean_across(outputs_pert)
+    return jax.tree.map(lambda x: jnp.mean(x, axis=0), outputs_pert)
 
+  @fun_perturb.defjvp
   def fun_perturb_jvp(
-      tangent: chex.ArrayTree, _: Any, inputs: chex.ArrayTree, rng: chex.PRNGKey
+          rng: chex.PRNGKey, primal_in: chex.ArrayTree, tangent_in: chex.ArrayTree
   ) -> chex.ArrayTree:
     """Computes the jacobian vector product.
 
     Following the method in [Berthet et al. 2020], for a vector `g`, we have
     Jac(fun_perturb)(inputs) * g =
-    - E[fun(inputs + sigma * Z) * <grad log_prob(Z), g>].
+    -1/sigma * E[fun(inputs + sigma * Z) * <grad log_prob(Z), g>].
     This implements a Monte-Carlo estimate
 
     Args:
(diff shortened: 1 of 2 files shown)


rdyro:
This is a really useful contribution, thanks!

Google requires you to sign the CLA first and I should be able to review your changes soon, I'd love to help merge this to also resolve the issues you opened.

adconner:
CLA signed. 

Looks like linter complains that test loss functions are defined in a loop in the test runner. I could replace them with a less readable one line lambda or lift them out of the loop with an extra argument and use functools.partial to bind it later. Neither seems like much of a code improvement to me. Are either of these changes desired?

rdyro:
It shouldn't change anything in your code, but this "loop variable close over" sometimes leads to pretty insidious bugs in python if you're not immediately consuming your closures (which you are).

It's a little annoying, but if you could refactor, it'd be great.

rdyro:
I apologize this linter feedback loop is long, I'm trying to add a local check to pre-commit here: [pull/1313](https://github.com/google-deepmind/optax/pull/1313). For now if you want to check it locally
```
ruff check --select=B023 {DIRECTORY}
```

adconner:
Sorry for all the pings. In hindsight I should have just looked at what the github actions were to see which lints were being run! 

In the end I also had to loosen up the test tolerances as they did not pass with high enough probability.

optax/perturbations/_make_pert.py
```diff
@@ -130,32 +101,31 @@ def _compute_residuals(
       inputs: chex.ArrayTree, rng: chex.PRNGKey
   ) -> tuple[chex.ArrayTree, chex.ArrayTree]:
     # random noise Zs to be added to inputs
-    samples = [
-        optax.tree.random_like(rng_, inputs, sampler=noise.sample)
-        for rng_ in jax.random.split(rng, num_samples)
-    ]
+    samples = jax.vmap(lambda rng_: optax.tree.random_like(rng_, inputs, sampler=noise.sample))(
```
rdyro: Can we use:     
```python
    broadcast_shapes = jax.tree.map(
        lambda x: jax.ShapeDtypeStruct((num_samples,) + x.shape, x.dtype),
        inputs
    )
    samples = optax.tree.random_like(rng, broadcast_shapes, sampler=noise.sample)
```
to save splitting the random key `num_samples` of times.

optax/perturbations/_make_pert_test.py
```diff
@@ -81,40 +87,53 @@ def setUp(self):
...
     ]
 
-  def test_pert_close_array(self):
+    # linear function x -> Ax + b
+    self.linear_A = jnp.array([[0.0, 1.0, 2.0],
+                               [3.0, 4.0, 5.0]])
+    self.linear_b = jnp.array([1.0,-2.0])
+    self.linear_x_test = jnp.array([-1.0, 1.0, 0.0])
+
+  def test_pert_argmax(self):
```
rdyro: Prefer passing the key to the test explicitly and decorating the test like so (the old design was bad):
```python
  @parameterized.product(key=[0, 1, 2])
  def test_pert_argmax(self, key):
    """Test that make_perturbed_fun matches theoretical expression of gradient.

    Applies to the case of an argmax. Includes a test for the gradients and the
    Hessian of a scalar loss.
    """
    
    rng_jax = jax.random.key(key)
```
Applies to all tests (fixing this would be great!)

optax/perturbations/_make_pert_test.py
```diff
@@ -45,17 +47,21 @@ def setUp(self):
     rng = np.random.RandomState(0)
 
     self.rng_jax = jax.random.key(0)
```
rdyro: Could we change this test class to not use a persistent random key, this doesn't follow the practice in the rest of the repo. We didn't catch this before.

Could you delete this line and use `parameterized.product` to pass keys to each test, see other comment.

optax/perturbations/_make_pert.py
carlosgmartin: @rdyro [`tree_batch_shape`](https://optax.readthedocs.io/en/latest/api/utilities.html#optax.tree_utils.tree_batch_shape) can be used for this.

optax/perturbations/_make_pert_test.py
carlosgmartin: @rdyro Shouldn't that argument be called `seed`?

rdyro: That's right, it should be `seed`!

```python
  @parameterized.product(seed=[0, 1, 2])
  def test_pert_argmax(self, seed):
    """Test that make_perturbed_fun matches theoretical expression of gradient.

    Applies to the case of an argmax. Includes a test for the gradients and the
    Hessian of a scalar loss.
    """
    
    key = jax.random.key(seed)
```

Thanks @carlosgmartin !

vroulet:
Great PR! Thank you @adconner ! 
The function is not twice differentiable not because jax would not be able to actually differentiate through the jvp but because it would give the wrong estimator of the second derivative right? (cf https://arxiv.org/abs/1802.05098)
If you agree, we can add this rather than just saying that the function is not twice differentiable. (I haven't actually tried to compute a second derivative, maybe jax would actually throw an error).

vroulet:
> I also considered making the design change discussed in https://github.com/google-deepmind/optax/issues/1214, where it was proposed to use the interface sample_and_log_prob, but I now believe this is impossible. We don't actually care about the log_prob for our samples, but its gradient. The data dependence through the differentiation of log_prob applied to the result of sample requires two separate forward passes, even if we have an efficient function which can compute the log_prob value along with sample in only one pass.

We could (in another PR if you want) consider removing the classes and add clear arguments for sampler and log_prob.

carlosgmartin:
Can we add a test that would've caught the incorrectness of the previous implementation, and checks the correctness of this one?

Perhaps something along the following lines:

```python3
import jax
import optax
from jax import numpy as jnp
from jax import random


def simple_make_perturbed_fun(
    f,
    num_samples=1000,
    sigma=0.1,
    noise=optax.perturbations.Gumbel(),
):
    # Monte Carlo estimate of E[f(x + σ Z)] where Z ~ noise
    # only works for differentiable f
    def g(x, key):
        zs_shape = optax.tree_utils.tree_batch_shape(x, [num_samples])
        zs = optax.tree_utils.tree_random_like(key, zs_shape, noise.sample)
        xs = optax.tree_utils.tree_add_scale(x, sigma, zs)
        ys = jax.vmap(f)(xs)
        ys_mean = jax.tree.map(lambda leaf: jnp.mean(leaf, 0), ys)
        return ys_mean

    return g


def test_make_perturbed_fun(
    num_samples=1000,
    sigma=1,
    noise=optax.perturbations.Normal(),
    atol=1e-8,
    seed=0,
):
    def f(x):
        # a differentiable test function
        x0, x1, x2 = x
        y0 = x0 * x1**2 * jnp.sin(x2 * x1 - x0) + jnp.cos(x0**3)
        y1 = x2 + x0 * jnp.sin(x1)
        return jnp.stack([y0, y1])

    f1 = optax.perturbations.make_perturbed_fun(
        f,
        num_samples=num_samples,
        sigma=sigma,
        noise=noise,
    )
    f2 = simple_make_perturbed_fun(
        f,
        num_samples=num_samples,
        sigma=sigma,
        noise=noise,
    )

    x = jnp.array([0.3, 0.4, 0.5])
    key = random.key(seed)
    j1 = jax.jacobian(f1)(x, key)
    j2 = jax.jacobian(f2)(x, key)
    print(j1)
    print(j2)
    assert jnp.allclose(j1, j2, atol=atol)


if __name__ == "__main__":
    test_make_perturbed_fun()
```

rdyro:
What is this test testing @carlosgmartin ? Is a check against a simpler reference implementation?

carlosgmartin:
@rdyro Yes (but the simpler version only works for differentiable functions; so we run both on a differentiable function and compare).

rdyro:
> @rdyro Yes (but the simpler version only works for differentiable functions; so we run both on a differentiable function and compare).

Let's add it in an immediate follow up PR since this one has quite a few moving parts already. @adconner would you be able to contribute the additional test in a follow up PR? @carlosgmartin or if you wanted to do it?

adconner:
> Great PR! Thank you @adconner ! The function is not twice differentiable not because jax would not be able to actually differentiate through the jvp but because it would give the wrong estimator of the second derivative right? (cf https://arxiv.org/abs/1802.05098) If you agree, we can add this rather than just saying that the function is not twice differentiable. (I haven't actually tried to compute a second derivative, maybe jax would actually throw an error).

Differentiating through the jvp would just differentiate through the Monte-Carlo estimation, rather than express the result as a new Monte-Carlo estimation. Put more directly, differentiating through the perturbed jvp differentiates the original function, which we are trying to avoid. 

I think it may be possible to make this work though, one just needs to express the jvp as another perturbed function. It looks like maybe one can take f perturbed with respect to a modified distribution. I need to think a bit more about it.

This suggests a test should be added testing that the derivative of f is not used in the derivative of f_perturb (by for example wrapping f with a custom jvp which errors out and calling the jvp of f_peturb).

> We could (in another PR if you want) consider removing the classes and add clear arguments for sampler and log_prob.

I'm happy with whatever you want to do here. This change is relatively orthogonal to the present one if we agree that we still want to use sampler and log_prob separately. Although the current interface is pretty ergonomic with the distributions in distrax, eg, as distributions are specified as classes providing at least the functions `sample` and `log_prob`

> Let's add it in an immediate follow up PR since this one has quite a few moving parts already. @adconner would you be able to contribute the additional test in a follow up PR? @carlosgmartin or if you wanted to do it?

I can do it if you like.

rdyro:
I appreciate the comments! To summarize:

@adconner could you:
* add your explanation to the docstring for why the higher order derivatives are non-informative / not supported in the current implementation? let's not handle the 2nd order derivative behavior for now (in this PR), but plan to either support it or catch it and error
* include @carlosgmartin 's additional test
* defer "removing the classes and add clear arguments for sampler and log_prob" to another PR

rdyro (approved):
Great PR, thanks so much!

vroulet:
@adconner 

> I think it may be possible to make this work though, one just needs to express the jvp as another perturbed function. It looks like maybe one can take f perturbed with respect to a modified distribution. I need to think a bit more about it.

This is what the dice paper is about https://arxiv.org/abs/1802.05098

There is some further literature on "automatic stochastic estimates of derivatives" like this [paper](https://proceedings.neurips.cc/paper_files/paper/2021/file/3dfe2f633108d604df160cd1b01710db-Paper.pdf) or even [this one](https://arxiv.org/pdf/2212.06386).

Essentially the make_perturb implements a score function estimator (aka REINFORCE). To enable higher automatic differentiation one should use the "magic_box" operator of the dice paper I think.

PR #1319, opened 2025-06-07
Clean up optax.perturbations and add optional variance-reducing baseline to make_perturbed_fun.

carlosgmartin:
Follow-up to https://github.com/google-deepmind/optax/pull/1310 and https://github.com/google-deepmind/optax/pull/1318 performing some cleanup of the [perturbations](https://optax.readthedocs.io/en/latest/api/perturbations.html) module.

Also adds an optional variance-reducing baseline to [make_perturbed_fun](https://optax.readthedocs.io/en/latest/api/perturbations.html#optax.perturbations.make_perturbed_fun) (see the Variance Reduction section of the [DiCE paper](https://arxiv.org/pdf/1802.05098#page=6)). Here's an example that illustrates its effect:

```python3
import jax
from jax import numpy as jnp
from jax import random
from optax.perturbations import make_perturbed_fun


def f(x):
    return x @ x + ((1 - x) * (2 * x**2)).sum()


x = jnp.array([0.1, 0.2, 0.3])
print(f"true: {jax.grad(f)(x)}")

key = random.key(0)
keys = random.split(key, 100_000)

for kwargs in [
    dict(num_samples=1, sigma=1e-3, use_baseline=True),
    dict(num_samples=1, sigma=1e-3, use_baseline=False),
    dict(num_samples=2, sigma=1e-3, use_baseline=False),
]:
    fp = make_perturbed_fun(f, **kwargs)
    grad_estimates = jax.vmap(jax.grad(fp, 1), [0, None])(keys, x)
    print(kwargs)
    print(f"mean: {grad_estimates.mean(0)}")
    print(f"std: {grad_estimates.std(0)}")
```
```
true: [0.53999996 0.96000004 1.26      ]
{'num_samples': 1, 'sigma': 0.001, 'use_baseline': True}
mean: [0.5373496  0.96939325 1.2595664 ]
std: [2.4107156 2.2428784 2.1961608]
{'num_samples': 1, 'sigma': 0.001, 'use_baseline': False}
mean: [-1.9298862   0.22652858 -0.26049805]
std: [350.75748 349.44354 350.60086]
{'num_samples': 2, 'sigma': 0.001, 'use_baseline': False}
mean: [-1.0758405  0.9094837  1.5837727]
std: [247.80414 246.84732 247.14584]
```

### DIFF
--- a/examples/perturbations.ipynb
+++ b/examples/perturbations.ipynb
@@ -1,924 +1,925 @@
 {
-  "cells": [
-    {
-      "cell_type": "markdown",
-      "metadata": {},
-      "source": [
-        "# Perturbed optimizers\n",
-        "\n",
-        "\n",
-        "[![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.sandbox.google.com/github/google-deepmind/optax/blob/main/examples/perturbations.ipynb)"
-      ]
-    },
-    {
-      "cell_type": "markdown",
-      "metadata": {
-        "id": "l98gJVtFFJPB"
-      },
-      "source": [
-        "We review in this notebook a universal method to transform any function $f$ mapping a pytree to another pytree to a differentiable approximation $f_\\varepsilon$, using pertutbations following the method of [Berthet et al. (2020)](https://arxiv.org/abs/2002.08676).\n",
-        "\n",
-        "For a random $Z$ drawn from a distribution with continuous positive distribution $\\mu$ and a function $f: X \\to Y$, its perturbed approximation defined for any $x \\in X$ by\n",
-        "\n",
-        "$$f_\\varepsilon(x) = \\mathbf{E}[f (x + \\varepsilon Z )]\\, .$$\n",
-        "\n",
-        "We illustrate here on some examples, including the case of an optimizer function $y^*$ over $C$ defined for any cost $\\theta \\in \\mathbb{R}^d$ by\n",
-        "\n",
-        "$$y^*(\\theta) = \\mathop{\\mathrm{arg\\,max}}_{y \\in C} \\langle y, \\theta \\rangle\\, .$$\n",
-        "\n",
-        "In this case, the perturbed optimizer is given by\n",
-        "\n",
-        "$$y_\\varepsilon^*(\\theta) = \\mathbf{E}[\\mathop{\\mathrm{arg\\,max}}_{y\\in C} \\langle y, \\theta + \\varepsilon Z \\rangle]\\, .$$"
-      ]
-    },
-    {
-      "cell_type": "code",
-      "execution_count": 1,
-      "metadata": {
-        "id": "S6tLyyy9VCEw"
-      },
-      "outputs": [],
-      "source": [
-        "import jax\n",
-        "import jax.numpy as jnp\n",
-        "import operator\n",
-        "from jax import tree_util as jtu\n",
-        "\n",
-        "import optax.tree\n",
-        "from optax import perturbations"
-      ]
-    },
-    {
-      "cell_type": "markdown",
-      "metadata": {
-        "id": "EmYn_jNUFfw2"
-      },
-      "source": [
-        "# Argmax one-hot"
-      ]
-    },
-    {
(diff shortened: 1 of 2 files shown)


rdyro:
Thanks! The inclusion of the variance reduction technique seems like a good idea, judging by the paper. Since there appears to be no downside to using it, I'd suggest:
* remove the keyword arguments for use_baseline and simply use it always time
* avoid rewriting the entire implementation, I think the baseline diff should just be 1 line

carlosgmartin:
@rdyro I left it optional because I thought someone might want to use the single-point estimator.

Regarding the diff (for the second commit), I refactored the implementation so that the baseline is evaluated only once outside the vmap. (The JAX compiler should optimize this, but in the absence of compilation, I thought it might make a difference.)

rdyro:
> @rdyro I left it optional because I thought someone might want to use the single-point estimator.
> 
> Regarding the diff (for the second commit), I refactored the implementation so that the baseline is evaluated only once outside the vmap. (The JAX compiler should optimize this, but in the absence of compilation, I thought it might make a difference.)

Oh, I see!

Perhaps this diff could work instead?
```
--- a/optax/perturbations/_make_pert.py
+++ b/optax/perturbations/_make_pert.py
@@ -132,19 +132,23 @@ def make_perturbed_fun(
     * :doc:`../_collections/examples/perturbations` example.
   """

-  def stoch_estimator(key: chex.PRNGKey, x: chex.ArrayTree) -> chex.ArrayTree:
+  def stoch_estimator(key: chex.PRNGKey, x: chex.ArrayTree,
+                      baseline: chex.ArrayTree) -> chex.ArrayTree:
     sample = optax.tree.random_like(key, x, sampler=noise.sample)
     shifted_sample = jax.tree.map(lambda x, z: x + sigma * z, x, sample)
     shifted_sample = jax.lax.stop_gradient(shifted_sample)
     sample = jax.tree.map(lambda x, y: (y - x) / sigma, x, shifted_sample)

     log_prob_sample = optax.tree.sum(jax.tree.map(noise.log_prob, sample))
-    out = optax.tree.scale(_magicbox(log_prob_sample), fun(shifted_sample))
+    box = _magicbox(log_prob_sample)
+    out = optax.tree.scale(box, fun(shifted_sample))
+    out = optax.tree.add_scale(out, 1 - box, baseline)
     return out

   def mc_estimator(key: chex.PRNGKey, x: chex.ArrayTree) -> chex.ArrayTree:
-    out = jax.vmap(stoch_estimator, in_axes=(0, None), out_axes=0)(
-        jax.random.split(key, num_samples), x
+    baseline = fun(jax.lax.stop_gradient(x))
+    out = jax.vmap(stoch_estimator, in_axes=(0, None, None), out_axes=0)(
+        jax.random.split(key, num_samples), x, baseline
     )
     return jax.tree.map(lambda x: jnp.mean(x, axis=0), out)
```

I also realized the `optax/examples/perturbations.ipynb` is not up to date because the `rng` (which should be called `key`) now comes first in the function call.

Would you be able to add the two changes to this PR @carlosgmartin ?

carlosgmartin:
@rdyro I've fixed the notebook.

Is there any particular reason you want to keep the old code over the (IMO much cleaner) new code?

rdyro:
Thank you!

The current version's already clean and easy to follow; the rewrite's unnecessary. If you can apply the patch from the comment above to add baseline variance reduction (on by default, no need for an additional keyword argument to `make_perturbed_fun`), I'll approve.

carlosgmartin:
@rdyro Done. Though I still think the baseline should be optional (I agree it should be on by default).

rdyro (approved):
Thanks!

rdyro:
> @rdyro Done. Though I still think the baseline should be optional (I agree it should be on by default).

That's fair, can you bring back the baseline conditional, please?

carlosgmartin:
@rdyro Done.

vroulet:
Thanks @carlosgmartin ! The use_baseline argument is not used now.
No worries. I'll take it from here and patch it in the internal codebase. Thank you again !

vroulet:
Merged

PR #1321, opened 2025-06-08
Add tree utility functions.

carlosgmartin:
Adds the following [tree utility functions](https://optax.readthedocs.io/en/latest/api/utilities.html):

- tree_size
- tree_neg
- tree_min

### DIFF
--- a/optax/tree_utils/_tree_math.py
+++ b/optax/tree_utils/_tree_math.py
@@ -183,6 +183,32 @@ def tree_max(tree: Any) -> chex.Numeric:
   return jax.tree.reduce(jnp.maximum, maxes, initializer=jnp.array(-jnp.inf))
 
 
+def tree_min(tree: Any) -> chex.Numeric:
+  """Compute the min of all the elements in a pytree.
+
+  Args:
+    tree: pytree.
+
+  Returns:
+    a scalar value.
+  """
+  mins = jax.tree.map(jnp.min, tree)
+  # initializer=jnp.inf should work but pytype wants a jax.Array.
+  return jax.tree.reduce(jnp.minimum, mins, initializer=jnp.array(jnp.inf))
+
+
+def tree_size(tree: Any) -> int:
+  r"""Total size of a pytree.
+
+  Args:
+    tree: pytree
+
+  Returns:
+    the total size of the pytree.
+  """
+  return sum(jnp.size(leaf) for leaf in jax.tree.leaves(tree))
+
+
 def tree_conj(tree: Any) -> Any:
   """Compute the conjugate of a pytree.
 
--- a/docs/api/utilities.rst
+++ b/docs/api/utilities.rst
@@ -110,6 +110,8 @@ Tree
     tree_norm
     tree_map_params
     tree_max
+    tree_min
+    tree_size
     tree_mul
     tree_ones_like
     tree_random_like
--- a/optax/tree/__init__.py
+++ b/optax/tree/__init__.py
@@ -40,6 +40,8 @@ conj = _tree_math.tree_conj
 div = _tree_math.tree_div
 full_like = _tree_math.tree_full_like
 max = _tree_math.tree_max  # pylint: disable=redefined-builtin
+min = _tree_math.tree_min  # pylint: disable=redefined-builtin
+size = _tree_math.tree_size
 mul = _tree_math.tree_mul
 norm = _tree_math.tree_norm
 ones_like = _tree_math.tree_ones_like
--- a/optax/tree_utils/__init__.py
+++ b/optax/tree_utils/__init__.py
@@ -39,6 +39,8 @@ from optax.tree_utils._tree_math import tree_conj
 from optax.tree_utils._tree_math import tree_div
 from optax.tree_utils._tree_math import tree_full_like
 from optax.tree_utils._tree_math import tree_max
+from optax.tree_utils._tree_math import tree_min
+from optax.tree_utils._tree_math import tree_size
 from optax.tree_utils._tree_math import tree_mul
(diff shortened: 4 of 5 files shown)


optax/tree_utils/_tree_math.py
```diff
@@ -24,6 +24,30 @@
 from optax._src import numerics
 
 
+def tree_size(tree: Any) -> int:
```
rdyro: with size you need to be careful handling int4 types, it might need to be necessary to perform a lookup to something like [https://github.com/jax-ml/jax/blob/c2a569096c1a28c2b9348737172aee381ad0268e/jax/_src/basearray.pyi#L260](https://github.com/jax-ml/jax/blob/c2a569096c1a28c2b9348737172aee381ad0268e/jax/_src/basearray.pyi#L260)

Also, are we attempting to get `size` or `byte_size`? What's the application for this you had in mind?

optax/tree_utils/_tree_math.py
```diff
@@ -24,6 +24,30 @@
...
+  Args:
+    tree: pytree
+
+  Returns:
+    the total size of the pytree.
+  """
+  return sum(leaf.size for leaf in jax.tree.leaves(tree))
+
+
+def tree_neg(tree: Any) -> Any:
```
rdyro: optax tree functions are generally useful when they add a layer on top of jax.tree.map, since there's no reduction here, I think we should allow this usage to require jax.tree.map

optax/tree_utils/_tree_math.py
```diff
@@ -183,6 +207,20 @@ def tree_max(tree: Any) -> chex.Numeric:
   return jax.tree.reduce(jnp.maximum, maxes, initializer=jnp.array(-jnp.inf))
 
 
+def tree_min(tree: Any) -> chex.Numeric:
```
rdyro, line 186: This one looks great!

optax/tree_utils/_tree_math.py
vroulet: I fully agree with Robert. 
I know tree_scale, tree_add etc... are one lines but in my opinion we should consider removing them in favor of jax.tree.map equivalent.

optax/tree_utils/_tree_math.py
carlosgmartin: The intent is an analogue of [jax.numpy.size](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.size.html) for trees, which counts the number of elements (not bytes).

This is useful, for example, to count the number of parameters of a model. For example, Haiku has a [tree_size](https://dm-haiku.readthedocs.io/en/latest/api.html#haiku.data_structures.tree_size) function. This quantity in turn is sometimes used in optimization algorithms to set hyperparameters.

Haiku also has a [tree_bytes](https://dm-haiku.readthedocs.io/en/latest/api.html#haiku.data_structures.tree_bytes) function. I'd also be happy to add that, if you'd like.

optax/tree_utils/_tree_math.py
carlosgmartin: @rdyro @vroulet To make sure I understand correctly, are you saying I should do the following:

1. Remove `tree_neg` from `docs/api/utilities.rst`.
2. Prepend an underscore to `tree_neg` in `optax/tree_utils/_tree_math.py`.
3. Add `neg = _tree_math._tree_neg` to `optax/tree/__init__.py`.
4. Add it to documentation somewhere. (I'm not sure where the functions under `optax.tree` are supposed to be documented.)

Or are you suggesting excluding `tree_neg` from this PR and removing tree arithmetic entirely in a future PR?

Personally, I would disagree with the latter. When handling a lot of tree arithmetic, compact code like `tu.add(a, b)` is easier to read, mentally parse, and reason about than code like `jax.tree.map(lambda a, b: a + b, a, b)`. This also leads to fewer bugs (less code = fewer places for bugs to hide, generally speaking).

To quote from Iverson's famous essay [_Notation as a Tool of Thought_](https://www.jsoftware.com/papers/tot.htm):

>  The quantity of meaning compressed into small space by algebraic signs, is another circumstance that facilitates the reasonings we are accustomed to carry on by their aid. — Charles Babbage

>  By relieving the brain of all unnecessary work, a good notation sets it free to concentrate on more advanced problems, and in effect increases the mental power of the race. — A.N. Whitehead

and

> Civilization advances by extending the number of important operations which we can perform without thinking about them. — A.N. Whitehead

This is my opinion. 🙂

vroulet: I like the quotes ^^
To me algebraic signs are more universal than words so `jax.tree.map(lambda x: -x, tree)` is more clear than `tree_neg(tree)`. But it may be a question of taste. 
Ideally we would use https://github.com/google/tree-math.
Happy to merge the tree_size, tree_min for now, and we can keep the discussion about tree_neg and other tree arithmetics for later

carlosgmartin: @vroulet I've removed tree_neg for now. If you want me to also add a [tree_bytes](https://dm-haiku.readthedocs.io/en/latest/api.html#haiku.data_structures.tree_bytes) function, as suggested [here](https://github.com/google-deepmind/optax/pull/1321#discussion_r2134821563), let me know.

vroulet: Thanks! tree_bytes would be quite nice too actually. You can do it in another PR.

vroulet (commented):
Could you add some basic tests?
Also for tree_size, it may be safer/more generic to use jax.numpy.size.
That way I think your tree_size would even work on simple tuples like `(1, 2, 3)` (rather than raising an attribute error I think)

carlosgmartin:
@vroulet Good idea. Done.

Issue #1335, opened 2025-06-12
Different gradient behaviors between `softmax_cross_entropy_with_integer_labels` and `softmax_cross_entropy`


AJINGOUP:
### Issue Description
I'm trying to handle dynamic batch sizes by using the `where` parameter in optax's cross-entropy functions to avoid JIT recompilation issues. While testing masked implementations, I noticed different gradient behaviors between `softmax_cross_entropy_with_integer_labels` and `softmax_cross_entropy` that seem inconsistent.

### Minimal Reproduction
```python
import jax
import jax.numpy as jnp
import optax

num_samples = 5
num_classes = 5

# Integer labels version (problematic)
def fn1(logits, labels, where):
    loss = optax.softmax_cross_entropy_with_integer_labels(
        logits=logits, 
        labels=labels, 
        where=jnp.expand_dims(where, -1)  # Expand for broadcasting
    )
    return jnp.mean(loss, where=where)

# One-hot version (works as expected)
def fn2(logits, labels, where):
    loss = optax.softmax_cross_entropy(
        logits=logits,
        labels=jax.nn.one_hot(labels, num_classes),
        where=jnp.expand_dims(where, -1)  # Same mask expansion
    )
    return jnp.mean(loss, where=where)

# Test data
logits = jax.random.normal(jax.random.PRNGKey(0), (num_samples, num_classes))
labels = jax.random.randint(jax.random.PRNGKey(1), (num_samples,), 0, num_classes)
masks = jax.random.bernoulli(jax.random.PRNGKey(2), p=0.5, shape=(num_samples,))
print("Masks:", masks)  # e.g. [False False  True False  True]

# Check fn1 (integer labels)
loss, grad = jax.value_and_grad(fn1)(logits, labels, masks)
print("fn1 grad:", grad)  # NaN in masked positions

# Check fn2 (one-hot)
loss, grad = jax.value_and_grad(fn2)(logits, labels, masks)
print("fn2 grad:", grad)  # Proper zeros in masked positions
```

### Actual Output
```
Masks: [False False  True False  True]

fn1 grad: 
[[        nan         nan         nan         nan         nan]  # ← masked
 [        nan         nan         nan         nan         nan]  # ← masked
 [-0.20860437  0.0046646   0.04716831  0.03858821  0.11818324]  # unmasked
 [        nan         nan         nan         nan         nan]  # ← masked
 [ 0.00761222  0.27513114  0.0512284  -0.38969693  0.05572514]] # unmasked

[...]

PR #912, opened 2024-04-08
Add axis and where arguments to loss functions.

carlosgmartin:
#902

PR #1336, opened 2025-06-12
Fix for https://github.com/google-deepmind/optax/issues/1335

rdyro:


optax/losses/_classification.py
```diff
@@ -251,7 +251,7 @@ def softmax_cross_entropy(
     labels: One-hot encoded labels, with shape `[batch_size, num_classes]`. Each
       row represents the true class distribution for a single example.
     axis: Axis or axes along which to compute.
-    where: Elements to include in the computation.
+    where: Elements to include in the computation of shape ``[batch_size]``.
```
vroulet: Or shape ``[batch_size, num_classes]``? You let that still be viable (and said it in the comments of the issue).
Same comment for the other docstring

optax/losses/_classification.py
```diff
@@ -388,14 +392,10 @@ def softmax_cross_entropy_with_integer_labels(
   # This is like jnp.take_along_axis(jax.nn.log_softmax(...), ...) except that
   # we avoid subtracting the normalizer from all values, just from the values
   # for the correct labels.
-  logits_max = jnp.max(
-      logits, axis, keepdims=True, where=where, initial=-jnp.inf
-  )
-  logits -= jax.lax.stop_gradient(logits_max)
   label_logits = jnp.take_along_axis(
```
vroulet, line 397: You need to have the where used also for label_logits

vroulet:
Could you add a little test? #912 seems to have just skipped the test that would have helped us avoid this bug. (see lack of test_mask in that PR)

optax/losses/_classification.py
rdyro: Added!

optax/losses/_classification.py
```diff
@@ -251,7 +251,8 @@ def softmax_cross_entropy(
     labels: One-hot encoded labels, with shape `[batch_size, num_classes]`. Each
       row represents the true class distribution for a single example.
     axis: Axis or axes along which to compute.
-    where: Elements to include in the computation.
+    where: Elements to include in the computation of shape ``[batch_size]`` or
+      logits.shape.
```
emilyfertig, line 255: Optional: `[batch_size, num_classes]` is clearer IMO than `logits.shape`.

emilyfertig (approved):

optax/losses/_classification.py
rdyro: We're technically allowing `[num_classes, batch_size]` and any other combination if `axis` is correctly specified I think

emilyfertig: Ah ok -- the `logits` and `labels` docstrings specify `[batch_size, num_classes]`.

Issue #906, opened 2024-04-05
unitwise_norm fails for 3D convolutions


froody:
`unitwise_norm`, used by `adaptive_grad_clip`, only supports a few values of `ndim`, and raises ValueError when applied to a conv3d kernel since `ndim=5` (HWDIO). Would it be acceptable to add an optional `axis` kwarg to `adaptive_grad_clip` and `unitwise_norm`? This would allow specifying the reduction axes at the callsite instead of baking every possible combination into the implementation of `unitwise_norm`.

I'm happy to submit a PR

vroulet:
Hello @froody,

Good catch. The behavior of `adaptive_grad_clip` hides indeed some logic that could mislead users indeed. 
If you are willing to do a pr to let this function handle `ndim=5` that would be great. I don't know exactly how you can add an `axis` and keep the current default behavior, I let you try and see :)

Thank you !

aymuos15:
Hi @vroulet 

I tried having a go at this below. Do you think this is sufficient to kick off a PR or am I missing something? Thanks :)

```diff
diff --git a/optax/transforms/_clipping.py b/optax/transforms/_clipping.py
index eebf6aa..cbb9156 100644
--- a/optax/transforms/_clipping.py
+++ b/optax/transforms/_clipping.py
@@ -18,6 +18,8 @@ Note that complex numbers are also supported, see
 https://gist.github.com/wdphy16/118aef6fb5f82c49790d7678cf87da29
 """
 
+from typing import Optional, Union
+
 import chex
 import jax
 import jax.numpy as jnp
@@ -241,9 +243,21 @@ def per_example_layer_norm_clip(
   )
 
 
-def unitwise_norm(x: chex.Array) -> chex.Array:
-  """Computes norms of each output unit separately."""
-  if jnp.squeeze(x).ndim <= 1:  # Scalars and vectors
+def unitwise_norm(x: chex.Array, axis: Optional[Union[int, tuple[int, ...]]] = None) -> chex.Array:
+  """Computes norms of each output unit separately.
+
+  Args:
+    x: Input array to compute norms for.
+    axis: Axis or axes along which to compute the norm. If None, uses default
+      behavior based on input dimensions.
+
+  Returns:
+    Array of the same shape as x with unit-wise norms.
+  """
+  if axis is not None:
+    # Use provided axes for reduction
+    squared_norm = jnp.sum(numerics.abs_sq(x), axis=axis, keepdims=True)
[...]

PR #1340, opened 2025-06-17
Enable adaptive gradient clipping for high-dimensional tensors

aymuos15:
Added an optional `axis` argument to `adaptive_grad_clip` / `unitwise_norm` for full 5-D-kernel support

Adaptive Gradient Clipping breaks on parameters whose **rank > 4** (e.g. Conv3D kernels **H × W × D × I × O**).  This is because `unitwise_norm` only supports ranks ≤ 4, causing a `ValueError`. 

Issue which first mentions this: https://github.com/google-deepmind/optax/issues/906

### DIFF
--- a/optax/transforms/_clipping.py
+++ b/optax/transforms/_clipping.py
@@ -18,6 +18,8 @@ Note that complex numbers are also supported, see
 https://gist.github.com/wdphy16/118aef6fb5f82c49790d7678cf87da29
 """
 
+from typing import Optional, Union
+
 import chex
 import jax
 import jax.numpy as jnp
@@ -148,8 +150,8 @@ def per_example_global_norm_clip(
 
   if not _check_arrays_have_batch_dim(grads):
     raise ValueError(
-        'Unlike other transforms, `per_example_global_norm_clip` expects'
-        ' `grads` to have a batch dimension in the 0th axis.'
+        "Unlike other transforms, `per_example_global_norm_clip` expects"
+        " `grads` to have a batch dimension in the 0th axis."
     )
 
   global_grad_norms = jax.vmap(linear_algebra.global_norm)(grads)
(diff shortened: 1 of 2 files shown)


selamw1 (changes requested):
LGTM with minor polishing.

Proposing an update to the docstring to improve clarity and align it with the Optax style. The proposed version clarifies key details, such as the default axis inference, so users won’t need to read the implementation to understand the function’s behavior.


```
"""Computes the L2 norm of each unit separately.

A "unit" is a slice of `x` along the dimensions specified by `axis`. If `axis`
is ``None``, the reduction axes are inferred from `x`'s rank based on
common layer conventions:
  - Rank-1: The whole vector.
  - Rank-2: Axis 0 (e.g., for linear layers).
  - Rank-3 or 4: Axes (0, 1, 2) (e.g., for multi-head attention or convolutions).

Args:
  x: Input array for which to compute unit-wise norms.
  axis: Axis or axes to normalize over. If ``None``, defaults are
    inferred from the input's rank.

Returns:
  Array with the same shape as `x`, where each unit is replaced by its
  L2 norm.
"""
```

aymuos15:
Thank you very much for the suggestion. Will keep that in mind moving forward.

rajasekharporeddy:
It seems some lines are too long which fail the internal Lint tests. Please use maximum 80 characters per line.

aymuos15:
Is there any way to mock the linter locally? Thanks!

aymuos15:
I apologise for so many commits. I am not sure why the cla is failing. Could someone please help me with this?

rdyro:
You can try running `pre-commit run -a` to run the linter locally, let me know if that doesn't work!

aymuos15:
Thanks a lot for the suggestion @rdyro 

I did run the pre-commit before (even mentioned it in one of the commits) but it seemed that I was missing something. Fixed that now. 

The cla issues seems to be that I have pushed commits from two separate email ids associated to the same account. And I have not been able get the cla to access that. Even though I have signed and rescanned multiple times. Have you faced anything like this before? Thanks again.

Could I potentially remove that commit?

rdyro:
Nice, thanks!

> The cla issues seems to be that I have pushed commits from two separate email ids associated to the same account. And I have not been able get the cla to access that. Even though I have signed and rescanned multiple times. Have you faced anything like this before? Thanks again.

I'd recommend squashing your commits into 1, it makes for a nicer git history and will probably remove the CLA issue, it should be fairly simple:
```bash
$ git rebase -i main  # then replace all but the first "pick" with "s"
```

Also it seems you reformatted both files. You can either confirm manually that your changes are the only thing in the actual diff (`git diff main`) or you could use alternatively use a formatter for the two files:
```
$ pyink --pyink-indentation=2 --line-length=80 --pyink-use-majority-quotes _clipping.py _clipping_test.py
```

aymuos15:
@rdyro (and the rest)

Thank you so much for taking the time and guiding me through the entire process. I have learnt a lot and will make sure make the future PRs are much smoother.

I think there are still some unnecessary changes regarding quotes, please let me know if i have to revert those.

PR #1341, opened 2025-06-18
Add tree_bits function.

carlosgmartin:
Add a `tree_bytes` function to the [tree utilities](https://optax.readthedocs.io/en/latest/api/utilities.html#tree), analogous to [Haiku's](https://dm-haiku.readthedocs.io/en/latest/api.html#haiku.data_structures.tree_bytes).

For context, see https://github.com/google-deepmind/optax/pull/1321#discussion_r2155170615.

### DIFF
--- a/optax/tree_utils/_tree_math_test.py
+++ b/optax/tree_utils/_tree_math_test.py
@@ -207,6 +207,39 @@ class TreeUtilsTest(parameterized.TestCase):
     assert tu.tree_allclose(1, 1 + 1e-7)
     assert not tu.tree_allclose(1, 2)
 
+  @parameterized.product(
+    size=[1, 10, 100, 1000],
+    dtype=[
+        jnp.int4,
+        jnp.int8,
+        jnp.int16,
+        jnp.int32,
+        jnp.uint4,
+        jnp.uint8,
+        jnp.uint16,
+        jnp.uint32,
+        jnp.float16,
+        jnp.float32,
+        jnp.bfloat16,
+    ],
+  )
+  def test_tree_bits(self, size, dtype):
+    tree = jnp.zeros(size, dtype)
+    bits = {
+      jnp.int4: 4,
+      jnp.int8: 8,
+      jnp.int16: 16,
+      jnp.int32: 32,
+      jnp.uint4: 4,
+      jnp.uint8: 8,
+      jnp.uint16: 16,
+      jnp.uint32: 32,
+      jnp.float16: 16,
+      jnp.float32: 32,
+      jnp.bfloat16: 16,
+    }[dtype]
+    assert tu.tree_bits(tree) == bits * size
+
   def test_tree_conj(self):
     expected = jnp.conj(self.array_a)
     got = tu.tree_conj(self.array_a)
--- a/docs/api/utilities.rst
+++ b/docs/api/utilities.rst
@@ -120,6 +120,7 @@ Tree
     tree_scale
     tree_set
     tree_size
+    tree_bits
     tree_sub
     tree_sum
     tree_vdot
@@ -229,6 +230,10 @@ Set values in a tree
 Tree size
 ~~~~~~~~~
 .. autofunction:: tree_size
+
+Tree bits
+~~~~~~~~~
+.. autofunction:: tree_bits
 
 Tree subtract
 ~~~~~~~~~~~~~
--- a/optax/tree/__init__.py
+++ b/optax/tree/__init__.py
@@ -48,6 +48,7 @@ ones_like = _tree_math.tree_ones_like
 real = _tree_math.tree_real
 scale = _tree_math.tree_scale
 size = _tree_math.tree_size
+bits = _tree_math.tree_bits
 sub = _tree_math.tree_sub
 sum = _tree_math.tree_sum  # pylint: disable=redefined-builtin
 update_infinity_moment = _tree_math.tree_update_infinity_moment
--- a/optax/tree_utils/__init__.py
+++ b/optax/tree_utils/__init__.py
@@ -47,6 +47,7 @@ from optax.tree_utils._tree_math import tree_ones_like
 from optax.tree_utils._tree_math import tree_real
 from optax.tree_utils._tree_math import tree_scale
 from optax.tree_utils._tree_math import tree_size
+from optax.tree_utils._tree_math import tree_bits
 from optax.tree_utils._tree_math import tree_sub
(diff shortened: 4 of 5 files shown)


rdyro:
I like the idea!

Can we check for the presence of `jnp.ones(10).on_device_size_in_bytes()` on the object, for `int4` arrays `itemsize=1`, but on device size is half a byte per element.

carlosgmartin:
@rdyro Are you suggesting adding a flag `on_device` to `tree_bytes` that replaces `jnp.asarray(leaf).nbytes` with `jnp.asarray(leaf).on_device_size_in_bytes()`?

rdyro:
I was thinking of using Python's `hasattr`, however, this will fail for tracers that are backend independent and will report they don't have `on_device_size_in_bytes`.

I'm not sure about the name of this function now, `size * itemsize` does not correctly report the byte size and `on_device_size_in_bytes` is not available on tracers.

carlosgmartin:
I'm now using [`nbytes`](https://docs.jax.dev/en/latest/_autosummary/jax.Array.nbytes.html). That seems like the simplest, most straightforward approach. Let's go with that for now.

In the meantime, I left a question about `on_device_size_in_bytes` [here](https://github.com/jax-ml/jax/discussions/29638).

rdyro:
Hmmm, `nbytes` appears to just use `size * itemsize`.

Can you test this on the following case:
```python
fn = lambda: tree_bytes(jnp.ones((1024,), dtype=jnp.int32))
fn()
jax.jit(fn)()
```
It should report 512 in both cases

carlosgmartin:
@rdyro Shouldn't it be 4096?

An int32 has 32 bits / 8 bits per byte = 4 bytes. And 1024 * 4 = 4096.

I get 4096 in both cases.

rdyro:
Should have been `int4`, that's the whole problem here :/

```python
fn = lambda: tree_bytes(jnp.ones((1024,), dtype=jnp.int4))
fn()
jax.jit(fn)()
```

carlosgmartin:
For `int4`, I get 1024 in both cases.

optax/tree_utils/_tree_math_test.py
```diff
@@ -172,6 +172,15 @@ def test_tree_size(self, key):
     got = tu.tree_size(tree)
     np.testing.assert_allclose(expected, got)
 
+  def test_tree_bytes(self):
+    fn = lambda: tu.tree_bytes(jnp.ones((1024,), dtype=jnp.int32))
+    assert fn() == 4096
+    assert jax.jit(fn)() == 4096
+
+    fn = lambda: tu.tree_bytes(jnp.ones((1024,), dtype=jnp.int4))
+    assert fn() == 1024
```
rdyro: This should be 512, right?

`1024 * 0.5 byte per element = 512`

rdyro:
> For `int4`, I get 1024 in both cases.

`.nbytes` and `.size * .itemsize` both give the **wrong** number. Currently the only way to get the true byte size for all dtypes is to call `.on_device_size_in_bytes()` in eager mode.

optax/tree_utils/_tree_math_test.py
carlosgmartin: I see what you mean.

That seems like an upstream issue with JAX's [nbytes](https://docs.jax.dev/en/latest/_autosummary/jax.Array.nbytes.html) function.

Should I open an issue for it in JAX's repo?

rdyro: It’s a thorny question. You’re trying to write a function that works for both a tracer and an actual array. Since the size on device is a platform question, the tracer most likely cannot know what its size in bytes is.

The alternative would be to make the function work on concrete arrays only, but I think then it shouldn’t be in `optax.tree.`

carlosgmartin:
This still works for the most common dtypes, so let's go with the `nbytes`-based implementation for now.

I added a warning to the function's docstring. Feel free to reword this warning.

rdyro:
Let’s wait until we have a solution on the JAX side, I’ll keep the PR open for now.

carlosgmartin:
Will you be opening an issue for that?

rdyro:
I'll follow up on this internally. For the JAX issue you opened, can you explicitly ask about the use case of getting int4 byte size under jit?

carlosgmartin:
@rdyro In your opinion, what would be the ideal output of `jnp.ones(1, dtype=jnp.int4).nbytes`, or whatever the equivalent to `nbytes` is? Seems ugly to mix `float` and `int` values.

Perhaps this suggests that, more generally, we ought to be counting in bits rather than bytes?

rdyro:
> @rdyro In your opinion, what would be the ideal output of `jnp.ones(1, dtype=jnp.int4).nbytes`, or whatever the equivalent to `nbytes` is? Seems ugly to mix `float` and `int` values.
> 
> Perhaps this suggests that, more generally, we ought to be counting in bits rather than bytes?

Currently on CPU, GPU and TPU the byte size of `jnp.ones(1, dtype=jnp.int4)` is 1, but the byte size `jnp.ones(2, dtype=jnp.int4)` is also 1 since it's packed.

However, it's possible that a platform doesn't guarantee packing for int4, I don't think it's possible to have a jit-compatible function counting bytes currently. I believe users interested in RAM/VRAM size should use a custom lambda with `jax.tree.map` to explicitly state their own assumptions.

I believe fp4 will suffer from the same problem as int4, I'm not sure there's a difference between integer or floating point representations.

We'd typically make an assumption that 1 byte is 8 bits, so it shouldn't change the calculation and it doesn't solve the packing representation problem of fp4/int4.

Perhaps this function could be `tree_nbytes`, but I'm really wary of exposing `bytes` counting in optax because it can be a source of issues and can be potentially confusing for users working with large models.

I'd prefer not to merge this function into optax. I find the [haiku version](https://github.com/google-deepmind/dm-haiku/blob/8f105828d3126decc62bafc9888d8049556b7bc6/haiku/_src/utils.py#L220#L255) actively confusing when working with int4 quantized models.

carlosgmartin:
To address the half-byte issue, I've switched to a `tree_bits` function.

It measures the _intrinsic size_ of a pytree (not how it's packed or laid out on devices, which is hardware-dependent). This can be used to measure the storage size or capacity of a model, for example.

Because it counts bits, it correctly handles fractional-byte types like `int4`, `int2`, `float4`, `bool`, etc.

carlosgmartin:
@rdyro I fixed the issue by switching to a `tree_bits` function.

Issue #1187, opened 2025-01-31
Support for L-BFGS-B


jithendaraa:
Is there currently support for L-BFGS-B? If not, is it planned to be included in optax?

fabianp:
there's no current plan, but contributions are welcomed

emilyfertig:
It would be great to have the [Jaxopt one](https://github.com/google/jaxopt/blob/main/jaxopt/_src/lbfgsb.py) ported

vroulet:
Classical optimization solvers are currently out of scope of optax.
Please consider looking at [optimistix](https://github.com/patrick-kidger/optimistix).

PR #1349, opened 2025-06-20
Add lbfgs-b optimizer

leochlon:
fixes https://github.com/google-deepmind/optax/issues/1187 with an implementation for L-BFGS-B using optax/jax syntax.

test plan:
integration tests python -m pytest test_integrated_lbfgs_b.py -v

### DIFF
--- a/optax/__init__.py
+++ b/optax/__init__.py
@@ -44,6 +44,7 @@ from optax._src.alias import fromage
 from optax._src.alias import lamb
 from optax._src.alias import lars
 from optax._src.alias import lbfgs
+from optax._src.alias import lbfgs_b
 from optax._src.alias import lion
 from optax._src.alias import MaskOrFn
 from optax._src.alias import nadam
@@ -366,6 +367,7 @@ __all__ = (
     "lamb",
     "lars",
     "lbfgs",
+    "lbfgs_b",
     "lion",
     "linear_onecycle_schedule",
     "linear_schedule",
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -16,19 +16,23 @@
 
 from collections.abc import Callable
 import functools
-from typing import Any, Optional, Union
+from typing import Any, Optional, Union, Tuple
 import warnings
 
 import jax
 import jax.numpy as jnp
+from jax import lax
 from optax._src import base
 from optax._src import clipping
 from optax._src import combine
+from optax._src import constraints
 from optax._src import factorized
 from optax._src import linesearch as _linesearch
 from optax._src import transform
 from optax._src import utils
+from optax import projections
 from optax._src import wrappers
+
 
 
 MaskOrFn = Optional[Union[Any, Callable[[base.Params], Any]]]
@@ -153,8 +157,8 @@ def adadelta(
   Adadelta is a stochastic gradient descent method that adapts learning rates
   based on a moving window of gradient updates. Adadelta is a modification of
   Adagrad.
-  It addresses the diminishing learning rates problem in Adagrad by maintaining running averages of squared
-  gradients.
+  It addresses the diminishing learning rates problem in Adagrad by maintaining
+  running averages of squared gradients.
 
   The weight update :math:`\Delta w_t` for this optimizer is given as follows:
 
@@ -162,7 +166,8 @@ def adadelta(
       \begin{align*}
 
       &E[g^2]_t = \rho \cdot E[g^2]_{t-1} + (1-\rho) \cdot g_t^2 \\
-      &\Delta w_t = -\frac{\sqrt{E[\Delta w^2]_{t-1} + \epsilon}}{\sqrt{E[g^2]_t + \epsilon}} \cdot g_t
+      &\Delta w_t = -\frac{\sqrt{E[\Delta w^2]_{t-1} + \epsilon}}{
+        \sqrt{E[g^2]_t + \epsilon}} \cdot g_t
 
       \end{align*}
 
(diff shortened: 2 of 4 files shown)


leochlon:
@fabianp when you have a moment could you please review my PR? No rush, but I'd appreciate your feedback when convenient =]

emilyfertig:
Hi, thank you for the PR! Could you confirm that this is L-BFGS-B as presented in this [paper?](https://users.iems.northwestern.edu/~nocedal/PDFfiles/limited.pdf) I'm having trouble seeing how this code maps to the algorithm in Section 2, and at first pass it looks to me like it's not doing the same thing. Another reference would be [Jaxopt L-BFGS-B](https://github.com/google/jaxopt/blob/main/jaxopt/_src/lbfgsb.py).

Also, could you add tests that compare the results with Scipy L-BFGS-B instead of L-BFGS?

leochlon:
Hey @emilyfertig, thanks for reviewing!

You're absolutely right that in the paper (Byrd et al. Sec 2), L-BFGS-B is presented as a single monolithic algorithm that:

1. Identifies the active set
2. Projects the gradient
3. Runs the two-loop recursion
4. Does bound-aware line search
5. Clips and updates parameters

Why I had to split it:
Optax's interface is rigid - every transform must follow:

init_fn(params) → state
update_fn(updates, state, params) → (updates, state)

There's no way to pass a loss function or return objective values through chain(). The Armijo line search needs both, so it has to live outside the transform pipeline.

I went with this approach to make it as native to Optax's flow as possible:
Chain handles: gradient projection → two-loop recursion → descent direction
External driver handles: per-coordinate step limits → Armijo backtracking → parameter clipping

All the math is unchanged - I just factored Byrd et al.'s monolithic routine into helper functions that fit Optax's composable design. The active set identification, projected gradients, two-loop recursion, bound-aware line search, and final clipping are all there, just split across the transform chain and external driver.

This separation lets users compose it with other Optax transforms while keeping the full L-BFGS-B algorithm intact. Would you prefer a different factorization that still respects Optax's interface constraints?


This is the Scipy comparison as requested:

`
import jax
import jax.numpy as jnp
import numpy as np
from scipy.optimize import minimize
import optax
from optax._src import alias


def test_lbfgs_b_vs_scipy():
    """Compare Optax L-BFGS-B with SciPy on standard optimization problems."""
    
    def rosenbrock(x):
        return (1.0 - x[0])**2 + 100.0 * (x[1] - x[0]**2)**2
    
    def quadratic(x):
        return x[0]**2 + 10 * x[1]**2
    
    # Test cases
    test_cases = [
        (quadratic, jnp.array([3.0, 2.0]), None, "Quadratic"),
        (rosenbrock, jnp.array([-1.2, 1.0]), None, "Rosenbrock"),
        (quadratic, jnp.array([3.0, 2.0]), [(0.5, 2.0), (-1.0, 1.0)], "Constrained"),
    ]
    
    
    for func, x0, bounds, name in test_cases:
        print(f"\n{name} test:")
        
        # SciPy optimization
        scipy_result = minimize(
            lambda x: float(func(jnp.array(x))),
            x0, method='L-BFGS-B',
            jac=lambda x: np.array(jax.grad(func)(jnp.array(x))),
            bounds=bounds,
            options={'ftol': 1e-12, 'gtol': 1e-8}
        )
        
        # Optax optimization  
        if bounds:
            lower = jnp.array([b[0] for b in bounds])
            upper = jnp.array([b[1] for b in bounds])
            optimizer = alias.lbfgs_b(lower_bounds=lower, upper_bounds=upper)
        else:
            optimizer = alias.lbfgs_b()
        
        params = x0.copy()
        
        for i in range(100):
            loss = func(params)
            grads = jax.grad(func)(params)
            
            if jnp.linalg.norm(grads) < 1e-8:
                break
                
            updates, opt_state = optimizer.update(
                grads, opt_state, params,
                value=loss, grad=grads, value_fn=func
            )
        
        # Compare results
        diff = jnp.linalg.norm(scipy_result.x - params)
        value_diff = abs(scipy_result.fun - func(params))
        
        print(f"  SciPy:  x={scipy_result.x}, f={scipy_result.fun:.2e}")
        print(f"  Optax:  x={params}, f={func(params):.2e}")
        print(f"  Difference: |x|={diff:.2e}, |f|={value_diff:.2e}")
        
        # Validation
        assert diff < 0.1, f"Solutions differ too much: {diff}"
        assert value_diff < 0.1, f"Function values differ: {value_diff}"
        print(f"Test passed")
    
    print(f"\n{'='*50}")
    print("All tests passed! Optax L-BFGS-B matches SciPy performance.")


if __name__ == "__main__":
    test_lbfgs_b_vs_scipy()
`


Quadratic test:
  SciPy:  x=[-2.18893242e-14  2.52296892e-13], f=6.37e-25
  Optax:  x=[-2.2204460e-14  2.5202063e-13], f=6.36e-25
  Difference: |x|=4.19e-16, |f|=1.38e-27
Test passed

Rosenbrock test:
  SciPy:  x=[1.00000018 1.0000004 ], f=1.44e-12
  Optax:  x=[1. 1.], f=0.00e+00
  Difference: |x|=3.77e-07, |f|=1.44e-12
Test passed

Constrained test:
  SciPy:  x=[5.00000000e-01 1.20411866e-15], f=2.50e-01
  Optax:  x=[ 5.000000e-01 -6.938894e-18], f=2.50e-01
  Difference: |x|=1.21e-15, |f|=0.00e+00
Test passed

emilyfertig:
The code appears to be missing local minimization along the piecewise linear path described in section 4 of the [paper](https://users.iems.northwestern.edu/~nocedal/PDFfiles/limited.pdf). If I'm missing it somehow, could you point me to where exactly it is (line numbers)?

leochlon:
I omitted Section 4 subspace minimisation for four reasons:

1) Full subspace solving would require major extensions (active sets, reduced Hessians, small QPs) that clash with Optax's minimal-transform philosophy.

2) Simple Armijo backtracking performs well in most ML cases. Subspace solving only helps on tiny, ill-conditioned problems and can slow down larger models.

3) Both SciPy and Jaxopt make this step optional/approximate.

4) JAX's high computational overhead already requires considerable warmup time. Adding the complex subspace step would significantly increase JIT compilation costs without meaningful benefit for most users.

The full Section 4 behaviour can be added if needed, but I prioritised lean implementation for typical use cases.

leochlon:
5d Rosenbrock Optimisation: 
SciPy L-BFGS-B:   38 iterations, final value: 3.945011e-14, time: 0.1679s
JIT-Compiled L-BFGS-B:        38 iterations, final value: 3.945009e-14, time: 0.1541s

leochlon:
@emilyfertig 

Since my last commit, I've completely rewritten the implementation to fully follow the Byrd et al. (1995) paper. Here's what's now implemented:

The code now implements the full L-BFGS-B algorithm as described in the paper:

Algorithm 1 (Main Iteration) - Lines 89-130 in constraints.py:
Step 1: Compute search direction (GCP for k=0, L-BFGS for k>0)
Step 2: Optax native zoom line search 
Step 3: Update parameters with bounds projection
Steps 4-5: Update limited memory with correction pairs {s_k, y_k}


Algorithm 2 (Generalised Cauchy Point) - Lines 31-65 in get_direction():
Implements the exact GCP computation from Section 5
Finds breakpoints along the projected gradient path P(x - t∇f(x))
Evaluates quadratic model at each breakpoint (Equation 5.8)
This was completely missing in my original implementation

RE: Section 4 Subspace Minimisation: You're absolutely right - this was missing before. I've now added proper subspace minimisation in the line search that finds the optimal step along the search direction while respecting bounds.

Key Additions Since Last Commit
Proper GCP Implementation: The get_direction() function now implements Algorithm 2 exactly as described, with breakpoint finding and quadratic model evaluation.

Bounds Projection: Added _project_bounds() helper that ensures all iterates respect box constraints.

SciPy Comparison Tests: Added tests comparing against scipy.optimize.minimize(method='L-BFGS-B') - our implementation now matches SciPy's results within 8% iteration count and achieves better final objective values.




Let me know if any further explanation or testing is needed

leochlon:
@emilyfertig Let me know if any further explanation or testing is needed. @vroulet would love to get your opinion too.

vroulet:
Hello @leochlon, 
After some discussion, we decided that optimization solvers like lbfgs-b are currently out of scope of optax.
We are a small team of maintainers and do not have the time to review carefully such PRs.
We invite you to consider participating to [optimistix](https://github.com/patrick-kidger/optimistix).
We hope you understand and thank you for your efforts.

leochlon:
@vroulet Thanks for the feedback. I understand the scope constraints and appreciate you taking the time to consider these PRs, I put a lot of work into them. I'll look into Optimistix as you suggested. Thanks for building such a great library.

PR #1352, opened 2025-06-21
Add function tree_allclose.

carlosgmartin:
Adds a function `tree_allclose` to Optax's [tree utilities](https://optax.readthedocs.io/en/latest/api/utilities.html#tree) that is analogous to [`jax.numpy.allclose`](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.allclose.html).

This can later be re-used for comparison tests throughout the codebase.


hbq1:
Hi! Thanks for your PR. 

Does this functionality differ from https://chex.readthedocs.io/en/latest/api.html#chex.assert_trees_all_close? If not, it might be simpler to use the one from chex as it's specialised on assertions.

carlosgmartin:
@hbq1 That function converts to numpy, makes assertions, and raises exceptions. In contrast, this function stays purely in JAX, so its output can be used upstream for purely-functional JAX computations.

docs/api/utilities.rst
```diff
@@ -124,6 +124,7 @@ Tree
     tree_vdot
     tree_where
     tree_zeros_like
+    tree_allclose
```
vroulet: Keep functions in alphabetical order

docs/api/utilities.rst
```diff
@@ -240,3 +241,7 @@ Tree where
 Tree zeros like
 ~~~~~~~~~~~~~~~
 .. autofunction:: tree_zeros_like
+
+Tree all close
```
vroulet: Same here

PR #1354, opened 2025-06-23
Simplify code by using new tree.size function.

carlosgmartin:
Simplify code by using the new [tree.size](https://optax.readthedocs.io/en/latest/api/utilities.html#optax.tree_utils.tree_size) function.


vroulet (approved):
Nice, thanks!

PR #1360, opened 2025-06-24
Fix ruff to check line-length=80

rdyro:


pyproject.toml
```diff
@@ -68,12 +68,16 @@ docs = [
...
+[tool.ruff]
+line-length = 80
+
 [tool.ruff.lint]
 select = [
     "F",
     "E",
     "W291",  # whitespace at the end of the line
     "B023",  # pylint's cell-var-over-loop, closures capturing variables in loop
+    "E501",
```
emilyfertig: Worth adding a comment here like in the above two?

emilyfertig (approved):

pyproject.toml
rdyro: ah, great catch, this just needs to be deleted, thanks!

Issue #1358, opened 2025-06-24
Add segmentation based losses


aymuos15:
Motivation - Segmentation is a pretty common CV (Computer Vision) task so I imagine it may increase adoption? I am trying to use it myself and hence the issue here. 

The two most commonly used losses are Dice loss and Dice + CE loss (I am biased to medical imaging, so not sure about general CV).

Would this be on the agenda of the repo? Concurrently, would this be okay for a PR? I would then like to work on it then.

Below is a rough implementation of code and tests based on the existing losses for Dice and Dice + CE Loss.

### Mock Code

```python
from typing import Union

import chex
import jax
import jax.numpy as jnp


def dice_loss(
    logits: chex.Array,
    labels: chex.Array,
    smooth: float = 1e-6,
    axis: Union[int, tuple[int, ...], None] = -1,
    where: Union[chex.Array, None] = None,
) -> chex.Array:
    """Computes the Dice loss between logits and labels.

    Args:
      logits: Unnormalized log probabilities, with shape [..., num_classes].
      labels: One-hot encoded labels or binary labels, with shape
        [..., num_classes].
      smooth: Smoothing factor to avoid division by zero. Defaults to 1e-6.
      axis: Axis or axes along which to compute. Defaults to -1.
      where: Elements to include in the computation. Defaults to None.

    Returns:
      Dice loss between each prediction and the corresponding target
      distributions, with shape [...].

    Examples:
      >>> import optax
      >>> import jax.numpy as jnp
      >>> jnp.set_printoptions(precision=4)
      >>> # Binary classification
      >>> logits = jnp.array([1.0, -1.0, 1.0])
      >>> labels = jnp.array([1.0, 0.0, 1.0])
      >>> print(optax.dice_loss(logits, labels))
      0.0
      >>> # Multi-class classification
      >>> logits = jnp.array([[1.0, -1.0, 2.0], [1.0, 2.0, -1.0]])
      >>> labels = jnp.array([[1.0, 0.0, 0.0], [0.0, 1.0, 0.0]])
      >>> print(optax.dice_loss(logits, labels))
      0.5
    """
    chex.assert_type([logits], float)
    
    # Convert logits to probabilities
    probs = jax.nn.sigmoid(logits) if logits.shape[-1] == 1 else jax.nn.softmax(logits, axis=-1)
    
    # Compute intersection and cardinality
    intersection = jnp.sum(probs * labels, axis=axis)
[...]

PR #1366, opened 2025-06-27
add segmentation based (dice) loss

aymuos15:
As proposed in https://github.com/google-deepmind/optax/issues/1358, it would be nice to have segmentation based losses as well. A quick google search (and perplexity's search) shows that there really isn't a standard implementation for dice loss in JAX.

The issue mentions DSC + CE as well. However, I am actually not very sure how exactly this wants to be implemented, given that there is no great reference standard to go by. My personal suggestion would be to just focus on dice first, and then do a separate PR for a follow up of including more variants. So I would love some advice on what would be the best way to proceed with this. Happy to include/remove anything.

For a first pass, I tried to cover the basic bases which I generally do, and keep a reasonable amount of tests to cover said bases.

Just to note: If it means anything to anyone here, there really isn't a standard followed implementation part of the torch ecosystem either.

### DIFF
--- a/docs/api/losses.rst
+++ b/docs/api/losses.rst
@@ -4,11 +4,13 @@ Losses
 .. currentmodule:: optax.losses
 
 .. autosummary::
+    binary_dice_loss
     convex_kl_divergence
     cosine_distance
     cosine_similarity
     ctc_loss
     ctc_loss_with_forward_probs
+    dice_loss
     hinge_loss
     huber_loss
     kl_divergence
@@ -16,6 +18,7 @@ Losses
     l2_loss
     log_cosh
     make_fenchel_young_loss
+    multiclass_generalized_dice_loss
     multiclass_hinge_loss
     multiclass_perceptron_loss
     multiclass_sparsemax_loss
@@ -33,6 +36,7 @@ Losses
     squared_error
     triplet_margin_loss
 
+
 Convex Kullback Leibler divergence
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 .. autofunction:: convex_kl_divergence
@@ -49,6 +53,12 @@ Connectionist temporal classification loss
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 .. autofunction:: ctc_loss
 .. autofunction:: ctc_loss_with_forward_probs
+
+Dice loss
+~~~~~~~~~
+.. autofunction:: dice_loss
+.. autofunction:: multiclass_generalized_dice_loss
+.. autofunction:: binary_dice_loss
 
 Fenchel Young loss
 ~~~~~~~~~~~~~~~~~~
--- a/optax/losses/_segmentation.py
+++ b/optax/losses/_segmentation.py
@@ -1,0 +1,269 @@
+# Copyright 2019 DeepMind Technologies Limited. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Segmentation losses."""
+
+from typing import Optional
+
+import chex
+import jax
+import jax.numpy as jnp
+
+
+def dice_loss(
+    predictions: chex.Array,
+    targets: chex.Array,
+    *,
+    class_weights: Optional[chex.Array] = None,
+    smooth: float = 1.0,
+    apply_softmax: bool = True,
+    reduction: str = "mean",
+    ignore_background: bool = False,
+    axis: Optional[chex.Array] = None,
+) -> chex.Array:
+  r"""Computes the Dice Loss for multi-class segmentation.
+
+  Computes the Soft Dice Loss for segmentation tasks. Works for both binary
+  and multi-class segmentation. For binary segmentation, use targets with
+  shape [..., 1] or [...] and predictions with corresponding logits.
+
+  The loss is computed per class and then averaged (or summed) across classes.
+  For class c:
+
+  .. math::
+    intersection_c = \sum_i^{N} p_{i,c} \cdot t_{i,c}
+    \\
+    dice_c = \frac{2 \cdot intersection_c + smooth}{
+      \sum_i^{N} p_{i,c} + \sum_i^{N} t_{i,c} + smooth
+    }
+
+  where:
+      - :math:`p_{i,c}` is the predicted probability for class c at pixel i
+      - :math:`t_{i,c}` is the target value (0 or 1) for class c at pixel i
+      - N is the total number of pixels
+
+  Args:
+      predictions: Logits of shape [..., num_classes] for multi-class or
+                  [..., 1] or [...] for binary segmentation.
+      targets: One-hot encoded targets of shape [..., num_classes] for
+              multi-class or binary targets of shape [..., 1] or [...] for
+              binary.
+      class_weights: Optional weights for each class of shape [num_classes].
+          If None, all classes weighted equally.
+      smooth: Smoothing parameter to avoid division by zero and improve
+             gradient stability.
+      apply_softmax: Whether to apply softmax to predictions. Set False if
+          predictions are already probabilities.
+      reduction: How to reduce across classes: 'mean', 'sum', or 'none'.
+        'none' returns per-class losses.
+      ignore_background: If True, excludes the first class (index 0) from loss
+            computation. Useful when class 0 represents background.
+      axis: Axis or sequence of axes to sum over when computing the loss.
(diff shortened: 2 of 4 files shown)


aymuos15:
Hi @rdyro . This is my first pass. I'll fix the lint issue (wasn't aware of this specific one, will keep in mind) based on what we decide here if thats okay. Thanks again :)

vroulet:
Thank you very much @aymuos15 ! Looks pretty good to me.

@selamw1 or @rajasekharporeddy would you have time to thoroughly review this PR? Thanks !

docs/api/losses.rst
```diff
@@ -32,6 +32,10 @@ Losses
     sparsemax_loss
     squared_error
     triplet_margin_loss
+    dice_loss
```
rajasekharporeddy: Keep functions in alphabetical order

docs/api/losses.rst
```diff
@@ -120,3 +124,9 @@ Sparsemax
 Triplet margin loss
 ~~~~~~~~~~~~~~~~~~~
 .. autofunction:: triplet_margin_loss
+
```
rajasekharporeddy: Keep functions in alphabetical order

optax/losses/_segmentation.py
```diff
@@ -0,0 +1,244 @@
...
+) -> chex.Array:
+    """Computes the Dice Loss for multi-class segmentation.
+
+    Computes the Soft Dice Loss for segmentation tasks. Works for both binary
+    and multi-class segmentation. For binary segmentation, use targets with
+    shape [..., 1] or [...] and predictions with corresponding logits.
+
+    The loss is computed per class and then averaged (or summed) across classes.
+    For class c: dice_c = (2 * intersection_c + smooth) /
+                         (pred_c + target_c + smooth)
```
rajasekharporeddy:
```suggestion
    For class ``c``: 

    .. math::
      dice_c = (2 * intersection_c + smooth) / (pred_c + target_c + smooth)
```

optax/losses/_segmentation.py
```diff
@@ -0,0 +1,244 @@
...
+def dice_loss(
+    predictions: chex.Array,
+    targets: chex.Array,
+    *,
+    class_weights: Optional[chex.Array] = None,
+    smooth: float = 1.0,
+    apply_softmax: bool = True,
+    reduction: str = "mean",
+    ignore_background: bool = False,
+) -> chex.Array:
```
rajasekharporeddy, line 34: I believe it would be more robust to provide `axis` as an explicit argument, allowing `spatial_axes` to be dynamically determined within the function. WDYT?

optax/losses/_segmentation.py
rajasekharporeddy: Also, define `intersection_c` here.

optax/losses/_segmentation.py
```diff
@@ -0,0 +1,244 @@
...
+    # Convert logits to probabilities
+    if apply_softmax:
+        if predictions.shape[-1] == 1:
+            # Binary case - use sigmoid
+            probs = jax.nn.sigmoid(predictions)
+        else:
+            # Multi-class case - use softmax
+            probs = jax.nn.softmax(predictions, axis=-1)
+    else:
+        probs = predictions
```
rajasekharporeddy: Use `jax.lax.cond` to remove these conditional branches and pass the `pylint` tests.
```suggestion
    probs = jax.lax.cond(
        apply_softmax,
        lambda preds: jax.lax.cond(
            preds.shape[-1] == 1,
            jax.nn.sigmoid,
            lambda p: jax.nn.softmax(p, axis=-1),
            preds
        ),
        lambda preds: preds,
        predictions
    )
```

optax/losses/_segmentation.py
aymuos15: I do agree with this. Will update this in the next commit. Thanks for pointing that out.

aymuos15:
The doctests were failing hence the repeated commits.

Thank you very much @rajasekharporeddy for the review.

rajasekharporeddy:
@vroulet The PR looks good to me.

optax/losses/_segmentation.py
```diff
@@ -0,0 +1,272 @@
...
+                f"Found sum range: [{jnp.min(pred_sums):.6f}, "
+                f"{jnp.max(pred_sums):.6f}]"
+            )
+
+    # Convert logits to probabilities using jax.lax.cond to avoid control flow
+    probs = jax.lax.cond(
+        apply_softmax,
+        # True branch: apply sigmoid or softmax based on number of classes
+        lambda: jax.lax.cond(
+            predictions.shape[-1] == 1,
```
rdyro: This should probably be python control flow since it's a static condition.

optax/losses/_segmentation.py
```diff
@@ -0,0 +1,272 @@
...
+            raise ValueError(
+                "When apply_softmax=False, predictions must be valid "
+                "probability distributions that sum to 1 along the class axis. "
+                f"Found sum range: [{jnp.min(pred_sums):.6f}, "
+                f"{jnp.max(pred_sums):.6f}]"
+            )
+
+    # Convert logits to probabilities using jax.lax.cond to avoid control flow
+    probs = jax.lax.cond(
+        apply_softmax,
```
rdyro: Likewise static condition, python control flow should be enough

optax/losses/_segmentation.py
aymuos15: Ah okay. The pylint test was failing here and that's why this was done. Will try to revert this in a way the tests don't fail. Thanks!

rdyro: Please revert. If pylint still complains about too many conditionals you can try refactoring a little (it just counts if statements and their nesting within a single parent scope).

If it's too annoying still, feel free to add a pylint ignore comment [https://pylint.pycqa.org/en/latest/user_guide/messages/message_control.html](https://pylint.pycqa.org/en/latest/user_guide/messages/message_control.html)

aymuos15: Thank you very much for the info and the suggestions.

optax/losses/_segmentation.py
```diff
@@ -0,0 +1,271 @@
...
+
+from typing import Optional
+
+import chex
+import jax
+import jax.numpy as jnp
+
+
+def dice_loss(
+    predictions: chex.Array,
```
Logon27, line 25: If these are the logits I think you should actually call the input `logits`. A lot of other optax functions have the parameter named like that. Just my opinion. I am gonna try out the code from this PR for my own uses and see how it goes.

aymuos15: That's fair. Thanks! Please update here if you face any issues. Then I will accordingly update this. Otherwise I will just make that small change.

Logon27: Oh I see you have an apply_softmax input parameter. Is there a scenario where you wouldn't want softmax applied? Just curious.

Also I tried out the binary dice loss yesterday and it seems to be working fine.

aymuos15: So I had a re-look at the code (sorry Its been kinda long since this started haha). 

Given that the the apply_softmax exists, I think ill resort to no change at this point.

There are many non-standard scenarios where you would need the logits themselves and hence i guess it is a good practice?

rdyro:
@aymuos15 could you reindent your changes to use indent-width=2?

This is needed to pass the internal checks, but I want you to get proper credit in git history.

Logon27:
Also it does not have to be this PR. But I think we should add tversky loss as well. Its basically dice loss but you can weight false positives and false negatives using input parameters.

optax/losses/_segmentation.py
Logon27: Thats fair. I was just wondering what kind of scenarios that would be. No need for changes.

aymuos15:
> @aymuos15 could you reindent your changes to use indent-width=2?
> 
> This is needed to pass the internal checks, but I want you to get proper credit in git history.

@rdyro Really appreciate the proper credit, thank you very much.

However, apologies, but this maybe an extremely naive question -- Could you please let me know a way to check for this? Looking around, I could not find a way to confirm through a linter about the index-width. I found yapf and autopep8 but they show that the other files (within the repo) are not completely using indent-width=2 so im just a bit confused.

@Logon27 I was planning to do another very small PR just for Dice + CE (and I think that should be more than enough because these two are the most common losses) anyways. I will include this as well. Thank you!

rdyro:
> However, apologies, but this maybe an extremely naive question -- Could you please let me know a way to check for this? Looking around, I could not find a way to confirm through a linter about the index-width. I found yapf and autopep8 but they show that the other files (within the repo) are not completely using indent-width=2 so im just a bit confused.

I unfortunately don't know how to set up a linter for this. You can try formatting your code with pyink (a google-compatible code formatter based on black):
```bash
$ pip install pyink
$ pyink --pyink --pyink-indentation=2 --pyink-use-majority-quotes --line-length=80 {FILENAME}
```
@aymuos15

aymuos15:
This is what I get with the current state of the commit

```bash
pyink --check --pyink-indentation=2 --pyink-use-majority-quotes --line-length=80 _segmentation.py 
pyink --check --pyink-indentation=2 --pyink-use-majority-quotes --line-length=80 _segmentation_test.py
pyink --check --pyink-indentation=2 --pyink-use-majority-quotes --line-length=80 __init__.py
All done! ✨ 🍰 ✨
1 file would be left unchanged.
All done! ✨ 🍰 ✨
1 file would be left unchanged.
All done! ✨ 🍰 ✨
1 file would be left unchanged.
```

PR #1365, opened 2025-06-27
Apply minor improvements using tree utility functions. Deprecate optax.global_norm in favor of optax.tree.norm.

carlosgmartin:
1. Edits code to use already-existent [tree utility functions](https://optax.readthedocs.io/en/latest/api/utilities.html#tree), where possible.
2. Deprecates the redundant function [`optax.global_norm`](https://optax.readthedocs.io/en/latest/api/transformations.html#optax.global_norm) in favor of [`optax.tree_utils.tree_norm`](https://optax.readthedocs.io/en/latest/api/utilities.html#optax.tree_utils.tree_norm).

PR #1369, opened 2025-06-27
Streamline interface of tree_norm function.

carlosgmartin:
Streamlines and simplifies the interface for [`optax.tree.norm`](https://optax.readthedocs.io/en/latest/api/utilities.html#optax.tree_utils.tree_norm) (no need for 3 different ways to specify the infinity norm, and 2 different ways to specify the 2-norm).

Split into a new PR from https://github.com/google-deepmind/optax/pull/1365.

### DIFF
--- a/optax/projections/_projections_test.py
+++ b/optax/projections/_projections_test.py
@@ -50,7 +50,7 @@ class ProjectionsTest(parameterized.TestCase):
         'l1': (proj.projection_l1_ball, partial(optax.tree.norm, ord=1)),
         'l2': (proj.projection_l2_ball, optax.tree.norm),
         'linf': (proj.projection_linf_ball,
-                 partial(optax.tree.norm, ord='inf')),
+                 partial(optax.tree.norm, ord=float('inf'))),
     }
 
   def test_projection_non_negative(self):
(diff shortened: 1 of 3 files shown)


vroulet:
The user will expect the same behavior as in jax, so we will stay as close as possible to jax.numpy.linalg.norm.
Maybe the "float" typing hint could be removed. But that could break some other functions and it's not a priority. 
In general cosmetic changes like this one may actually cost us more energy than it solves problems so maybe consider the cost of these changes before doing PRs. Thank you though for all your efforts and fixing many bugs!

carlosgmartin:
@vroulet [`jax.numpy.linalg.norm`](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.linalg.norm.html) doesn't allow `ord='inf'` or `ord='infinity'`:

```shell
$ py -c "from jax import numpy as jnp; jnp.linalg.norm(jnp.zeros(3), ord='inf')"
ValueError: Invalid order 'inf' for vector norm.Use 'jax.numpy.inf' instead.
```

```shell
$ py -c "from jax import numpy as jnp; jnp.linalg.norm(jnp.zeros(3), ord='infinity')"
ValueError: Invalid order 'infinity' for vector norm.
```

I'd also argue that the narrower type of `ord` (excluding `float` & matching `jax.numpy.linalg.norm`) can help prevent future bugs by catching an incorrect argument at type-check time.

PR #1374, opened 2025-07-06
Fix tree_min and tree_max to handle zero-size array leafs.

carlosgmartin:
Edit [`tree_min`](https://optax.readthedocs.io/en/latest/api/utilities.html#optax.tree_utils.tree_min) and [`tree_max`](https://optax.readthedocs.io/en/latest/api/utilities.html#optax.tree_utils.tree_max) to correctly handle pytrees that contain leafs that are zero-size arrays. For example:

Before:

```
$ py -c "import optax, jax; print(optax.tree.min([3, jax.numpy.zeros(0)]))"
ValueError: zero-size array to reduction operation min which has no identity
```

After:

```
$ py -c "import optax, jax; print(optax.tree.min([3, jax.numpy.zeros(0)]))"
3.0
```

### DIFF
--- a/optax/tree_utils/_tree_math.py
+++ b/optax/tree_utils/_tree_math.py
@@ -188,9 +188,13 @@ def tree_max(tree: Any) -> jax.typing.ArrayLike:
   Returns:
     a scalar value.
   """
-  maxes = jax.tree.map(jnp.max, tree)
-  # initializer=-jnp.inf should work but pytype wants a jax.Array.
-  return jax.tree.reduce(jnp.maximum, maxes, initializer=jnp.array(-jnp.inf))
+  def f(array):
+    if jnp.size(array) == 0:
+      return None
+    else:
+      return jnp.max(array)
+  maxes = jax.tree.map(f, tree)
+  return jax.tree.reduce(jnp.maximum, maxes, initializer=-float('inf'))
 
 
 def tree_min(tree: Any) -> jax.typing.ArrayLike:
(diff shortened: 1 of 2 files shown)


optax/tree_utils/_tree_math.py
```diff
@@ -176,9 +176,13 @@ def tree_max(tree: Any) -> chex.Numeric:
...
+  def f(array):
+    if jnp.size(array) == 0:
+      return None
+    else:
+      return jnp.max(array)
+  maxes = jax.tree.map(f, tree)
+  return jax.tree.reduce(jnp.maximum, maxes, initializer=-float("inf"))
 
 
 def tree_min(tree: Any) -> chex.Numeric:
```
vroulet: The following seems more elegant to me (similar for tree_max):

```python
import jax.numpy as jnp
import jax

def tree_min(tree) :
  mins = jax.tree.map(lambda a: jnp.min(a, initial=jnp.inf), tree)
  return jax.tree.reduce(jnp.minimum, mins, initializer=jnp.asarray(jnp.inf))

print(tree_min([jnp.asarray([1., 2., 3.]), jnp.asarray([])]))
```

carlosgmartin: I considered that approach as well, but it yields incorrect results when the dtype of `a` doesn't allow it to "reach infinity", because `jnp.min(a, initial=jnp.inf)` is constrained to the range of the dtype of `a`. Example:

```python3
import jax.numpy as jnp
import jax

def tree_min(tree) :
  mins = jax.tree.map(lambda a: jnp.min(a, initial=jnp.inf), tree)
  return jax.tree.reduce(jnp.minimum, mins, initializer=jnp.asarray(jnp.inf))

print(tree_min([1000, jnp.zeros(0, jnp.uint8)])) # incorrectly yields 255.0
print(tree_min([1000, jnp.zeros(0, jnp.int8)])) # incorrectly yields 127.0
```

I added a test that would catch this incorrect implementation.

vroulet (approved):
Thanks for adding the test, looks good to me

carlosgmartin:
@rdyro I think this might need a second approval, like you mentioned [here](https://github.com/google-deepmind/optax/pull/1439#pullrequestreview-3613911237).

rdyro:
It required some patching (inconsistent quotes), but I've pushed it through the system now, should be merged soon, thanks!

vroulet:
This got merged, so closing it manually.

PR #1376, opened 2025-07-11
Fix dtype casting inside tree_add_scale.

carlosgmartin:
The current behavior of [`tree_add_scale`](https://optax.readthedocs.io/en/latest/api/utilities.html#optax.tree_utils.tree_add_scale) is confusing:

```
$ py -c "import optax; print(optax.tree_utils.tree_add_scale([1], 0.1, [1]))"
[Array(1.1, dtype=float32, weak_type=True)]
```

This is because its [current implementation](https://github.com/google-deepmind/optax/blob/a02de8401f7dc406a6324e8dec83c5526e6db249/optax/tree_utils/_tree_math.py#L120) casts `scalar` to the dtype of `x`.

This PR modifies `tree_add_scale` to remove this casting.

Subsequently, it fixes the scan loop bodies inside [`_precondition_by_lbfgs`](https://github.com/google-deepmind/optax/blob/a02de8401f7dc406a6324e8dec83c5526e6db249/optax/_src/transform.py#L1497) to use `tree_cast_like` to ensure the carry output matches the carry input.

### DIFF
--- a/optax/_src/transform.py
+++ b/optax/_src/transform.py
@@ -1543,7 +1543,8 @@ def _precondition_by_lbfgs(
         lambda x: x[idx], (diff_params_memory, diff_updates_memory)
     )
     alpha = rhos[idx] * optax.tree.real(optax.tree.vdot(dwi, vec))
-    vec = optax.tree.add_scale(vec, -alpha, dui)
+    vec_new = optax.tree.add_scale(vec, -alpha, dui)
+    vec = optax.tree.cast_like(vec_new, vec)
     return vec, alpha
 
   precond_updates, alphas = jax.lax.scan(
@@ -1558,7 +1559,8 @@ def _precondition_by_lbfgs(
         lambda x: x[idx], (diff_params_memory, diff_updates_memory)
     )
     beta = rhos[idx] * optax.tree.real(optax.tree.vdot(dui, vec))
-    vec = optax.tree.add_scale(vec, alpha - beta, dwi)
+    vec_new = optax.tree.add_scale(vec, alpha - beta, dwi)
+    vec = optax.tree.cast_like(vec_new, vec)
     return vec, beta
 
   precond_updates, _ = jax.lax.scan(
--- a/optax/tree_utils/_tree_math.py
+++ b/optax/tree_utils/_tree_math.py
@@ -116,8 +116,7 @@ def tree_add_scale(
   """
   scalar = jnp.asarray(scalar)
   return jax.tree.map(
-      lambda x, y: (None if x is None else
-                    (x + jnp.astype(scalar, jnp.asarray(x).dtype) * y)),
+      lambda x, y: (None if x is None else (x + scalar * y)),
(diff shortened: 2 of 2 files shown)


vroulet:
@mtthss you introduced this dtype conversion in #926. Do you remember why?

@carlosgmartin add a test. It's clearly a bug so a test is welcome. Thanks again!

vroulet:
Also @carlosgmartin, can you explain
1. the bug that you then get with lbfgs 
2. whether it was caught by a test or whether you found it yourself

There may be other functions that work well thanks to this undocumented dtype conversion. So maybe we'll have to make a pass on references to tree_add_scale

carlosgmartin:
@vroulet The induced LBFGS bug was caught by a test. Without the two
```python3
vec_new = optax.tree.cast_like(vec_new, vec)
```
lines,
```shell
$ pytest optax/_src/alias_test.py -k test_state_shape_dtype_shard_stability
```
outputs
```
optax/_src/alias_test.py ..s..........s..ss...........ss...........ss.....ss [ 25%]
.................s......ss..s........ss.....ss.............................. [ 62%]
..........................ss...........ss..................ss....FF..FF..... [ 99%]
.                                                                            [100%]

===================================== FAILURES =====================================
________________ LBFGSTest.test_state_shape_dtype_shard_stability0 _________________

self = <alias_test.LBFGSTest testMethod=test_state_shape_dtype_shard_stability0>
dtype = <class 'jax.numpy.float16'>
linesearch = GradientTransformationExtraArgs(init=<function scale_by_backtracking_linesearch.<locals>.init_fn at 0x127a46980>, update=<function scale_by_backtracking_linesearch.<locals>.update_fn at 0x127a46b60>)

...

E       TypeError: scan body function carry input and carry output must have equal types, but they differ:
E       
E       The input carry vec has type float16[3] but the corresponding output carry component has type float32[3], so the dtypes do not match.
E       
E       Revise the function so that all output types match the corresponding input types.

...
```

optax/tree_utils/_tree_math_test.py
```diff
@@ -118,6 +118,11 @@ def test_tree_add_scale(self):
     got = tu.tree_add_scale(self.tree_a, 0.5, self.tree_b)
     chex.assert_trees_all_close(expected, got)
 
+  def test_tree_add_scale_dtype(self):
+    got = tu.tree_add_scale(3, 0.1, 2)
```
rdyro, line 122: Remove this test please

rdyro:
Removing the dtype cast in `tree_add_scale` is a good idea and making this explicit for L-BFGS is great! Thanks @carlosgmartin

optax/tree_utils/_tree_math_test.py
carlosgmartin: @vroulet asked me to add a test [here](https://github.com/google-deepmind/optax/pull/1376#issuecomment-3071507966). Can you clarify?

PR #1378, opened 2025-07-11
Fix 'WARNING: Title underline too short.' in utilities.rst caused by Non-negative least squares (nnls).

carlosgmartin:


rdyro:
Done as part of an internal change since that's easier to merge, but thank you for catching this!

PR #1379, opened 2025-07-11
Delete reference to non-existing document control_variates in docs/index.rst.

carlosgmartin:
Delete reference to non-existing document `control_variates` in `docs/index.rst`. Fixes the following warning when `test.sh` is executed:

```
optax/docs/index.rst:52: WARNING: toctree contains reference to non-existing document 'api/control_variates' [toc.not_readable]
```

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -64,7 +64,6 @@ for instructions on installing JAX.
    api/perturbations
    api/projections
    api/losses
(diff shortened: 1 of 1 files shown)


rdyro:
Done as part of an internal change since that's easier to merge, but thank you for catching this!

PR #1383, opened 2025-07-15
MuonClip draft in optax

rdyro:


optax/contrib/_muon.py
```diff
@@ -297,3 +297,151 @@ def muon(
...
+  return base.GradientTransformationExtraArgs(init_fn, update_fn)
+
+
+def _example_qk_proj_param_label_fn(params: base.Params):
+    params_with_paths, treedef = jax.tree.flatten_with_path(params)
+
+    def mask_fn(path_param):
+        path, param = path_param
+        str_path = "/".join(map(str, path))
+        if param.ndim == 2 and "q_proj" in str_path:
```
gagika, line 358: shouldn't you also use moun_clip for "k_proj" ?

optax/contrib/_muon.py
```diff
@@ -297,3 +297,151 @@ def muon(
...
+            return "muon_clip"
+        elif param.ndim == 2:
+            return "muon"
+        else:
+            return "adam"
+
+    return jax.tree.unflatten(treedef, jax.tree.map(mask_fn, params_with_paths))
+
+
+def muon_clip(
```
gagika, line 368: should this also take `alpha` param and be used as:

Query scaling factor: eta**alpha
Key scaling factor: eta**(1−alpha)

optax/contrib/_muon.py
rdyro: oh, right, definitely

optax/contrib/_muon.py
rdyro: eta here is just "a" scalar, so perhaps the user can pass the correctly exponentiated eta

PR #1401, opened 2025-08-11
Add lbfgs with linesearch convenience function

alexhallam:
## Summary
I never use `optax.lbfgs` without linesearch. I thought it would be a nice user experience to add this with some defaults I found from the stan library. I have tested these defaults out on some linear regression tasks and get the same results as other linear regression libraries.

## Goal

Add `optax.lbfgs_ls()` convenience function that combines L-BFGS with zoom linesearch using Stan library defaults.

## Changes
- Add `optax.lbfgs_ls()` function that provides a simpler API compared to manually configuring `optax.lbfgs` with linesearch
- Use Stan GLM defaults for linesearch parameters:
  - `max_linesearch_steps=20`
  - `initial_guess_strategy="one"`
  - `slope_rtol=1e-4`
  - `curv_rtol=0.9`
- Update documentation and exports
- Add detailed references to Stan source code
- Add simple testing

## Motivation
Currently, using L-BFGS with linesearch requires manual configuration:
```python
linesearch = optax.scale_by_zoom_linesearch(...)
optimizer = optax.lbfgs(linesearch=linesearch, ...)
```

This PR adds a convenience function `optax.lbfgs_ls()` that can be called like other optimizers:
```python
optimizer = optax.lbfgs_ls()  # Uses Stan defaults
```

### DIFF
--- a/docs/api/optimizers.rst
+++ b/docs/api/optimizers.rst
@@ -18,6 +18,7 @@ Optimizers
     lamb
     lars
     lbfgs
+    lbfgs_ls
     lion
     nadam
     nadamw
@@ -89,6 +90,10 @@ Lars
 LBFGS
 ~~~~~
 .. autofunction:: lbfgs
+
+LBFGS with Linesearch
+~~~~~~~~~~~~~~~~~~~~~
+.. autofunction:: lbfgs_ls
 
 Lion
 ~~~~
--- a/optax/__init__.py
+++ b/optax/__init__.py
@@ -44,6 +44,7 @@ from optax._src.alias import fromage
 from optax._src.alias import lamb
 from optax._src.alias import lars
 from optax._src.alias import lbfgs
+from optax._src.alias import lbfgs_ls
 from optax._src.alias import lion
 from optax._src.alias import MaskOrFn
 from optax._src.alias import nadam
@@ -373,6 +374,7 @@ __all__ = (
     "lamb",
     "lars",
     "lbfgs",
+    "lbfgs_ls",
     "lion",
     "linear_onecycle_schedule",
     "linear_schedule",
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -2717,3 +2717,81 @@ def lbfgs(
       base_scaling,
       linesearch,
   )
+
+
+def lbfgs_ls(
+    memory_size: int = 10,
+    scale_init_precond: bool = True,
+) -> base.GradientTransformationExtraArgs:
+  r"""L-BFGS optimizer with linesearch using Stan library defaults.
+
+  This is a convenience function that combines L-BFGS with a zoom linesearch
+  using default parameters from the Stan library's GLM implementation. It provides a simpler
+  API compared to manually configuring `optax.lbfgs` with linesearch.
+
+  The linesearch ensures sufficient decrease and small curvature using the
+  zoom method with the following default parameters:
+  - max_linesearch_steps=20
+  - initial_guess_strategy="one"
+  - slope_rtol=1e-4
+  - curv_rtol=0.9
+
+  Args:
+    memory_size: number of past updates to keep in memory to approximate the
+      Hessian inverse.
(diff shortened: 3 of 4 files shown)


rdyro:
Hey @alexhallam , thanks for raising this issue! We should definitely have the best heuristic defaults possible.

The current lbfgs alias should already have zoom linesearch  by default, does it not?

What do you think about using stan's defaults in the current L-BFGS implementation instead of creating a new alias?

alexhallam:
> The current lbfgs alias should already have zoom linesearch by default, does it not?

You may have to correct me. I was under the impression that the user would chose if they want to use learning rate or linesearch.

> What do you think about using stan's defaults in the current L-BFGS implementation instead of creating a new alias?

I think that is fine. I am assuming this would also set linesearch as the default as opposed to learning rate for this optimization method.

rdyro:
It seems to be linesearch by default: https://github.com/google-deepmind/optax/blob/74c68567280cc9d68bde0bdca41258808385295c/optax/_src/alias.py#L2580

Would you be able to contribute adjusting the alias defaults (like the linesearch step count) and writing a test to verify there's a lineasearch state when the optimizer is created with default arguments?

alexhallam:
Is this what you were expecting?

optax/_src/alias_test.py
```diff
@@ -1083,5 +1083,99 @@ def step(params, state):
           assert jax.tree.leaves(params)[0].dtype == dtype
 
 
+class LBFGSDefaultTest(chex.TestCase):
+  """Tests for the default lbfgs function with Stan library defaults."""
+
+  def test_lbfgs_default_has_linesearch_state(self):
```
rdyro, line 1089: only keep this test please and move it to the LBFGSTest

optax/_src/alias.py
```diff
@@ -2702,6 +2705,14 @@ def lbfgs(
     1 at the first iteration. The choice of :math:`\gamma_0` is not detailed in
     the references above, so this is a heuristic choice.
 
+  .. note::
+    The default linesearch uses Stan library defaults from the GLM
```
rdyro, line 2709: add link to the stan library

optax/__init__.py
```diff
@@ -373,6 +374,7 @@
     "lamb",
     "lars",
     "lbfgs",
+
```
rdyro, line 377: remove this space

optax/__init__.py
```diff
@@ -44,6 +44,7 @@
 from optax._src.alias import lamb
 from optax._src.alias import lars
 from optax._src.alias import lbfgs
+
```
rdyro, line 47: remove this space

docs/api/optimizers.rst
```diff
@@ -90,6 +90,7 @@ LBFGS
 ~~~~~
 .. autofunction:: lbfgs
 
+
```
rdyro, line 93: remove this space

rdyro (commented):
Thanks for the contribution, I left a couple comments.

rdyro:
> Is this what you were expecting?

That's right, thanks! I left a couple of comments

optax/_src/alias.py
```diff
@@ -2584,7 +2584,10 @@ def lbfgs(
     linesearch: Optional[
         Union[base.GradientTransformationExtraArgs, base.GradientTransformation]
     ] = _linesearch.scale_by_zoom_linesearch(
-        max_linesearch_steps=20, initial_guess_strategy='one'
+        max_linesearch_steps=20,
+        initial_guess_strategy='one',
+        slope_rtol=1e-4,
```
vroulet, line 2589:
```
  slope_rtol=1e-4,
  curv_rtol=0.9,
```
are already default parameters for the zoom linesearch.
So no need to add them.

vroulet (commented):
Thank you. I don't believe this PR is necessary. All default parameters were already properly defined. Maybe simply add that the hyper-parameters match standard implementations with a link to stan.

PR #1416, opened 2025-09-04
Implementing ANO(All new Optimizer) in Optax.

tinker495:
hello! I found a promising looking optimizer from the paper ["Ano: Faster is Better in Noisy Landscapes"](https://doi.org/10.5281/zenodo.16422081) and implemented it. 

It seems to be working, and I've left an [issue](https://github.com/Adrienkgz/ano-optimizer/issues/1#issue-3383884816) in the paper's official implementation asking the authors of the paper to confirm that their implementation is justified.


adrienkegreisz:
Hi @tinker495, thank you for your contribution and for taking the time to implement this optimizer. I’m the author of the paper and the optimizer, and I can confirm that your implementation is faithful to the paper and valid. I really appreciate your effort in bringing this optimizer to Optax!

tinker495:
I clearly verified that running test.sh and generating docs worked correctly for everything I wrote, but I don't understand why it failed in tests/doctests. If any Google developers reviewing this PR could advise how to proceed, I'd appreciate it. This test failure appears to be a package installation error rather than an issue with the content changes.

optax/contrib/_ano.py
```diff
@@ -0,0 +1,178 @@
...
+    ...  print('Objective function: {:.2E}'.format(f(params)))
+    Objective function: 1.40E+01
+    Objective function: 1.39E+01
+    Objective function: 1.39E+01
+    Objective function: 1.39E+01
+    Objective function: 1.38E+01
+
+  References:
+    Kegreisz, `Ano: Faster is Better in Noisy Landscapes
+    <https://github.com/Adrienkgz/ano-optimizer>`_.
```
aymuos15: Maybe would be nice to link the paper here instead of the repo (or alongside it)?

tinker495: Fixed!

Issue #1389, opened 2025-07-21
AMSGrad implementation differs from PyTorch/TensorFlow


sklenard:
Hello,

We noticed that models trained using the AMSGrad optimizer in Optax tend to yield slightly poorer results compared to the same models trained using PyTorch. The two implementations differ as follows:

Current Optax implementation:
![equation](https://latex.codecogs.com/png.image?\dpi{120}\bg{white}$$\hat{\nu}_t\leftarrow\nu_t/(1-\beta_2^t)$$$$\hat{\nu}_t^{\text{max}}\leftarrow\max(\hat{\nu}_{t-1}^{\text{max}},\hat{\nu}_t)$$$$\theta_{t}\leftarrow\theta_{t-1}-\gamma\hat{m_t}(\sqrt{\hat{\nu}_t^{\text{max}}}+\epsilon)$$)

PyTorch implementation (the implementation in TensorFlow follows the same algorithm):
![equation](https://latex.codecogs.com/png.image?\dpi{120}\bg{white}$$\nu_t^{\text{max}}\leftarrow\max(\nu_{t-1}^{\text{max}},\nu_t)$$$$\hat{\nu_t}\leftarrow\nu_t^{\text{max}}/(1-\beta_2^t)$$$$\theta_{t}\leftarrow\theta_{t-1}-\gamma\hat{m_t}(\sqrt{\hat{\nu_t}}&plus;\epsilon)$$)

Would it be possible to align the Optax implementation with the PyTorch/TensorFlow version? This would improve consistency across the different ML frameworks and possibly improve performance.

vroulet:
Hello @sklenard ,

Thank you for the issue, this is very well spotted. Clearly, we should try to align implementations.
Optax's version makes in fact a bit more sense to me (the bias correction should be just after the ema computation). 
We can change that, but, if you have a code ready, could you test whether removing the bias correction (for both implementations) improves results?

Thanks again!

sklenard:
Hello @vroulet ,

Indeed, removing the bias correction on the 2nd moment in both versions aligns the 2 implementations and it actually corresponds to the [original AMSGrad paper](https://openreview.net/pdf?id=ryQu7f-RZ). I have tested it on a toy model (an MLP implemented in Flax, trained to learn a polynomial function) and it seems to improve the training :

<img width="570" height="427" alt="Image" src="https://github.com/user-attachments/assets/f1566b8e-8112-4b86-8153-81187a180a2b" />

[...]

PR #1423, opened 2025-09-26
Add a bias_correction_v flag to scale_by_amsgrad to align with the original AMSGrad paper and Pytorch/tensorflow impl

vvsvictor:
Resolves [#1389](https://github.com/google-deepmind/optax/issues/1389)

### DIFF
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -1028,6 +1028,7 @@ def amsgrad(
     eps: float = 1e-8,
     eps_root: float = 0.0,
     mu_dtype: Optional[Any] = None,
+    bias_correction_v: bool = True
 ) -> base.GradientTransformationExtraArgs:
   """The AMSGrad optimizer.
 
@@ -1077,7 +1078,12 @@ def amsgrad(
   """
   return combine.chain(
       transform.scale_by_amsgrad(
-          b1=b1, b2=b2, eps=eps, eps_root=eps_root, mu_dtype=mu_dtype
+          b1=b1, 
+          b2=b2, 
+          eps=eps, 
+          eps_root=eps_root, 
+          mu_dtype=mu_dtype,
+          bias_correction_v=bias_correction_v
       ),
       transform.scale_by_learning_rate(learning_rate),
   )
--- a/optax/_src/transform.py
+++ b/optax/_src/transform.py
@@ -323,6 +323,7 @@ def scale_by_amsgrad(
     eps: float = 1e-8,
     eps_root: float = 0.0,
     mu_dtype: Optional[chex.ArrayDType] = None,
+    bias_correction_v: bool = True
 ) -> base.GradientTransformation:
   """Rescale updates according to the AMSGrad algorithm.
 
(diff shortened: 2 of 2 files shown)


rdyro:
Wouldn't setting `bias_correction_v=False` not skip applying bias b2 entirely now?

Is the point to not use the bias at all or change the order of the operations as discussed here: https://github.com/pytorch/pytorch/issues/142323

Pytorch applies this the bias in amsgrad to this day: https://github.com/pytorch/pytorch/blob/2164b661219ab0a76aa018e955ba3d8e8f99c083/torch/optim/adam.py#L509

But tensorflow does not (I think): https://github.com/keras-team/keras/blob/f6c4ac55692c132cd16211f4877fac6dbeead749/keras/src/optimizers/adam.py#L130-L150

optax/_src/transform.py
```diff
@@ -336,6 +337,9 @@ def scale_by_amsgrad(
       numerical stability when backpropagating gradients through the rescaling.
     mu_dtype: Optional `dtype` to be used for the first order accumulator; if
       `None` then the `dtype` is inferred from `params` and `updates`.
+    bias_correction_v: Whether to apply bias correction to the second moment
+      estimate before taking the elementwise maximum (``nu_max``). Set to
+      ``False`` to match the original AMSGrad paper and PyTorch/Keras behavior.
```
vroulet, line 342: "match pytorch behavior" -> no, see the conversation. 

So just say "set to `False` to match original AMSGrad paper"

optax/_src/alias.py
```diff
@@ -1077,7 +1078,12 @@ def amsgrad(
   """
   return combine.chain(
       transform.scale_by_amsgrad(
-          b1=b1, b2=b2, eps=eps, eps_root=eps_root, mu_dtype=mu_dtype
+          b1=b1,
+          b2=b2,
+          eps=eps,
+          eps_root=eps_root,
+          mu_dtype=mu_dtype,
+          bias_correction_v=bias_correction_v
```
vroulet, line 1086: `bias_correction_v` is not a great name.

`bias_correction_nu` is already better

`debias_nu` may even be better but "bias_correction" as a boolean argument is already used in e.g. rmsprop (shame on me for that naming).

@rdyro what do you think?

vroulet:
> Wouldn't setting `bias_correction_v=False` not skip applying bias b2 entirely now?
> 
> Is the point to not use the bias at all or change the order of the operations as discussed here: [pytorch/pytorch#142323](https://github.com/pytorch/pytorch/issues/142323)
> 
> Pytorch applies this the bias in amsgrad to this day: https://github.com/pytorch/pytorch/blob/2164b661219ab0a76aa018e955ba3d8e8f99c083/torch/optim/adam.py#L509
> 
> But tensorflow does not (I think): https://github.com/keras-team/keras/blob/f6c4ac55692c132cd16211f4877fac6dbeead749/keras/src/optimizers/adam.py#L130-L150

I think optax original implementation is the one that makes most sense (doing the bias correction after taking the max does not make sense to me).
However one could also simply remove the bias correction see plots in https://github.com/google-deepmind/optax/issues/1389. It seems to potentially improve and most importantly it aligns with the paper.

rdyro (approved):
Approving to incorporate the bulk of this contribution, with internal changes to reflect the latest discussion and @vroulet's comments.

Thank you!

Issue #1403, opened 2025-08-15
`piecewise_interpolate_schedule` sometimes generate nan learning rate if the number of transition steps is low.


VRichardJP:
For example:

```python
import optax

sched = optax.cosine_onecycle_schedule(
    transition_steps=3,
    peak_value=0.03668598220619202,
    pct_start=0.19848356535253986,
    div_factor=4480.102412853588,
    final_div_factor=3598.8465977101146,
)

for step in range(3):
    print(f"Step {step}: {sched(step):.6f}")

```

Output:
```
Step 0: nan
Step 1: nan
Step 2: nan
```

This is caused by 0 appearing in the boundaries, leading to `interval_sizes[0] == 0`

```
# locals() of piecewise_interpolate_schedule
{
    'interpolate_type': 'cosine',
    'init_value': 8.188648121288145e-06,
    'boundaries_and_scales': {0: 4480.102412853588, 3: 6.202242220930262e-08},
    'boundaries': (0, 3),
    'scales': (4480.102412853588, 6.202242220930262e-08),
    'bounds': array([0, 0, 3]),
    'interpolate_fn': <function _cosine_interpolate at 0x7f487f325f80>,
    'interval_sizes': array([0, 3]),
    'values': array([8.18864812e-06, 3.66859822e-02, 2.27535348e-09])
}
```

Note that a low number of transition steps will typically occur in unit tests (e.g. testing a training script with just a few items).

rdyro:
Good catch, I'll attempt to merge a fix soon, thanks for reporting!

rdyro:
This should be fixed now, thanks for reporting! Feel free to re-open if you find any issues after the fix.

PR #1425, opened 2025-09-27
Fix piecewise interpolate with 0 first split

rdyro:
Addresses https://github.com/google-deepmind/optax/issues/1403

### DIFF
--- a/optax/schedules/_schedule.py
+++ b/optax/schedules/_schedule.py
@@ -465,7 +465,7 @@ def piecewise_interpolate_schedule(
 
   bounds = np.stack((0,) + boundaries)
   values = np.cumprod(np.stack((init_value,) + scales))
-  interval_sizes = bounds[1:] - bounds[:-1]
+  interval_sizes = np.maximum(bounds[1:] - bounds[:-1], 1)
 
   def schedule(count):
     indicator = (bounds[:-1] <= count) & (count < bounds[1:])
--- a/optax/schedules/_schedule_test.py
+++ b/optax/schedules/_schedule_test.py
@@ -654,6 +654,14 @@ class PiecewiseInterpolateTest(chex.TestCase):
     with self.assertRaises(ValueError):
       _schedule.piecewise_interpolate_schedule('linear', 13.0, {5: -3})
 
+  def test_zero_interval_size(self):
+    sched = _schedule.piecewise_interpolate_schedule(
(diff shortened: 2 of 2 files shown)


optax/schedules/_schedule.py
```diff
@@ -465,7 +465,7 @@ def piecewise_interpolate_schedule(
 
   bounds = np.stack((0,) + boundaries)
   values = np.cumprod(np.stack((init_value,) + scales))
-  interval_sizes = bounds[1:] - bounds[:-1]
+  interval_sizes = np.maximum(bounds[1:] - bounds[:-1], 1)
```
emilyfertig, line 468: Maybe I'm reading this wrong but is this just checking that the keys of boundaries_and_scales is unique? Would it be better to check that explicitly and raise if not?

optax/schedules/_schedule_test.py
```diff
@@ -654,6 +654,13 @@ def test_invalid_scale(self):
     with self.assertRaises(ValueError):
       _schedule.piecewise_interpolate_schedule('linear', 13.0, {5: -3})
 
+  def test_zero_interval_size(self):
+    sched = _schedule.piecewise_interpolate_schedule(
+      'linear', 13.0, {0: 1.0, 5: 1.0}
+    )
+    if not all(jnp.isfinite(sched(i)) for i in range(6)):
+      raise ValueError('Piecewise schedule gives nan values for finite scales.')
```
emilyfertig: I'm confused by this test, should this be an assertion? (Also it looks like the test deals with non-unique scales, not bounds?)

optax/schedules/_schedule.py
rdyro: Bounds are already passed as a dictionary with bounds as the keys, so they already have to be unique. This is a sanity check normalization for if the first bound is 0, resulting in 0 - 0 = 0

optax/schedules/_schedule_test.py
rdyro: The message is confusing and this should be an assert, thanks, fiixed!

optax/schedules/_schedule.py
emilyfertig: I see, thanks! Maybe consider adding a brief test that the interpolated values are as expected if 0 is in the keys.

PR #1435, opened 2025-10-23
Add consistent rms scaling for muon update

shuningjin:
## Description

This PR introduces "consistent RMS scaling" for Muon update. 
- This is proposed by [Liu et al., 2025](https://arxiv.org/pdf/2502.16982) (aka Moonlight). See Section 2.2 and Equation 4. Their code is [here](https://github.com/MoonshotAI/Moonlight/blob/c2ad5b20c605086526a179d36901bfc41b52b44b/examples/toy_train.py#L196-L203). It is subsequently adopted in Kimi-K2 and GLM-4.5.
- In short, we multiply a scaling factor `consistent_rms * sqrt(max(fan_in, fan_out))` to update matrix. 
  - The scaling is shape-dependent, and varies for each parameter. 
  - `consistent_rms` is a hyper-parameter, usually set as `0.2` to match empirical RMS of AdamW update. Default to `None` to use the original width scaling

## Background

The codebase already has 1. I am adding 2, but keeping 1 as the default. 1 & 2 are detailed below.

### 1 Width Scaling - Previous

The previous behavior is to scale Muon update by `sqrt(max(1, fan_out / fan_in))`, following [Muon repo](https://github.com/KellerJordan/Muon/blob/f90a42b28e00b8d9d2d05865fe90d9f39abcbcbd/muon.py#L40).

Likely, it is to allow learning rate transfer across width, achieving MuP transfer automatically. Reference: [blog](https://jeremybernste.in/writing/deriving-muon), [tweet](https://x.com/YouJiacheng/status/1895019124646125791), [readme](https://github.com/KellerJordan/Muon/blob/f90a42b28e00b8d9d2d05865fe90d9f39abcbcbd/README.md?plain=1#L47). 

Implementation note: This scaling is shape-dependent, and is asymmetric in `fan_in` and `fan_out`. When reshaping, we should carefully distinguish in/out axes. On the other hand, the new scaling is symmetric.

### 2 Consistent RMS Scaling - Newly add

Main idea
- make Muon update have the same Root Mean Square (RMS) as AdamW update
- as a result, Muon can use the same hyper-parameters (learning rate, weight decay) as AdamW

Justification
- Adam and AdamW update has theoretical RMS of 1. The actual RMS is usually less than 1 (since beta1 < beta2 and epsilon > 0, e.g., beta1=0.9, beta2=0.95 eps=1e-8)
- Muon update has theoretical RMS of `sqrt(1/max(fan_in, fan_out))`, varying on matrix shape
- To cancel the effect of shape, multiply `sqrt(max(fan_in, fan_out))` to Muon update. As a result, the RMS becomes 1.
- The empirical RMS of AdamW update is around 0.2-0.4. To match this, additionally multiply `0.2`. As a result, the RMS becomes 0.2.

**Reference**: This is originally proposed by [Liu et al., 2025-02](https://arxiv.org/pdf/2502.16982) for training 16B deepseek-style model. Subsequent works follow the practice: [Essential AI, 2025-05](https://arxiv.org/pdf/2505.02222) for 4B gemma-style model, [Kimi-K2, 2025-07](https://arxiv.org/pdf/2507.20534) for 1T deepseek-style model, [GLM-4.5, 2025-08](https://arxiv.org/pdf/2508.06471) for 355B MoE.

While there are debates around the scaling scheme, it is worth adding the option as the above works have shown empirical success.

## Test

I run end-to-end test on MaxText, copying optax.muon. 

1 Width Scaling: optax before vs. optax now 
- https://diff.googleplex.com/#key=VVYKn017vzPX
- loss is same, tflop similar

2 Consistent RMS Scaling: optax before + naively change vs. optax now 
- https://diff.googleplex.com/#key=JCaD9qNl1rg8
- loss is same, tflop similar

### DIFF
--- a/optax/contrib/_muon.py
+++ b/optax/contrib/_muon.py
@@ -20,6 +20,7 @@ by Keller Jordan
 """
 
 
+import functools
 import math
 from typing import Any, Callable, NamedTuple, Optional, Union, Sequence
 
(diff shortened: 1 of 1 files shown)


optax/contrib/_muon.py
```diff
@@ -116,18 +117,85 @@ def _compute_muon_reshape(x: jax.Array, dim_nums: MuonDimensionNumbers
...
+  fan_in = math.prod(x.shape[ax] for ax in reduction_axes)
+  fan_out = math.prod(x.shape[ax] for ax in output_axes)
+  return fan_in, fan_out
+
+
+def _scale_update_for_width_transfer(
+    update: jax.Array, dim_nums: MuonDimensionNumbers
+):
+  fan_in, fan_out = _get_shape_products(update, dim_nums)
+  scale = jnp.sqrt(jnp.maximum(1, fan_out / fan_in))
```
rdyro, line 134: is this new? this is now the default, right?

optax/contrib/_muon.py
```diff
@@ -116,18 +117,85 @@ def _compute_muon_reshape(x: jax.Array, dim_nums: MuonDimensionNumbers
...
-      x.shape[ax] for ax in reduction_axes)
+  fan_in = math.prod(x.shape[ax] for ax in reduction_axes)
+  fan_out = math.prod(x.shape[ax] for ax in output_axes)
+  return fan_in, fan_out
+
+
+def _scale_update_for_width_transfer(
+    update: jax.Array, dim_nums: MuonDimensionNumbers
+):
+  fan_in, fan_out = _get_shape_products(update, dim_nums)
```
rdyro, line 133: could you leave 1 line docstring here

optax/contrib/_muon.py
```diff
@@ -116,18 +117,85 @@ def _compute_muon_reshape(x: jax.Array, dim_nums: MuonDimensionNumbers
...
+  scale = jnp.sqrt(jnp.maximum(1, fan_out / fan_in))
+  return scale * update
+
+
+def _scale_update_for_consistent_rms(
+    update: jax.Array, dim_nums: MuonDimensionNumbers, consistent_rms: float
+):
+  fan_in, fan_out = _get_shape_products(update, dim_nums)
```
rdyro, line 142: could you leave 1 line docstring here

rdyro: essentially where this comes from (e.g., the paper)

optax/contrib/_muon.py
shuningjin: Thank you. Added docstring with source.

optax/contrib/_muon.py
shuningjin: This is the previous default behavior. It remains to be the new default, as I set `consistent_rms=None` by default.

optax/contrib/_muon.py
shuningjin: Thank you. Added docstring with source.

Issue #1414, opened 2025-09-03
Honor the dtype in the function signature in `inject_hyperparams`.


sizhit2:
There is discrepancy of dtype when using `inject_hyperparams` and without injection.

When using `inject_hyperparams`, the hyperparams dtype is inferred from the params dtype.
See this:
```
import optax
import jax.numpy as jnp
optimizer = optax.inject_hyperparams(optax.adamw)(learning_rate=1e-3)
print(optimizer.init(jnp.array([1, 2], dtype=jnp.bfloat16)))
```
this will output
`InjectStatefulHyperparamsState(count=Array(0, dtype=int32), hyperparams={'learning_rate': Array(0.000999451, dtype=bfloat16), 'b1': Array(0.898438, dtype=bfloat16), 'b2': Array(1, dtype=bfloat16), 'eps': Array(1.00117e-08, dtype=bfloat16), 'eps_root': Array(0, dtype=bfloat16), 'weight_decay': Array(0.000100136, dtype=bfloat16)}, hyperparams_states={}, inner_state=(ScaleByAdamState(count=Array(0, dtype=int32), mu=Array([0, 0], dtype=bfloat16), nu=Array([0, 0], dtype=bfloat16)), EmptyState(), EmptyState()))`

The hyperparams dtype is inferred from the params dtype and uses bfloat16.

However, if check dtype of hyperparams without injection, it's float32. Example code [https://github.com/google-deepmind/optax/blob/main/optax/_src/alias.py#L601](url).
Should we respect the dtype in function signature in `inject_hyperparams` instead?

rdyro:
@sizhit2 This should now be fixed, can you confirm?

rdyro:
I want to be more precise:

```py
# hyperparameters are not available for inspection/change
optimizer = optax.adamw(1e-5)

# hyperparameter dtypes default to parameter dtypes
optimizer = optax.inject_hyperparams(optax.adamw)(learning_rate=1e-3)

# hyperparameter `learning_rate` is in float32, the rest in params dtype
optimizer = optax.inject_hyperparams(optax.adamw)(
    learning_rate=jnp.array(1e-3, "float32")
)

# only hyperparameter `learning_rate` is exposed and is in float32
optimizer = optax.inject_hyperparams(
    lambda learning_rate: optax.adamw(learning_rate)
)(lr=jnp.array(1e-3, "float32"))

[...]

PR #1445, opened 2025-10-25
fix: default hyperparams to float32 unless specified

akramcodez:
Respect hyperparam_dtype in inject_hyperparams.

If hyperparam_dtype is not provided, injected hyperparameters now default to jnp.float32 rather than inheriting dtypes from parameters/updates. Explicit hyperparam_dtype values continue to be honored. This avoids accidental low-precision hyperparameters (e.g., bfloat16) while preserving explicit control.

- Files: _inject.py
- Tests: Local suite passed + targeted schedule tests passed. Also verified via small reproducer: default → float32; explicit hyperparam_dtype=jnp.bfloat16 → bfloat16.
- Note: A unit test for default dtype is prepared but omitted per instruction; can include if requested.

Fixes: #1414

### DIFF
--- a/optax/schedules/_inject.py
+++ b/optax/schedules/_inject.py
@@ -152,9 +152,7 @@ def inject_hyperparams(
     def init_fn(params):
       count = jnp.zeros([], jnp.int32)
       if hyperparam_dtype is None:
-        dtype = getattr(
-            next(iter(jax.tree.leaves(params)), None), 'dtype', None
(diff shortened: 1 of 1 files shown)


akramcodez:
PTAL @sizhit2

rdyro:
We're addressing this internally as it requires passing additional internal tests, so I'm going to close this PR, sorry!

PR #1458, opened 2025-11-02
Generalize dice_loss with alpha/beta weighting

aymuos15:
This should now close: https://github.com/google-deepmind/optax/issues/1358

As per the discussion here: https://github.com/google-deepmind/optax/pull/1366#issuecomment-3218399437 the weighing with dice should accommodate some more overlap losses automatically.

### DIFF
--- a/optax/losses/_segmentation.py
+++ b/optax/losses/_segmentation.py
@@ -21,12 +21,27 @@ import jax.numpy as jnp
 from optax._src import utils
 
 
+def _reduce_loss(
+    loss: jax.Array, reduction: str, axis: Optional[int] = None
+) -> jax.Array:
+  if reduction == "mean":
+    return jnp.mean(loss, axis=axis)
+  elif reduction == "sum":
+    return jnp.sum(loss, axis=axis)
+  elif reduction == "none":
+    return loss
+  else:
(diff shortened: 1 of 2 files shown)


optax/losses/_segmentation.py
```diff
@@ -236,6 +277,83 @@ def multiclass_generalized_dice_loss(
   )
 
 
+def dice_plus_ce_loss(
```
rdyro: In the spirit of modularity, maybe this can be removed? Can the user add the dice and ce losses themselves? Do you see an advantage to exposing this as a single function?

rdyro (commented):
Nice, thank you for this PR! I left 1 comment

optax/losses/_segmentation.py
aymuos15: Yes, I agree. I will remove this and the related stuff and recommit. 

Apologies for such a late reply.

aymuos15:
@rdyro Thank you very much for the comment. Should've realised that myself haha. This PR has now changed to just adding the weighing. It should still close  https://github.com/google-deepmind/optax/issues/1358

vroulet (approved):
Thanks, I'll patch it internally.

Issue #1486, opened 2025-11-12
Add `label_smoothing` argument to `softmax_cross_entropy_with_integer_labels`


carlosgmartin:
**Feature request:**

1. Add a `label_smoothing` argument to [`softmax_cross_entropy_with_integer_labels`](https://optax.readthedocs.io/en/latest/api/losses.html#optax.losses.softmax_cross_entropy_with_integer_labels). Default: `0.0`, which yields the current behavior.

2. Add an `axis` argument to [`smooth_labels`](https://optax.readthedocs.io/en/latest/api/losses.html#optax.losses.smooth_labels). Default: `-1`, which yields the current behavior.

**Current approach:**

```python3
onehot_labels = nn.one_hot(labels, n_classes, axis=axis)
onehot_labels = jnp.moveaxis(onehot_labels, axis, -1)
smooth_labels = optax.smooth_labels(onehot_labels)
smooth_labels = jnp.moveaxis(smooth_labels, -1, axis)
loss = softmax_cross_entropy_with_integer_labels(logits, smooth_labels, axis=axis)
```

**Disadvantages of current approach:**

1. Cumbersome from a coding perspective.

2. Inefficient in terms of ops and memory consumption. Creates a one-hot representation of the labels and works with it. There's a more efficient implementation of the label-smoothed version that skips this one-hot representation, just like the current implementation `softmax_cross_entropy_with_integer_labels` does.

**Next steps:**

If this sounds good, I can submit a PR.

rdyro:
Thanks to your contribution in https://github.com/google-deepmind/optax/pull/1492! The flow would now be

```py
smooth_labels = optax.smooth_labels(nn.one_hot(labels, n_classes, axis=axis), axis=axis)
loss = softmax_cross_entropy(logits, smooth_labels, axis=axis)
```
which is much better (thanks!). 

I'm leaning towards **not** including label smoothing in the [`softmax_cross_entropy_with_integer_labels`](https://github.com/google-deepmind/optax/blob/27707b0319533bf80ec64c621de660e09a3a31b6/optax/losses/_classification.py#L299#L407) primarily because the logic there is already quite complicated to make use of the integer label representation. Adding smoothed labels would essentially make that function call two modes: (1) the existing integer implementation and (2) a fallback to [`softmax_cross_entropy`](https://github.com/google-deepmind/optax/blob/27707b0319533bf80ec64c621de660e09a3a31b6/optax/losses/_classification.py#L227C5-L227C26).

[...]

PR #1492, opened 2025-11-12
Add axis and where arguments to smooth_labels function.

carlosgmartin:
Adds an `axis` argument (default `-1`) and `where` argument (default `None`) to [`smooth_labels`](https://optax.readthedocs.io/en/latest/api/losses.html#optax.losses.smooth_labels).

Partially addresses https://github.com/google-deepmind/optax/issues/1486.


carlosgmartin:
@emilyfertig How does this look?

PR #1502, opened 2025-11-17
Fix momo crash when loss value is a Python float

### LINKED ISSUE #1500: `contrib.momo` crashes when loss value is a Python float
`contrib.momo.update` declares value: `Optional[jax.Array] = None`, but it never enforces that type. It calls `value.astype(state.barf.dtype)` unconditionally, so passing any Python float (or NumPy scalar) for the loss raises `AttributeError` before the optimizer can run.

```python
import jax.numpy as jnp
from optax import contrib

optimizer = contrib.momo(
    learning_rate=1.0,
    beta=0.0,
    lower_bound=0.0,
    weight_decay=0.0,
    adapt_lower_bound=False,
)

params = [jnp.array([0.0])]
grads = [jnp.array([0.0])]

optimizer.update(grads, opt_state, params, value=0.0)
```

```
  File "/data/src/test.py", line 16, in <module>
    optimizer.update(grads, opt_state, params, value=0.0)
  File "/home/hdd/miniconda3/envs/py312/lib/python3.12/site-packages/optax/contrib/_momo.py", line 136, in update_fn
    barf = bt * state.barf + (1 - bt) * value.astype(state.barf.dtype)
                                        ^^^^^^^^^^^^
AttributeError: 'float' object has no attribute 'astype'
```

Aaryan-549:
Fixes #1500

The momo and momo_adam update functions declared the value parameter as Optional[jax.Array] but never enforced that type. When calling value.astype(state.barf.dtype), passing a Python float or NumPy scalar would raise AttributeError since these types don't have an astype method.

This fix uses jnp.asarray(value, dtype=state.barf.dtype) instead of value.astype(state.barf.dtype) to handle Python floats, NumPy scalars, and JAX arrays uniformly.

### DIFF
--- a/optax/contrib/_momo.py
+++ b/optax/contrib/_momo.py
@@ -119,7 +119,7 @@ def momo(
       state: MomoState,
       params: Optional[base.Params],
       *,
-      value: Optional[jax.Array] = None,
+      value: jax.typing.ArrayLike,
       **extra_args,
   ) -> tuple[base.Updates, MomoState]:
     del extra_args  # complies with signature of GradientTransformationExtraArgs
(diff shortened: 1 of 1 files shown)


rdyro:
Thanks! Can you adjust the type annotation to reflect the value requirement? Should probably drop Optional and use `jax.typing.ArrayLike`

Aaryan-549:
@rdyro Thanks for the review! I've made the suggested changes and updated the type annotation to use jax.typing.ArrayLike. Ready for another look.

Issue #1498, opened 2025-11-17
Should `tree_sum` use `jax.tree.reduce_associative`?


SobhanMP:
[In](https://github.com/google-deepmind/optax/blob/bcda8a21e52a906a737b98c39e1a189bd54a8d25/optax/tree_utils/_tree_math.py#L169#L179) `tree_sum` is implemented using `jax.tree.map` and `jax.tree.reduce` but shouldn't `jax.tree.reduce_associative` be used instead? The runtime seems to be very close but the compile time is significantly lower (18s vs 23s) on the one problem I tested.

Aaryan-549:
Hi, I have opened a pull request #1503 that addresses this issue.

I implemented the change to use jax.tree_util.tree_reduce_associative as suggested, which should help optimize the compile time. Please take a look when you have a chance!

rdyro:
Can you point to your use case where the compile time is reduced?

Looking at [`tree_reduce_associative`](https://github.com/jax-ml/jax/blob/816e644cc4eb1a4353894ba52d5e0d92716936e5/jax/_src/tree_util.py#L468) it uses a different order of summation (associative vs cumulative from left to right), but it is unclear to me how this interacts with the compiler.

Given this is an often-used optax function and this change might affect numerics, I'd prefer to not make any changes to it unless we can identify clear benefits.

@SobhanMP @Aaryan-549

Aaryan-549:
Hey @rdyro,
The primary benefit here is for PyTrees with a very large number of leaves (e.g., large flat models or optimizers handling many separate state tensors).

Standard `tree_reduce` performs a sequential fold, creating a computation graph with linear depth O(N) relative to the number of leaves. For trees with thousands of leaves, this results in deep dependency chains that can significantly slow down XLA compilation. `tree_reduce_associative` performs the reduction in a tree-structure, reducing the graph depth to O(log N).

Regarding numerics: You are correct that this changes the summation order, which will result in slightly different floating-point values compared to a left-to-right accumulation. However, this associative reduction pattern is standard for parallel reductions (like `jax.lax.psum`) and is usually considered an acceptable trade-off for the compilation speedup in large-scale settings.

If you think the numerical drift is too risky for `tree_sum` specifically in the context of Optax, I am happy to close this, but I believe the compile time gains for large model users could be significant.

rdyro:
The  graph depth argument sounds reasonable, but we'd need to check with XLA engineers whether it actually slows down compilation.

[...]

PR #1503, opened 2025-11-18
Optimize tree_sum compile time using tree_reduce_associative

Aaryan-549:
In Response to  #1498
Changed tree_sum implementation to use jax.tree_util.tree_reduce_associative instead of jax.tree.reduce. Since addition is an associative operation, tree_reduce_associative can provide better compilation performance.

Testing shows runtime is very close but compile time is significantly lower (18s vs 23s in reported cases).

### DIFF
--- a/optax/tree_utils/_tree_math_test.py
+++ b/optax/tree_utils/_tree_math_test.py
@@ -151,7 +151,9 @@ class TreeUtilsTest(parameterized.TestCase):
     if hasattr(jax.tree, 'reduce_associative'):
       with self.subTest('associative_version_matches_regular'):
         regular_result = tu.tree_sum(self.tree_a)
-        associative_result = tu.tree_sum(self.tree_a, associative_reduction=True)
+        associative_result = tu.tree_sum(
+            self.tree_a, associative_reduction=True
+        )
         np.testing.assert_allclose(regular_result, associative_result)
 
   @parameterized.parameters(


optax/tree_utils/_tree_math.py
```diff
@@ -176,7 +177,16 @@ def tree_sum(tree: Any) -> jax.typing.ArrayLike:
     a scalar value.
   """
   sums = jax.tree.map(jnp.sum, tree)
-  return jax.tree.reduce(operator.add, sums, initializer=0)
+  # Use tree_reduce_associative for better compile time performance when
+  # available (JAX >= 0.6.0). However, tree_reduce_associative doesn't
+  # support empty trees, so we need to check for that case.
+  if hasattr(jtu, 'tree_reduce_associative'):
```
SobhanMP: I think this should be
`jax.tree.reduce_associative(operator.add, sums, initializer=0)`

and the pythonic way would probably use `AttributeError` and try catch

Aaryan-549: Hi @SobhanMP 
Thanks for the review! However, jax.tree.reduce_associative doesn't support the initializer parameter - it only accepts (function, tree). That's why the original implementation failed with "Must specify identity for parallel reduction of empty sequence" when encountering empty trees. The current approach with hasattr is appropriate here because:

1. We're checking for API availability across JAX versions (0.5.3 vs 0.6.0+)
2. The empty tree check is necessary since tree_reduce_associative lacks initializer support
3. Using hasattr for version compatibility is a common pattern in the JAX ecosystem

A try/except approach would be:
```
try:
    leaves = jax.tree.leaves(sums)
    if not leaves:
        return 0
    return jtu.tree_reduce_associative(operator.add, sums)
except AttributeError:
    return jax.tree.reduce(operator.add, sums, initializer=0)
```
But this doesn't provide much benefit over hasattr for this use case, and the hasattr check makes the version compatibility intent clearer.

carlosgmartin: @Aaryan-549 [`jax.tree.reduce_associative`](https://docs.jax.dev/en/latest/_autosummary/jax.tree.reduce_associative.html) has an `identity` argument.

Aaryan-549: You're absolutely right, thank you for the correction! I've updated the implementation to use identity=0 instead of manually checking for empty trees. Much cleaner now:
```
if hasattr(jtu, 'tree_reduce_associative'):
    return jtu.tree_reduce_associative(operator.add, sums, identity=0)
else:
    return jax.tree.reduce(operator.add, sums, initializer=0)
```

rdyro:
Please see my comments in: https://github.com/google-deepmind/optax/issues/1498

If we decide to proceed with this change, can you also switch to using `jax.tree.reduce_associative` from `jtu.tree_reduce_associative` more in line with optax's code style.

optax/tree_utils/_tree_math.py
```diff
@@ -176,7 +177,12 @@ def tree_sum(tree: Any) -> jax.typing.ArrayLike:
     a scalar value.
   """
   sums = jax.tree.map(jnp.sum, tree)
-  return jax.tree.reduce(operator.add, sums, initializer=0)
+  # Use tree_reduce_associative for better compile time performance when
```
rdyro: nit: shorten this to `use reduce_associative for a potential compilation time speedup`

optax/tree_utils/_tree_math.py
```diff
@@ -176,7 +177,12 @@ def tree_sum(tree: Any) -> jax.typing.ArrayLike:
...
+  # available (JAX >= 0.6.0), otherwise fall back to tree.reduce.
+  if hasattr(jtu, 'tree_reduce_associative'):
```
rdyro: check for an optional opt-in argument, maybe something like `associative_reduction: bool = False`

Aaryan-549:
Hey @rdyro,
Thanks for the guidance, I have modified the PR. Let me know if I missed anything by mistake.

rdyro:
One last thing, can you add a small test case at the end of this test  https://github.com/google-deepmind/optax/blob/a194389e29bc3bea2cdb0ab19158933e8ebd807b/optax/tree_utils/_tree_math_test.py#L141 as a `self.subTest` if the associative version matches the regular version?

Aaryan-549:
@rdyro
Added the test too!

rdyro (approved):
Thanks!

PR #1506, opened 2025-11-20
Fix conjugation in Newton-Schulz iterator and update tests for comple…

### LINKED ISSUE #1494: Muon: support complex parameters
The Muon optimizer currently does not support complex parameters.
A very simple and natural extension would be to make the Newton-Schulz iterator push the gradients to be semi-unitary for complex parameters, rather than the current semi-orthogonalization.
I.e. $U \Sigma V^* \rightarrow  U V^*$, rather than the current  $U \Sigma V^\top \rightarrow  U V^\top$.
The only update needed is to add a `.conj()` in `_newton_schulz_iterator.`
```
def _newton_schulz_iterator(x: jax.Array, coeffs: jax.Array) -> jax.Array:
  # Implements Newton-Schulz step f(X) = c_0 X + c_1 (XX^T)X + c_2 (XX^T)^2X,
  # with quintic form f(X) = c_0 X + (c_1 A + c_2 AA)X, where A = XX^T.
  # The NS step has the property f(X) = f(X^T)^T. That is, we can get equivalent
  # result by transposing input and output. In particular, we may transpose X
  # when rows > cols for efficiency.
  a = x @ x.T.conj()
  b = coeffs[1] * a + coeffs[2] * a @ a
  return coeffs[0] * x + b @ x
```

I also think one would have to update the dual norm scaling to reflect the complex inner product, but I am not entirely sure what this part refers to in the referenced article.
```
if adaptive:
  # Scale the orthogonalized updates by the dual norm of the original
  # updates. See https://arxiv.org/abs/2409.20325 for the derivation.
  updates = jax.tree.map(
      lambda x, y: jnp.sum(x.conj() * y) * y, mu_hat, updates
  )
```

This simple extension does not affect the use on real parameters, as the `.conj()` will be ignored under `jax.jit` afaik.

eirikfagerbakke:
Adds support for complex parameters in Muon, by adding .conj() to Newton-Schulz and the adaptive scaling step.

Also adds two tests. The first checks that the Newton-Schulz algorithm is able to make the updates orthogonal for real parameters and unitary for complex parameters. The second tests the Muon optimizer on a simple complex optimization problem.

Fixes #1494.

### DIFF
--- a/optax/contrib/_muon.py
+++ b/optax/contrib/_muon.py
@@ -199,7 +199,7 @@ def _newton_schulz_iterator(x: jax.Array, coeffs: jax.Array) -> jax.Array:
   # The NS step has the property f(X) = f(X^T)^T. That is, we can get equivalent
   # result by transposing input and output. In particular, we may transpose X
   # when rows > cols for efficiency.
-  a = x @ x.T
+  a = x @ x.T.conj()
   b = coeffs[1] * a + coeffs[2] * a @ a
   return coeffs[0] * x + b @ x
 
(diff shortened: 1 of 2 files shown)


vroulet (approved):
That's perfect, great contribution, thank you! I'll take it from here.

PR #1511, opened 2025-11-26
Minor doc fixes: Include hyperlinks to functions, classes references and correct typos

rajasekharporeddy:


optax/contrib/_ademamix.py
```diff
@@ -72,7 +72,7 @@ def scale_by_ademamix(
     b1: Exponential decay rate to track the fast EMA.
     b2: Exponential decay rate to track the second moment of past gradients.
     b3: Exponential decay rate to track the slow EMA.
-    alpha: Mixing coefficient in the linear combination fo the fast and slow
+    alpha: Mixing coefficient in the linear combination for the fast and slow
```
emilyfertig, line 75: I think this should be "linear combination of"

Issue #1509, opened 2025-11-23
Code for DoG is duplicated


carlosgmartin:
There seems to be duplicate code for [DoG](https://arxiv.org/abs/2302.12022) in the codebase:

1. optax/_src/transform.py: [`ScaleByDistanceOverGradientsState`](https://github.com/google-deepmind/optax/blob/daecb91b3f0e5de6d3e763c67362e7ac2737bb24/optax/_src/transform.py#L1343), [`scale_by_distance_over_gradients`](https://github.com/google-deepmind/optax/blob/daecb91b3f0e5de6d3e763c67362e7ac2737bb24/optax/_src/transform.py#L1351)

2. optax/contrib/_dog.py: [`DoGState`](https://github.com/google-deepmind/optax/blob/daecb91b3f0e5de6d3e763c67362e7ac2737bb24/optax/contrib/_dog.py#L37), [`scale_by_dog`](https://github.com/google-deepmind/optax/blob/daecb91b3f0e5de6d3e763c67362e7ac2737bb24/optax/contrib/_dog.py#L46)

These two should be merged together.

abheeyeee:
@carlosgmartin can you review this?

abheeyeee:
@rajasekharporeddy This PR will fix the issue

abheeyeee:
Hi @vroulet  could you please take a look when you have a moment? Thanks. Also does deepmind have communication channel? slack or discord?

emilyfertig:
Hi @abheeyeee , I just talked about this with @vroulet . Your PR demonstrates that there isn't much code overlap between the functions -- the code structure and math is the same, but one maps over pytrees and the other reduces, so it isn't possible to share much code (hence all of the branching on `layer_wise`). Our preference is to keep them as separate functions and move `scale_by_distance_over_gradients` into `contrib` and rename it `scale_by_l_dog` (l for layerwise). Let us know if you're up for making this change -- your refactoring of the `scale_by_distance_over_gradients` (or `layer_wise=True`) code is nice, since it highlights the mathematical similarity with the non-layerwise version.

abheeyeee:
@emilyfertig so what i have understood is that Adding scale_by_l_dog to Contrib, i will  create a separate scale_by_l_dog (layerwise DoG) function in the contrib module based on the refactored layer-wise implementation, and update scale_by_distance_over_gradients to be a deprecated wrapper around it.

Proposed Changes
(_dog.py)
[...]

PR #1512, opened 2025-11-28
Fix: #1509  Merge duplicate DoG implementations and add layer-wise support

abheeyeee:
Fix Issue: #1509
This PR merges the duplicate Distance over Gradients (DoG) implementations found in optax/contrib/_dog.py and optax/_src/transform.py into a single, unified implementation in optax/_src/dog.py.
Created optax/_src/dog.py which consolidates DoG and DoWG. 
The new scale_by_dognow supports a layer_wise argument.
Re-implemented scale_by_distance_over_gradients in optax/_src/transform.py to use the new scale_by_dog with layer_wise=True.
Deprecated scale_by_distance_over_gradients in favor of scale_by_dog.
Updated optax/contrib/_dog.py to be a compatibility shim importing from optax/_src/dog.py.
Added optax/_src/dog_test.py to verify both global and layer-wise behaviors, as well as legacy compatibility.

### DIFF
--- a/optax/_src/dog_test.py
+++ b/optax/_src/dog_test.py
@@ -39,7 +39,7 @@ class DoGTest(parameterized.TestCase):
   ])
   def test_scalers(self, scaler_constr):
     params = self.init_params
-    if scaler_constr == dog.scale_by_dog:
+    if scaler_constr is dog.scale_by_dog:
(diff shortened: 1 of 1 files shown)


abheeyeee:
@mtthss @rdyro @vroulet   Can you review this.

PR #1514, opened 2025-11-29
Clarify difference between kl_divergence and convex_kl_divergence

### LINKED ISSUE #1513: Clarify difference between `kl_divergence` and `convex_kl_divergence`
The difference between [`kl_divergence`](https://optax.readthedocs.io/en/latest/api/losses.html#optax.losses.kl_divergence) and [`convex_kl_divergence`](https://optax.readthedocs.io/en/latest/api/losses.html#optax.losses.convex_kl_divergence) should be clarified.

The description for `kl_divergence` is
> Computes the Kullback-Leibler divergence (relative entropy) loss.
>
> Measures the information gain achieved if target probability distribution would be used instead of predicted probability distribution.

The description for `convex_kl_divergence` is
> Computes a <ins>convex version</ins> of the Kullback-Leibler divergence loss.
>
> Measures the information gain achieved if target probability distribution would be used instead of predicted probability distribution. This version is <ins>jointly convex</ins> in p (targets) and q (log_predictions).

But the [KL divergence](https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence) is already jointly convex in P and Q, so it's not clear what the difference is. Is the difference that `convex_kl_divergence` is convex in *log* Q (which the description confusingly calls Q, even though that usually refers to the actual probabilities)? If so, this should be stated more clearly.

Implementation-wise, the only difference is that `convex_kl_divergence` adds the term $$\sum_i Q_i - \sum_i P_i$$.

zer-art:
# Description

This PR addresses issue #1513 by clarifying the documentation for `kl_divergence` and `convex_kl_divergence`.

The previous documentation did not explicitly state the mathematical difference between the two implementations, nor did it explain the specific use case for the convex variant.

**Changes made:**
- Updated `kl_divergence` docstring to identify it as the standard definition.
- Updated `convex_kl_divergence` docstring to:
  - Explicitly mention the added term: `sum(exp(log_predictions)) - sum(targets)`.
  - Clarify that this is the "Generalized KL Divergence" intended for unnormalized distributions.
  - Improved phrasing for readability and elegance.

Fixes #1513

## Type of change

- [x] Documentation update (no code logic change)


optax/losses/_classification.py
```diff
@@ -601,11 +601,14 @@ def convex_kl_divergence(
...
+  """Computes the generalized Kullback-Leibler divergence loss.
 
-  Measures the information gain achieved if target probability distribution
-  would be used instead of predicted probability distribution.
-  This version is jointly convex in p (targets) and q (log_predictions).
+  Measures the generalized divergence between distributions, specifically
+  designed to handle unnormalized predicted probabilities.
+
+  This differs from ``kl_divergence`` by adding the term
+  ``sum(exp(log_predictions)) - sum(targets)``. This ensures the loss remains
```
emilyfertig: Could you explain the "This ensures the loss remains convex" statement? The KL divergence is jointly convex in p and q, and the additional term is linear in p and q, so I don't think the additional term affects the convexity. 

Page 90 here is a good reference: https://stanford.edu/~boyd/cvxbook/bv_cvxbook.pdf Consider referencing that in the docstring too.

emilyfertig (commented):
Thank you for the PR! 

This function is misnamed, since as @carlosgmartin points out on #1514 , the KL divergence is already jointly convex. A better name would be `generalized_kl_divergence`, as you referenced in the docstring. 

If you're up for it, could you change the function name and deprecate `convex_kl_divergence`? This would mean adding `generalized_kl_divergence` to `__init__.py`, `losses/__init__.py`, and `docs/api/losses.rst`, deprecating `convex_kl_divergence` using `warn_deprecated_function`, and updating `losses/_classification_test.py` to call the new function instead.

Or, if you'd prefer, we can merge this PR since it's already an improvement, and rename the function in a follow-up.

optax/losses/_classification.py
carlosgmartin: The earliest reference I could find for this generalization of the KL divergence appears is Equation 1.5 of the following 1967 paper by [L. M. Bregman](https://en.wikipedia.org/wiki/Lev_M._Bregman):

https://www.sciencedirect.com/science/article/pii/0041555367900407

So it might be good to cite that.

carlosgmartin: @emilyfertig Since the functions are so similar, would it perhaps be better to just add a `normalized` flag to [`kl_divergence`](https://optax.readthedocs.io/en/latest/api/losses.html#optax.losses.kl_divergence)?

It can default to `True` and, when set to `False`, add the `q.sum(axis=axis, where=where) - p.sum(axis=axis, where=where)` term.

zer-art: Thanks for the detailed  review @emilyfertig!

I agree that `generalized_kl_divergence` is the correct name here. I would prefer to handle the refactoring in this PR to keep the codebase history clean.

### **My Plan**

#### **1. Refactor**
- Rename the main function to `generalized_kl_divergence`.
- Add the `warn_deprecated_function` wrapper for the old `convex_kl_divergence` (using the `functools.partial` pattern consistent with the repo).

#### **2. Docstring**
- I will remove the confusing "ensures convexity" claim. Instead, I will accurately describe this as the Bregman divergence generated by *f(x) = x log x − x*, referencing **Boyd & Vandenberghe (p. 90)**.

#### **3. Files to Update**
- `__init__.py`
- `docs/api/losses.rst`
- Test suite (to reflect the new function name)

I’ll push these changes shortly!

emilyfertig: > The earliest reference I could find for this generalization of the KL divergence appears is Equation 1.5 of the following 1967 paper by [L. M. Bregman](https://en.wikipedia.org/wiki/Lev_M._Bregman):
> 
> https://www.sciencedirect.com/science/article/pii/0041555367900407
> 
> So it might be good to cite that.

Thanks -- I'd prefer to cite Boyd, since it's more accessible/canonical, but will leave it up to @zer-art 

> @emilyfertig Since the functions are so similar, would it perhaps be better to just add a `normalized` flag to [`kl_divergence`](https://optax.readthedocs.io/en/latest/api/losses.html#optax.losses.kl_divergence)?
> 
> It can default to `True` and, when set to `False`, add the `q.sum(axis=axis, where=where) - p.sum(axis=axis, where=where)` term.

Good point. I think either could work. I'm newer to the library so I'm not sure if one or the other is more consistent with Optax's philosophy (more functions with one job, vs. fewer consolidated functions with more config options). Let me run this by the other Optax devs and respond later.

zer-art: Hi @carlosgmartin, I missed your comments while drafting my previous reply—thanks for catching that!

1. **Citation:** Great find on the Bregman (1967) paper. I will definitely include that reference in the updated docstring.
2. **Flag suggestion:** Adding a `normalized` argument to `kl_divergence` is an interesting alternative to keep the API surface smaller, though it would involve modifying the core function.

@emilyfertig Do you prefer we stick to the current plan (separate `generalized_kl_divergence` function + deprecation) or pivot to the flag approach suggested by Carlos? I am happy to implement whichever you prefer!

zer-art: Understood! Thanks for running this by the team, @emilyfertig.

I will hold off on pushing any changes until you confirm whether we should proceed with the **renaming/deprecation** approach or the `normalized` flag approach. I'm ready to implement whichever direction the team prefers.

emilyfertig: Hi @zer-art , the team decided we prefer the original renaming/deprecation plan so let's stick with that. Thanks!

zer-art: Awesome, thanks for the confirmation @emilyfertig! I'll proceed with the renaming and deprecation plan (and include the citations). Pushing the updates shortly

optax/losses/_classification.py
```diff
@@ -465,10 +466,10 @@ def multiclass_perceptron_loss(
   return jnp.max(scores, axis=-1) - _dot_last_dim(scores, one_hot_labels)
 
 
+@functools.partial(chex.warn_only_n_pos_args_in_future, n=2)
```
emilyfertig: I think this might have been a bad rebase -- this (and the chex import) were removed recently at head so they shouldn't be here (+ elsewhere).

optax/__init__.py
```diff
@@ -239,6 +239,7 @@
 WrappedSchedule = schedules.WrappedSchedule
 
 # TODO(mtthss): remove loss aliases from flat namespace once users have updated.
+generalized_kl_divergence = losses.generalized_kl_divergence
```
emilyfertig: Keep alphabetical order? (and below/elsewhere)

docs/api/losses.rst
```diff
@@ -6,6 +6,7 @@ Losses
 .. autosummary::
     binary_dice_loss
     convex_kl_divergence
+    generalized_kl_divergence
```
emilyfertig: Could you keep alphabetical order please?

zer-art:
Thanks for the feedback! I've signed off for the night, but I will fix the alphabetical ordering and clean up the rebase/imports first thing tomorrow morning.

zer-art:
I performed a clean reset to fix the bad rebase history and restored the chex-free imports. I also fixed the alphabetical ordering as requested. Ready for review!

vroulet:
Thank you for the PR!

PR #1518, opened 2025-12-02
 Fix: #1509 Merge duplicate DoG implementations and add layer-wise support 

abheeyeee:
Fix Issue: #1509 
This PR merges the duplicate Distance over Gradients (DoG) implementations found in optax/contrib/_dog.py and optax/_src/transform.py into a single, unified implementation in optax/_src/dog.py.
Created optax/_src/dog.py which consolidates DoG and DoWG.
The new scale_by_dognow supports a layer_wise argument.
Re-implemented scale_by_distance_over_gradients in optax/_src/transform.py to use the new scale_by_dog with layer_wise=True.
Deprecated scale_by_distance_over_gradients in favor of scale_by_dog.
Updated optax/contrib/_dog.py to be a compatibility shim importing from optax/_src/dog.py.
Added optax/_src/dog_test.py to verify both global and layer-wise behaviors, as well as legacy compatibility.


vroulet:
Can you keep _dog.py and _dog_test.py files, and keep them in the contrib folder?
This will ease the review process. 
Try to keep the pr as concise as possible. (It looks rather good otherwise :) )

abheeyeee:
> Can you keep _dog.py and _dog_test.py files, and keep them in the contrib folder? This will ease the review process. Try to keep the pr as concise as possible. (It looks rather good otherwise :) )

Thanks for the Feedback, i will do the changes rn.

abheeyeee:
@vroulet moved the files as you asked. This should fix the issue

optax/contrib/_dog.py
```diff
@@ -79,37 +83,87 @@ def init_fn(params: base.Params) -> DoGState:
...
 
+    if layer_wise:
+        if init_step_type == "heuristic":
+             # r_epsilon is already a tree of scalars
+             max_dist = jax.tree.map(
+                 lambda r: jnp.asarray(r, dtype=params_dtype), r_epsilon
+             )
+        else:
+             # r_epsilon is a scalar, create tree of scalars
+             max_dist = jax.tree.map(
```
selamw1: LGTM!

One small suggestion regarding mixed-precision support:

In `init_fn`, `params_dtype` is currently calculated globally and used to initialize both `max_dist` and `sum_sq_norm_grads`. When `layer_wise=True`, this effectively forces all layers to share the same precision, which could be problematic for mixed-precision models (e.g., mixing `float32` and `bfloat16`).

Consider initializing the state variables using the leaf's dtype directly so the optimizer state matches the precision of each corresponding parameter tensor:

```
max_dist = jax.tree.map(lambda p: jnp.full([], r_epsilon, dtype=p.dtype), params)
```

abheeyeee: okay, understood. i have to initialize max_dist and sum_sq_norm_grads using the dtype of each leaf parameter instead of a global params_dtype. this is how it will look like

# Before
max_dist = jax.tree.map(
    lambda p: jnp.full([], r_epsilon, dtype=params_dtype), params
)
sum_sq_norm_grads = jax.tree.map(
    lambda p: jnp.zeros([], dtype=params_dtype), params
)
# After
max_dist = jax.tree.map(
    lambda p: jnp.full([], r_epsilon, dtype=p.dtype), params
)
sum_sq_norm_grads = jax.tree.map(
    lambda p: jnp.zeros([], dtype=p.dtype), params
)

optax/_src/transform.py
```diff
@@ -1339,14 +1342,14 @@ def f(grad, prev_grad):
...
-  grad_sum_of_squares: base.OptState
-  init_params: base.OptState
+ScaleByDistanceOverGradientsState = dog.DoGState
 
 
+@functools.partial(
+    warn_deprecated_function,
+    replacement="scale_by_dog",
+    version_removed="0.2.3",
+)
```
emilyfertig, line 1398: Can we keep `scale_by_distance_over_gradients` and deprecate `scale_by_dog` instead? The former has a better name, and I don't think the core library should depend on contrib. We could add the other args at the end (here and in the state) so it's backwards compatible. We could also consider deprecating the `global_state` arg.

optax/contrib/_dog.py
```diff
@@ -79,37 +83,87 @@ def init_fn(params: base.Params) -> DoGState:
...
-    sum_sq_norm_grads = state.sum_sq_norm_grads + optax.tree.norm(
-        updates, squared=True
-    )
-    learning_rate = max_dist / jnp.sqrt(sum_sq_norm_grads + eps)
 
-    if init_step_type == "learning_rate":
-      learning_rate = jnp.where(
-          state.is_init_step, init_step_value, learning_rate
-      )
+    if layer_wise:
```
emilyfertig: This function body is almost completely different depending on `layer_wise`. To simplify the logic, could you factor out two helper functions, and call one or the other depending on the value of `layer_wise`? 

(Also, if I understand right, the motivation for this PR was to reduce repeated code, but it seems like ultimately there wasn't that much repeated code? I think it makes sense to combine the two DoG functions, it just isn't clear to me that we're gaining much in terms of code reuse.)

optax/contrib/_dog_test.py
```diff
@@ -0,0 +1,142 @@
...
+    # Layer-wise DoG
+    scaler_layer = dog.scale_by_dog(
+        init_step=("heuristic", 1e-6), layer_wise=True
+    )
+    state_layer = scaler_layer.init(params)
+    updates_layer, _ = scaler_layer.update(
+        self.per_step_updates, state_layer, params
+    )
+
+    # They should be different because the updates are non-uniform across layers
```
emilyfertig: Could you condense or remove this comment?

emilyfertig (commented):
Thanks!

optax/_src/transform.py
abheeyeee: https://github.com/google-deepmind/optax/pull/1518#issuecomment-3603552666
can you take a look earlier it wasnt inside contrib folder.

optax/contrib/_dog_test.py
abheeyeee: i have removed the comments

abheeyeee:
@emilyfertig I made those changes that you prefered I added the new scale_by_l_dog function. [137d6a7](https://github.com/google-deepmind/optax/pull/1518/commits/137d6a7547ed3ecce58e17f7590de435cae38275)
commit is not visible, i mistakenly did rebased or reset my branch after making it while trying to fix the merge conflict. 
Would really love your feedback. thanks

emilyfertig (commented):
Thanks! `scale_by_l_dog` should be the same as `scale_by_dog` with `layer_wise = True`, right? Can we remove the `layer_wise` arg, and make `scale_by_l_dog` the same as `scale_by_dog` with `layer_wise=True`?

abheeyeee:
> Thanks! `scale_by_l_dog` should be the same as `scale_by_dog` with `layer_wise = True`, right? Can we remove the `layer_wise` arg, and make `scale_by_l_dog` the same as `scale_by_dog` with `layer_wise=True`?

Thanks for your feedback @emilyfertig i did as you asked made scale_by_l_dog the same as scale_by_dog with layer_wise=True.
But i am facing some Checks faiure of Pytest versions on ubuntu jax, and i tried solving it but one passes and another fails. can you help me out here. Otherwise all changes have been made and this should fix the issue

optax/contrib/_dog.py
```diff
@@ -39,8 +41,48 @@ class DoGState(NamedTuple):
 
   is_init_step: jax.Array  # bool
   init_params: chex.ArrayTree
-  max_dist: jax.Array
-  sum_sq_norm_grads: jax.Array
+  max_dist: chex.ArrayTree
+  sum_sq_norm_grads: chex.ArrayTree
+
```
emilyfertig, line 45: Could you leave just two lines between definitions?

optax/contrib/_dog.py
```diff
@@ -39,8 +41,48 @@ class DoGState(NamedTuple):
...
+  .. versionadded:: 0.2.3
+
+  .. warning::
+    The authors recommend using model averaging with this optimizer.
+  """
+  return _scale_by_dog(
+      init_step=("heuristic", reps_rel),
+      eps=eps,
+      layer_wise=True,
```
emilyfertig: Sorry, what I meant is to please get rid of the `layer_wise` arg everywhere, and make separate implementations of `scale_by_dog` and `scale_by_l_dog`. Does that make sense?

emilyfertig (changes requested):
Thanks! I think the pytest failure is unrelated and should clear up if you rebase.

optax/contrib/_dog.py
abheeyeee: @emilyfertig Refactored the DoG optimizer implementation in optax/contrib/_dog.py to separate the global and layer-wise variants.
 Refactored optax/contrib/_dog.py: Removed the internal _scale_by_dog helper function.
Implemented scale_by_dog (global DoG) and scale_by_l_dog (layer-wise DoG) as distinct, standalone functions.
Removed the layer_wise argument from scale_by_dog to enforce clear separation of concerns.

Updated optax/contrib/_dog_test.py:
Renamed test_dog_layer_wise to test_l_dog_vs_dog to reflect the API changes.
Updated comments to remove outdated references to the layer_wise argument.

Verified that all tests pass with pytest optax/contrib/_dog_test.py.

optax/contrib/_dog.py
```diff
@@ -39,13 +41,98 @@ class DoGState(NamedTuple):
...
-  max_dist: jax.Array
-  sum_sq_norm_grads: jax.Array
+  max_dist: chex.ArrayTree
+  sum_sq_norm_grads: chex.ArrayTree
+
+
+def scale_by_l_dog(
+    reps_rel: jax.typing.ArrayLike = 1e-6,
+    eps: jax.typing.ArrayLike = 1e-8,
+    param_dtype: Optional[jax.typing.DTypeLike] = None,
```
emilyfertig: Please remove the unused param_dtype arg.

optax/contrib/_dog.py
```diff
@@ -39,13 +41,98 @@ class DoGState(NamedTuple):
...
+
+  .. versionadded:: 0.2.3
+
+  .. warning::
+    The authors recommend using model averaging with this optimizer.
+  """
+
+  def init_fn(params: base.Params) -> DoGState:
+    params_dtype = optax.tree.dtype(params, "lowest")
+    if param_dtype is not None:
```
emilyfertig: Why is this done here but not in `scale_by_dog`?

optax/contrib/_dog.py
```diff
@@ -67,13 +154,9 @@ def scale_by_dog(
     This optimizer's ``init`` function should receive the actual parameters (not
     just dummy parameters) when the ``heuristic`` initial step is used.
   """
-
   init_step_type, init_step_value = init_step
 
   def init_fn(params: base.Params) -> DoGState:
-    # Define state parameters with the lowest dtype of the parameters to avoid
```
emilyfertig, line 73: Restore this comment?

optax/contrib/_dog.py
```diff
@@ -87,11 +170,14 @@ def init_fn(params: base.Params) -> DoGState:
...
+    max_dist = jnp.asarray(r_epsilon, dtype=params_dtype)
+    sum_sq_norm_grads = jnp.asarray(0.0, dtype=params_dtype)
+
     return DoGState(
         is_init_step=jnp.asarray(True),
-        init_params=params,
-        max_dist=jnp.asarray(r_epsilon, dtype=params_dtype),
-        sum_sq_norm_grads=jnp.asarray(0.0, dtype=params_dtype),
+        init_params=optax.tree.cast(params, params_dtype),
+        max_dist=max_dist,
```
emilyfertig, line 180: Leave this inlined?

optax/contrib/_dog_test.py
```diff
@@ -0,0 +1,120 @@
+# Copyright 2024 DeepMind Technologies Limited. All Rights Reserved.
```
emilyfertig: nit: 2025

optax/contrib/_dog_test.py
```diff
@@ -0,0 +1,120 @@
...
+    )
+
+    # Layer-wise DoG
+    scaler_layer = dog.scale_by_l_dog(reps_rel=1e-6)
+    state_layer = scaler_layer.init(params)
+    updates_layer, _ = scaler_layer.update(
+        self.per_step_updates, state_layer, params
+    )
+
+
```
emilyfertig: Remove extra blank line?

optax/contrib/_dog_test.py
```diff
@@ -0,0 +1,120 @@
...
+    with self.assertRaises(AssertionError):
+        test_utils.assert_trees_all_close(updates_global, updates_layer)
+
+  def test_legacy_compatibility(self):
```
emilyfertig, line 87: This test doesn't have much point if the `scale_by_distance_over_gradients` implementation is changed to call `scale_by_l_dog`. Can you revert `scale_by_distance_over_gradients` to its former implementation and still deprecate it?

abheeyeee:
@emilyfertig Understood, this is what i am going to do.
Refactor DoG and Revert Legacy Compatibility:
Proposed Changes

optax/contrib/_dog.py
Remove param_dtype argument from scale_by_l_dog.
Remove param_dtype usage in init_fn inside scale_by_l_dog if it's unused or redundant.
Address consistency between scale_by_l_dog and scale_by_dog regarding init_fn and dtype casting.
Revert scale_by_distance_over_gradients to its legacy implementation (likely a standalone implementation instead of calling scale_by_l_dog) but ensure it relies on scale_by_dog.

optax/contrib/_dog_test.py
Ensure test_legacy_compatibility is meaningful after reverting scale_by_distance_over_gradients.

optax/_src/transform.py
```diff
@@ -1372,48 +1375,63 @@ def scale_by_distance_over_gradients(
     Ivgi et al, `DoG is SGD's Best Friend: A Parameter-Free Dynamic Step Size
     Schedule <https://arxiv.org/pdf/2302.12022.pdf>`_, 2023
   """
+  reps_rel = 1e-6 if reps_rel is None else reps_rel
```
emilyfertig: This is not the original implementation. You appear to be using an LLM. Please check what it outputs before you request a review.

optax/contrib/_dog.py
emilyfertig: Please revert this so it's inlined.

optax/contrib/_dog.py
```diff
@@ -39,13 +41,94 @@ class DoGState(NamedTuple):
...
+  .. versionadded:: 0.2.3
+
+  .. warning::
+    The authors recommend using model averaging with this optimizer.
+  """
+
+  def init_fn(params: base.Params) -> DoGState:
+    params_dtype = optax.tree.dtype(params, "lowest")
+
+    # r_epsilon is already a tree of scalars
```
emilyfertig, line 79: Please remove or clarify comment

emilyfertig (changes requested):

abheeyeee:
@emilyfertig  @vroulet   All checks are passing now. Please let me know if any changes are needed. Happy to update further. Thanks

abheeyeee:
@rdyro @carlosgmartin can you review this PR and let me know, if any further changes are required.

PR #1520, opened 2025-12-03
Update make_perturbed_fun

### LINKED ISSUE #1342: Minor suggestions for make_perturbed_fun
I have some minor suggestions for [make_perturbed_fun](https://optax.readthedocs.io/en/latest/api/perturbations.html#make-perturbed-function):

- Rename `sigma` to `scale`, which is more generic and self-explanatory. Note that [Berthet et al.](https://arxiv.org/abs/2002.08676) sometimes uses $\varepsilon$ for this, instead of $\sigma$.
- Change the default value for `num_samples` from 1000 (a large arbitrary value) to 1. This is a less arbitrary value, and more common in stochastic optimization (where the objective function is itself stochastic).

If one or both of these are desirable, I can create a PR. Otherwise, this issue can be closed.

haroon0x:
FIxes #1342

- update sigma to scale
- change default num_samples to 1
- update examples notebook of perturbations


vroulet (approved):
Looks good, thanks!

PR #1536, opened 2025-12-18
Remove TensorFlow dependency and migrate mlp_mnist to Flax NNX

selamw1:
This PR removes the `TensorFlow` dependency from the `mlp_mnist` notebook and updates the code to use Flax NNX instead of `flax.linen.` It also transitions the data loading pipeline to use `grain`, aligning the example with modern JAX best practices.


rdyro (approved):
Thank you!

examples/mlp_mnist.ipynb
```diff
@@ -25,14 +24,15 @@
...
         "import jax\n",
         "import jax.numpy as jnp\n",
         "import optax\n",
         "import numpy as np\n",
-        "import tensorflow as tf\n",
-        "import tensorflow_datasets as tfds"
+        "from flax import nnx\n",
+        "\n",
+        "import grain.python as pygrain\n",
+        "from torchvision.datasets import MNIST\n",
```
vroulet, line 34: Why are we using torchvision? Would be best to have either only grain or grain and tensorflow-datasets (if possible avoid tensorflow because of versioning issues).

rdyro: `torchvision` as a source for the dataset might be correct as it's a widely used and a well maintained package. This also show that optax (and jax) can be highly interoperable with the existing ecosystem.

vroulet:
@selamw1, sorry for the delay, we chatted offline with Robert about using or not torchvisions.
Ideally, we would like the same setup as in #1568, i.e. TFDS + grain (for image transformations). No torchvision

PR #1538, opened 2025-12-25
feat: extend Muon optimizer to support Optax-style callable masks

### LINKED ISSUE #1537: Extend Muon optimizer to support callable masks in weight_dimension_numbers
I found a TODO in optax/contrib/_muon.py around the scale_by_shape and scale_by_muon update functions indicating that callable weight_dimension_numbers isn’t supported via _masking._mask_callable.

  While adding _mask_callable support, I noticed another subtle behavior that must be preserved:
  weight_dimension_numbers=None at the root means “apply default MuonDimensionNumbers to all parameters” (per the docstring), whereas None at leaf level means “skip this parameter.” This distinction must remain
  true after adding callable support.

  Correct Scope

  - Add _masking._mask_callable handling for weight_dimension_numbers.
  - After resolving callables, if the result is None at the root, broadcast MuonDimensionNumbers() across the update tree.
  - Leaf-level None should still act as “skip,” and MaskedNode should be treated as identity.

  Expected Behavior

  - Callable masks should be accepted (like other Optax transforms).
  - Root-level None still implies “apply to all 2D params.”
  - Leaf-level None and MaskedNode remain no‑ops.

  If you want, I can post this and align the PR with that behavior.

yash194:
## Description
This PR implements support for Optax-style callable masking in the Muon optimizer, resolving the TODOs noted in `optax/contrib/_muon.py` around the `scale_by_shape` and `scale_by_muon` update functions.

## What changed
* Added `_masking._mask_callable` handling for `weight_dimension_numbers`.
* Safely handle `_masking.MaskedNode` by treating masked leaves as identity updates.
* Fixed `adaptive=True` behavior so masked parameters are not accidentally rescaled by their own norms.
* Ensured callable masks behave consistently with other Optax transforms.

## Why
Previously, callable `weight_dimension_numbers` were treated as raw spec functions and not normalized via Optax’s masking utilities. This could lead to masked parameters still being transformed, and in the `adaptive=True` case, masked parameters could be incorrectly scaled by their own norm.

### Important semantic detail: handling of None
While adding callable support, care was taken to preserve the existing semantics of `None` in `weight_dimension_numbers`:
* `None` at the root level continues to mean “apply default `MuonDimensionNumbers()` to all eligible parameters,” as documented.
* `None` at the leaf level continues to mean “skip this parameter.”
* `_masking.MaskedNode` is treated as identity and never transformed.

Preserving this distinction is important for correct partitioning behavior and ensures no regressions when introducing callable mask support.

## Related Issues
Fixes #1537


rdyro:
I don't believe this PR is necessary (please reopen if that's not the case), see my comments here: [https://github.com/google-deepmind/optax/issues/1537](https://github.com/google-deepmind/optax/issues/1537)

yash194:
Understood! Thanks for clarifying that the TODO was stale. I saw it and thought it was an active gap, but since the current implementation covers the intended cases, I agree this isn't needed. THANKS !

PR #1439, opened 2025-10-24
Fix _projections_test.py: Remove prints and use optax.tree.allclose.

carlosgmartin:
Follow-up to #1351. @rdyro

PR #1351, opened 2025-06-21
Add more projection functions.

carlosgmartin:
Partially addresses https://github.com/google-deepmind/optax/issues/1280. Adds the following functions to the [projections](https://optax.readthedocs.io/en/latest/api/projections.html) module:
- `projection_vector`
- `projection_hyperplane`
- `projection_halfspace`

Issue #1280, opened 2025-04-27
Add missing projections from jaxopt


carlosgmartin:
Related: https://github.com/google-deepmind/optax/issues/977

The following projections are [present in jaxopt](https://jaxopt.github.io/stable/constrained.html#projections) but missing in optax:

- [ ] [sparse_simplex](https://jaxopt.github.io/stable/_autosummary/jaxopt.projection.projection_sparse_simplex.html#jaxopt.projection.projection_sparse_simplex)
- [ ] [hyperplane](https://jaxopt.github.io/stable/_autosummary/jaxopt.projection.projection_hyperplane.html#jaxopt.projection.projection_hyperplane)
- [ ] [halfspace](https://jaxopt.github.io/stable/_autosummary/jaxopt.projection.projection_halfspace.html#jaxopt.projection.projection_halfspace)
- [ ] [affine_set](https://jaxopt.github.io/stable/_autosummary/jaxopt.projection.projection_affine_set.html#jaxopt.projection.projection_affine_set)
- [ ] [polyhedron](https://jaxopt.github.io/stable/_autosummary/jaxopt.projection.projection_polyhedron.html#jaxopt.projection.projection_polyhedron)
- [ ] [box_section](https://jaxopt.github.io/stable/_autosummary/jaxopt.projection.projection_box_section.html#jaxopt.projection.projection_box_section)
- [ ] [transport](https://jaxopt.github.io/stable/_autosummary/jaxopt.projection.projection_transport.html#jaxopt.projection.projection_transport)
- [ ] [birkhoff](https://jaxopt.github.io/stable/_autosummary/jaxopt.projection.projection_birkhoff.html#jaxopt.projection.projection_birkhoff)

I can work on some of these.

aymuos15:
I am interested in making contributions (and learning as I have only recently started becoming adept with jax in general) to this. 

So could someone potentially outline the skeleton for the best way to approach this?

Thanks!

carlosgmartin:
@aymuos15 To see how Optax currently implements its [projection](https://optax.readthedocs.io/en/latest/api/projections.html) functions, see the [projections](https://github.com/google-deepmind/optax/blob/main/optax/projections/_projections.py) module.

[...]

PR #1545, opened 2026-01-03
Add sparse simplex, transport, and Birkhoff projections from JAXopt

aymuos15:
I have set it up so one can feed the solver for Transport/Birkhoff projections via make_solver argument. @carlosgmartin for this PR is this enough (as the tests are namesake for these) or should I only port the sparse_simplex for now?

Continuation of https://github.com/google-deepmind/optax/issues/1280, https://github.com/google-deepmind/optax/pull/1351 and https://github.com/google-deepmind/optax/pull/1439

### DIFF
--- a/docs/api/projections.rst
+++ b/docs/api/projections.rst
@@ -31,8 +31,13 @@ For instance, here is an example how we can project parameters to the non-negati
 Available projections
 ~~~~~~~~~~~~~~~~~~~~~
 .. autosummary::
+    kl_projection_birkhoff
+    kl_projection_transport
+    projection_birkhoff
     projection_box
+    projection_halfspace
     projection_hypercube
+    projection_hyperplane
     projection_l1_ball
     projection_l1_sphere
     projection_l2_ball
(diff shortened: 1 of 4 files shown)


carlosgmartin:
Thanks!

I'd ask one of the project maintainers, perhaps @rdyro, about the desired API.

rdyro (approved):
Thanks!

vroulet:
It looks pretty good to me. Just to be sure @aymuos15 : 
- what changes did you do compared to jaxopt? None?
- did you take all tests?
- did you check if you kept optax formatting of e.g. docstrings? I think references are not well formatted (compare to optimizers). You can build the docs locally to see if the docs look good.

Thanks again!

aymuos15:
I did just push a fix for the docstrings. Thank you very much for pointing that out. 

@vroulet For simplex, I have made sure we are matching everything. For transport and birkhoff, they are usable with an outside solver.

vroulet:
Ok, would it be possible to make two separate PRs? One for the sparse simplex projection and one for the transport/birkhoff projections.
This one can be the sparse simplex.

The problem with the transport/birkhoff is that (i) it requires a solver, (ii) the api for the solver is not documented and it takes a specific form like `solver.run(...).params`. Such api choices need to be documented. If we add these projections I would actually use a default solver rather than letting the user select. Right now, there is no "solver" in optax that could fit the requirements

aymuos15:
Okay, that's a very fair argument. Thank you. I will change this to just the sparse simplex. Instead of raising another PR right away, should I first raise a separate issue to discuss what the default solver should be (or continue the convo in one of the other issues)?

vroulet:
@aymuos15 you may raise the issue and open a discussion with maybe @carlosgmartin if he's interested. 

For now, just prune this PR from the Birkhoff and transport projections. I'm happy to merge after that.

Thanks again!

PR #1555, opened 2026-01-07
Add edge-case tests for warmup_cosine_decay_schedule

edawite:
Adds edge-case coverage for warmup_cosine_decay_schedule

### DIFF
--- a/optax/schedules/_schedule_test.py
+++ b/optax/schedules/_schedule_test.py
@@ -559,6 +559,41 @@ class WarmupCosineDecayTest(parameterized.TestCase):
       for t in range(warmup_steps, decay_steps + 1):
         self.assertLessEqual(float(vals2[t]), float(vals[t]) + 1e-12)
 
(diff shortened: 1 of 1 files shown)


selamw1:
Thanks!

The existing test `test_raises_when_decay_equals_warmup` explicitly asserts that `ValueError` is raised if `decay_steps == warmup_steps`.
If the `test_decay_smaller_than_warmup_defined_or_raises` is intended to pass by catching that `ValueError`, then the test is redundant.
If the intention is to allow `decay_steps < warmup_steps` (e.g., training stops during warmup), then the implementation needs to change, and the existing `test_raises_when_decay_equals_warmup` needs to be updated.

vroulet:
@edawite you are not giving any context. I am going to close this PR.
Feel free to reopen if you can give more context later.

Thanks again @selamw1 ! You have an interesting point on letting training stops during warmup but I think it can be achieved more easily.

PR #1562, opened 2026-01-13
implements Mars optimizer [contrib]

### LINKED ISSUE #1561: [Feature Request] Adding the MARS Optimizer (Variance Reduction) from Hu et al. 2024
Hey! I’ve been reading through the recent paper "MARS: Unleashing the Power of Variance Reduction for Training Large Models" (Hu et al., 2024). It proposes a really interesting corrected momentum estimator to handle the gradient variance issues we usually see when training LLMs.I noticed Optax doesn't have this implemented yet, so I took a crack at designing a JAX-compliant prototype myself. I think it would be a great addition to optax.contrib.Here is how I’ve set it up so far:State Management: I used a MarsState NamedTuple to track prev_grad needed for the correction term.The Logic: I implemented the variance reduction term: $c_t = \gamma \frac{\beta_1}{1 - \beta_1} (g_t - g_{t-1})$.Stability: As per Section 3.2 of the paper, I applied global norm clipping specifically to the correction term—this seems critical for keeping things stable.Edge Cases: I handled the $t=0$ step implicitly so the correction term vanishes at the start (preventing initial spikes).

I’ve got the code ready to go—would you be open to a PR for this?

Reference Link : https://arxiv.org/abs/2411.10438

KumarADITHYA123:
Resolves #1561 .

This PR implements the MARS optimizer in optax.contrib. It introduces 
scale_by_mars
 which applies a correction term based on the difference between current and previous gradients to reduce variance.

Key Implementation Details:

Logic: Implements the correction term $c_t = \gamma \cdot \frac{\beta_1}{1 - \beta_1} (g_t - g_{t-1})$.

Stability: Supports "max_norm" clipping applied specifically to the correction term (as opposed to the global gradient), which is critical for stability.

Edge Case: Handles $t=0$ by implicitly masking the correction term.

Verification:Added  "optax/contrib/_mars_test.py"  with tests for shapes, convergence on quadratic objectives, and the $t=0$ edge case.

Verified that setting gamma=0 correctly recovers standard AdamW behavior.

### DIFF
--- a/optax/contrib/_mars.py
+++ b/optax/contrib/_mars.py
@@ -23,9 +23,9 @@ import jax.numpy as jnp
 from optax import tree
 from optax._src import base
 from optax._src import combine
+from optax._src import linear_algebra as la
 from optax._src import numerics
 from optax._src import transform
-from optax._src import utils
 
 __all__ = ["mars_adamw", "scale_by_mars", "MarsState"]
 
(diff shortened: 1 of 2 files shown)


KumarADITHYA123:
Apologies for the noisy commit history, I hit a few environment snags. 
Happy to squash these into a single commit if preferred!

vroulet:
Thanks!
Can you contact the authors of the algorithm to check with them that your implementation is faithful? We do not want to misrepresent their work.
Also can you add the optimizer to the list of optimizers tested in _common_test.py in contrib? 
By doing that you may check if some of your tests are redundant with the ones in _common_test.py

PR #1574, opened 2026-01-23
adding additional non 2d array test in galore (refrence to #1541)

yash194:
adding the additional tests for the galore optimizer as determined in the pr #1541


optax/contrib/_galore_test.py
```diff
@@ -23,6 +23,7 @@
 from optax._src import transform
 from optax._src import update
 from optax.contrib import _galore
+import optax
```
vroulet, line 26: You should not need to make this general "optax" import. Try importing adequate subpackages

vroulet: Nevermind I'll take care of it internally.

PR #1581, opened 2026-01-25
Add madgrad optimizer

### LINKED ISSUE #1582: [Feature Request] Add MADGRAD Optimizer
I would like to propose adding the MADGRAD optimizer (Defazio & Jelassi, 2021).

It is an adaptive method that bridges the gap between SGD and Adam.

I have explained and implemented this in PR #1581. All CI checks have been passed. Thanks.

divye-joshi:
## Description

Closes #1582

This PR implements the MADGRAD optimizer in optax.contrib, porting the mathematical operations from the official PyTorch implementation to JAX.

MADGRAD is a general-purpose "Dual Averaging" optimizer that aims to bridge the gap between adaptive methods (like Adam) and stochastic gradient descent (SGD). It is designed to match the fast convergence of adaptive methods on NLP tasks while maintaining the superior generalization of SGD on vision tasks.

## References

- **Paper:** [Adaptivity without Compromise: A Momentumized, Adaptive, Dual Averaged Gradient Method for Stochastic Optimization](https://arxiv.org/abs/2101.11075) (Defazio & Jelassi, 2021)

- **Source:** Ported faithfully from the [official PyTorch implementation](https://github.com/facebookresearch/madgrad).

## Technical Overview

MADGRAD computes updates using a weighted sum of gradients relative to the *initial* parameters ($x_0$), rather than just the previous step. The implementation follows the Dual Averaging formulation:

$$\lambda_k = \text{lr} \cdot \sqrt{k+1}$$

> **Step Size Scaling:** The effective step size scales up with the square root of the iteration count ($k$), ensuring older gradients don't vanish too quickly.

$$s_{k+1} = s_k + \lambda_k \cdot g_k$$

> **Gradient Accumulator:** Tracks the weighted sum of all past gradients (conceptually similar to the "velocity" or momentum vector).

$$G_{k+1} = G_k + \lambda_k \cdot g_k^2$$

> **Adaptive Normalizer:** Tracks the weighted sum of squared gradients to normalize updates, preventing the optimizer from getting stuck in steep regions (similar to Adagrad).

$$z_{k+1} = x_0 - \frac{s_{k+1}}{(G_{k+1})^{1/3} + \epsilon}$$

> **Dual Update:** Computes the new "target" parameter $z$ by applying the accumulated gradient $s$ (normalized by the cube root of $G$) directly to the initial weights $x_0$.

$$x_{k+1} = (1 - \text{momentum}) z_{k+1} + \text{momentum} \cdot x_k$$

> **Final Update:** The actual parameter value is a momentum-weighted blend of the new target $z$ and the previous parameter $x$.


optax/contrib/_madgrad_test.py
rdyro, line 1: This file is probably unnecessary since this is not a meta-optimizer and you're testing in common_tests anyway. Delete it please

optax/contrib/_madgrad.py
```diff
@@ -0,0 +1,164 @@
...
+    # Ensure stability by adding eps to the learning rate, matching the
+    # official PyTorch implementation.
+    lr_stable = lr + eps
+
+    # lamb = lr * sqrt(k + 1)
+    lamb = lr_stable * jnp.sqrt(count + 1)
+
+    # Convert lamb to the dtype of the updates to avoid implicit promotion.
+    # This prevents errors where float16 parameters are promoted to float32.
+    lamb_tree = jax.tree.map(
```
rdyro: Do not create the lamb_tree, instead, every time you use lam in an expression cast the result to the original type:
```py
(lam * (g ** 2)).astype(g.dtype)
```

divye-joshi:
Thanks for the review @rdyro! 

I have addressed both points:

Deleted the redundant _madgrad_test.py since _common_test.py covers it.

Removed the intermediate lamb_tree creation and switched to inline .astype(g.dtype) casting as suggested.

Please let me know if any more changes need to done. thanks.

PR #1604, opened 2026-02-19
Adahessian optimizer and refactor Hutchinson estimator as shared utility to be used by Adahessian, Sophia optimizer

viralvgupta:
## Summary
This PR implements the AdaHessian optimizer in contrib. It also refactors the Hutchinson estimator into a shared utility and wires both Sophia and AdaHessian to use it, while keeping the existing public APIs intact. In addition, it adds AdaHessian tests to contrib.

## Motivation
Sophia and AdaHessian rely on the same Hutchinson estimator, but the implementation was duplicated. This made maintenance more error-prone and led to subtle drift (e.g., multi-sample support only existing in AdaHessian). Centralizing the estimator makes the behavior consistent, reduces duplication, and provides a single place to evolve the estimator logic.

## Algorithm details
AdaHessian approximates second‑order scaling using only the **diagonal of the Hessian**. At each step:

1) **Gradient + momentum**
- Compute gradient: `g_t = ∇_θ L(θ_t)`
- First moment (momentum): `m_t = β1 m_{t-1} + (1-β1) g_t`
- Bias‑corrected moment: `m̂_t = m_t / (1 - β1^t)`

2) **Hessian diagonal (Hutchinson)**
- Sample Rademacher probes `v` (entries in {−1, +1})
- Compute Hessian‑vector product via JVP: `Hv`
- Estimate diagonal: `d_t ≈ v ⊙ (H v)`
- Optionally average over `n_samples` probes
- The Hessian diagonal is recomputed every `update_interval` steps; otherwise the cached value is reused.

3) **Second‑order EMA**
- Track squared diagonal with EMA:
  - `ν_t = β2 ν_{t-1} + (1-β2) (d_t)^2`
  - `ν̂_t = ν_t / (1 - β2^t)`

4) **Scaling and update**
- Per‑parameter denominator:
  - `denom = (ν̂_t)^(hessian_power / 2) + ε`
- Parameter update:
  - `Δθ_t = m̂_t / denom`
  - `θ_{t+1} = θ_t - lr * Δθ_t`

This PR does **not** change AdaHessian defaults; it only centralizes the Hutchinson estimator and tunes **test‑only** hyperparameters for convergence in the common test suite.

### 1) Shared Hutchinson estimator
- Added `optax/contrib/_hutchinson.py` with:
  - `HutchinsonState`
  - `hutchinson_estimator_diag_hessian(random_seed=None, n_samples=1)`
- The estimator:
  - Draws Rademacher probe vectors
  - Uses JVP of `grad(obj_fn)` to compute Hessian-vector products
  - Computes `v ⊙ (H v)` and averages across samples for variance reduction

### 2) Sophia / AdaHessian use the shared utility
- Sophia’s `hutchinson_estimator_diag_hessian` now forwards to the shared utility with `n_samples=1`
- AdaHessian’s version forwards to the shared utility with `n_samples` passthrough
- Both preserve existing signatures and behavior

### 3) contrib re-exports
- `optax.contrib.hutchinson_estimator_diag_hessian` and `HutchinsonState` now re-exported from the shared module
- AdaHessian symbols (`adahessian`, `scale_by_adahessian`, `AdaHessianState`) added to `optax.contrib.__init__` so tests and users can import them from `optax.contrib`

### 4) Tests / convergence
- AdaHessian added to contrib test suite (`_common_test.py`) and uses `obj_fn` similarly to Sophia
- Tuned AdaHessian test-only hyperparameters:
  - `hessian_power=0.25`
  - (keeps `learning_rate=1e-2`, `update_interval=10`)
- This makes Rosenbrock convergence pass in the common test suite without any skip

### 5) Documentation / comments
- Added clearer, top‑level comments in AdaHessian update logic and Hutchinson utility to explain estimator flow and stability steps
- Added comments in AdaHessian tests to clarify expectations

## Behavioral considerations
- No change to public Sophia/AdaHessian API signatures
- Hutchinson estimator is now defined in a shared module; behavior is equivalent to the previous implementations
- `n_samples` support remains available (now shared)

## Testing
- `pytest optax/contrib -vv`

## Notes / follow‑ups
- If desired, we can later revisit AdaHessian default `hessian_power` (currently unchanged in implementation; only the test suite uses `0.25`).

### DIFF
--- a/optax/contrib/_adahessian.py
+++ b/optax/contrib/_adahessian.py
@@ -38,15 +38,17 @@ class AdaHessianState(NamedTuple):
   hessian_fn_state: Any
 
 
-def _average_conv_kernel_hessian(hessian_diag):
-  """Average per-kernel Hessian values across spatial dims for conv weights."""
-  def maybe_average(h):
+def _postprocess_hessian_diag(hessian_diag):
+  """Apply PyTorch AdaHessian-style postprocessing to Hessian estimates."""
+  def maybe_postprocess(h):
+    if h.ndim <= 2:
+      return jnp.abs(h)
     if h.ndim == 4:
       mean = jnp.mean(jnp.abs(h), axis=(2, 3), keepdims=True)
       return jnp.ones_like(h) * mean
     return h
 
-  return jax.tree.map(maybe_average, hessian_diag)
+  return jax.tree.map(maybe_postprocess, hessian_diag)
 
 
 def scale_by_adahessian(
@@ -100,8 +102,8 @@ def scale_by_adahessian(
           updates, hess_fn_state, params=params, **hess_fn_kwargs
       )
       if average_conv_kernel:
-        # Stabilize conv kernels by averaging across spatial dimensions.
-        hessian_diag = _average_conv_kernel_hessian(hessian_diag)
+        # Match PyTorch AdaHessian postprocessing for tensor rank branches.
+        hessian_diag = _postprocess_hessian_diag(hessian_diag)
       return hess_fn_state, hessian_diag
 
     # Recompute the Hessian diagonal periodically, otherwise reuse the cached
@@ -168,8 +170,8 @@ def adahessian(
     update_interval: Interval for updating the Hessian diagonal estimate.
     weight_decay: Strength of decoupled weight decay regularization.
     weight_decay_mask: mask controlling which params receive weight decay.
-    average_conv_kernel: If True, average Hessian values across spatial
-      dimensions for 4D convolution kernels.
+    average_conv_kernel: If True, apply AdaHessian-style postprocessing:
+      abs values for rank <=2 tensors and spatial averaging for 4D kernels.
     hessian_diagonal_fn: GradientTransformation that computes the diagonal of
       the Hessian. Default is Hutchinson's estimator.
     mu_dtype: dtype of the first moment estimates.


optax/contrib/_adahessian_test.py
```diff
@@ -0,0 +1,62 @@
+# Copyright 2026 DeepMind Technologies Limited. All Rights Reserved.
```
rdyro: This file appears to be redundant since `_common_test.py` already tests for these, right?

optax/contrib/_adahessian_test.py
```diff
@@ -0,0 +1,62 @@
...
+import jax
+import jax.numpy as jnp
+
+from optax import contrib
+from optax._src import test_utils
+
+
+class AdaHessianTest(absltest.TestCase):
+
+  def test_hutchinson_estimator_quadratic_diag(self):
```
rdyro: This test is/should be in _hutchinson_test.py

optax/contrib/_hutchinson.py
```diff
@@ -0,0 +1,94 @@
...
+  def update_fn(updates, state, params=None, obj_fn=None, **extra_args):
+    # Comply with GradientTransformationExtraArgs signature; ignore extra args.
+    del extra_args, updates
+    if params is None:
+      raise ValueError("params must be provided to hutchinson update function.")
+    if obj_fn is None:
+      raise ValueError("obj_fn must be provided to hutchinson update function.")
+
+    # Split the RNG once per Hutchinson probe.
+    key, *subkeys = jax.random.split(state.key, n_samples + 1)
```
rdyro: Lines 64-91 should be done with vmap and `jnp.mean` instead of python looping

optax/contrib/_sophia.py
```diff
@@ -33,10 +33,9 @@
...
 import optax.tree
+from optax.contrib import _hutchinson
 
-
-class HutchinsonState(NamedTuple):
-  key: jax.Array
+HutchinsonState = _hutchinson.HutchinsonState
 
 
 def hutchinson_estimator_diag_hessian(random_seed: Optional[jax.Array] = None):
```
rdyro: It'd be better to use the hutchinson module member directly instead of defining a lambda alias

optax/contrib/_adahessian.py
```diff
@@ -0,0 +1,214 @@
...
+from optax._src import base
+from optax._src import combine
+from optax._src import numerics
+from optax._src import transform
+from optax._src import utils
+import optax.tree
+from optax.contrib import _hutchinson
+
+
+def hutchinson_estimator_diag_hessian(
```
rdyro: It'd be better to use the hutchinson module member directly instead of defining a lambda alias

optax/contrib/_common_test.py
```diff
@@ -322,12 +325,12 @@ def f(params_state, _):
       ):
         params = contrib.schedule_free_eval_params(state, params)
 
-    with self.subTest('Test that the optimization converges'):
-      platform = list(jax.tree.leaves(params)[0].devices())[0].platform
-      # GaLore is sensitive to the SVD accuracy and can fail on this example
-      if not (opt_name == 'galore' and platform != 'cpu'):
-        test_utils.assert_trees_all_close(
-            params, final_params, rtol=3e-2, atol=3e-2)
+      with self.subTest('Test that the optimization converges'):
```
rdyro: This appears to be wrongly re-indented

rdyro (commented):
Thanks, I left a couple of comments!

optax/contrib/_adahessian_test.py
viralvgupta: created a separate file for hutchitson tests

optax/contrib/_hutchinson.py
viralvgupta: made the changes. thanks for pointing this out

optax/contrib/_adahessian_test.py
viralvgupta: Ideally yes, I will rely on your inputs here. The main idea was to - 

AdaHessian can be initialized and run one update step with required obj_fn.
The computed updates are finite (no nan/inf).
The returned optimizer state is also finite.

but happy to remove this. I see we are checking convergence on objectives like Rosenbrock in common_tests already

rdyro: I think it should be fine to remove this file because _common_test.py should already indirectly test for those things. Thanks!

viralvgupta: done

viralvgupta:
FYI @rdyro Responded to all the comments

rdyro:
Hey, I'm rereading the implementation to merge this internally and I noticed you only included spatial averaging that the paper mentions for Conv2D, but the adahessian repo hints at other cases for spatial averaging as well.

Also, depending on which version of the implementation I'm reading, the update formula appears slightly different.
```py
hv = hv
hv = torch.abs(hv)
hv = torch.abs(hv * vi)
```
 Which reference implementation were you following? 

@viralvgupta

viralvgupta:
Thats an excellent catch and I think the code diverges from pytorch implementation for dim = 0/1/2. 

Note - even though implementation is slightly different both are accurate. I show my thoughts below. However I feel its best to stay consistent with reference pytorch implementation so I will make the changes.


Let

a_i = (H v)_i, v_i ∈ {−1, +1}

Jax code’s sample per element:

s_i = a_i. v_i

PyTorch code (that branch) uses:

t_i = |a_i|

Compare carefully:

s_i and t_i are NOT equal in general (the sign differs).

But after taking absolute value:

|s_i| = |a_i v_i|
= |a_i| |v_i|
= |a_i|
= t_i

because |v_i| = 1.

So the equivalence claim only holds under magnitude-based downstream use.

In AdaHessian, downstream uses a second-moment / square-like accumulation, so the sign disappears:

s_i² = (a_i v_i)² = a_i²

t_i² = |a_i|² = a_i²

FYI @rdyro - made the code changes

viralvgupta:
FYI @rdyro - curious if you have further questions.

rdyro:
> FYI @rdyro - curious if you have further questions.

Still working on merging it, this is on my TODO, thanks for the contribution!

PR #1610, opened 2026-02-26
Remove TensorFlow dependency in the `differentially private CNN on MNIST` example

rajasekharporeddy:
This PR removes `TensorFlow` as a dependency in `differentially_private_sgd.ipynb`, and uses only TFDS, with [grain](https://github.com/google/grain) for loading and preparing the dataset.


vroulet (commented):
Thanks!

vroulet (approved):

PR #1614, opened 2026-03-05
Rename log_prob to unnormalized_log_prob in perturbations (#1214)

### LINKED ISSUE #1214: Rename `optax.perturbations` distribution method `log_prob` to `unnormalized_log_prob`

jinukuntlaakhilakumargoud-web:
Resolves #1214.

This PR renames the log_prob method to unnormalized_log_prob in the Normal and Gumbel distribution classes under optax.perturbations, as these methods return the logarithm of the **unnormalized** probability density. This follows the naming convention used in [TensorFlow Probability](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Distribution#unnormalized_log_prob) and prevents user confusion with normalized log probabilities.

### Changes
- Renamed Normal.log_prob -> Normal.unnormalized_log_prob
- - Renamed Gumbel.log_prob -> Gumbel.unnormalized_log_prob
- - Updated docstring reference in make_perturbed_fun
- - Updated internal usage in make_perturbed_fun

### DIFF
--- a/optax/perturbations/__init__.py
+++ b/optax/perturbations/__init__.py
@@ -19,3 +19,7 @@
 from optax.perturbations._make_pert import Gumbel
 from optax.perturbations._make_pert import make_perturbed_fun
 from optax.perturbations._make_pert import Normal
+from optax.perturbations._make_pert import normal_noise_sampler
+from optax.perturbations._make_pert import unnormalized_normal_log_prob
+from optax.perturbations._make_pert import gumbel_noise_sampler
+from optax.perturbations._make_pert import unnormalized_gumbel_log_prob
--- a/optax/perturbations/_make_pert.py
+++ b/optax/perturbations/_make_pert.py
@@ -16,7 +16,8 @@
 """Creates a differentiable approximation of a function with perturbations."""
 
 
-from typing import Callable
+from typing import Callable, Optional
+import warnings
 
 import jax
 import jax.numpy as jnp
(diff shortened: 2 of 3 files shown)


rdyro:
Hey, thanks for the contribution, but this seems like an incomplete solution given the discussion in https://github.com/google-deepmind/optax/issues/1214

Can you argue for one of the more principled solutions discussed in the thread? It seems like a much better approach.

jinukuntlaakhilakumargoud-web:
@rdyro Thanks for the feedback! I've updated the PR to implement the function-based API discussed in #1214. make_perturbed_fun now accepts noise_sampler and noise_log_prob as separate callable arguments instead of a noise object, following vroulet's preferred approach. Added standalone functions: normal_noise_sampler, unnormalized_normal_log_prob, gumbel_noise_sampler, unnormalized_gumbel_log_prob. The Normal and Gumbel classes and the noise parameter are deprecated with warnings but remain functional for backwards compatibility. All  existing tests pass with the new API.

PR #1616, opened 2026-03-05
Remove TensorFlow dependency in `Reduce on Plateau Learning Rate Scheduler` example

rajasekharporeddy:
This PR removes `TensorFlow` as a dependency in `examples/contrib/reduce_on_plateau.ipynb`, and uses only TFDS along with [grain](https://github.com/google/grain) for loading and preparing the dataset. 

Ref #656


rdyro (approved):
Thank you!

PR #1619, opened 2026-03-06
Improve lookahead docstrings with example and usage notes

rdyro:


optax/_src/lookahead.py
```diff
@@ -89,6 +89,19 @@ def lookahead(
...
+    >>> import jax.numpy as jnp
+    >>> fast_opt = optax.sgd(1e-2)
+    >>> opt = optax.lookahead(fast_opt, sync_period=5, slow_step_size=0.5)
+    >>> params = optax.LookaheadParams.init_synced(jnp.ones((2,)))
+    >>> state = opt.init(params)
+    >>> # Calculate gradients wrt the fast parameters
+    >>> grads = jax.grad(lambda p: jnp.sum(p**2))(params.fast)
+    >>> updates, state = opt.update(grads, state, params)
+    >>> params = optax.apply_updates(params, updates)
+
```
vroulet, line 108: Add prints of loss values so that we see that for eval we need the slow params

PR #1632, opened 2026-03-18
added feat/AdaMuon in _contrib

raghulchandramouli:
## Summary
This PR implements Adamuon in Contrib, It acts as an extenstion to muon optimizer, and adds upon it

### Contribution
AdaMuon is the adaptive variant of Muon. After orthogonalizing the momentum via newton-schulz, this update is scaled by its inner product with the original (pre-orthogonalization) momentum. This inner product equals the norm (sum of singluar values), making the step size adaptive to the gradient magnitude

### Algorithm Novelty:
1. an elementwise-second momentum estimator applied to update direction of flow
2. a sign-stabilized update where momentum is a first-sign transformed before orthoganalizzation

### Guide:
Paper link : https://arxiv.org/abs/2507.11005

### DIFF
--- a/optax/contrib/_muon.py
+++ b/optax/contrib/_muon.py
@@ -772,7 +772,7 @@ def adamuon(
     - Standard Muon:  update = NS_orthogonalize(G)  ~ U V^T
     - AdaMuon:        update = <G, U V^T> * U V^T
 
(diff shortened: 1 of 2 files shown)


raghulchandramouli:
@rdyro can you take a look at this!!!

raghulchandramouli:
@emilyfertig can you please take a look at this PR, would be really nice!!!

rdyro:
This PR does not adhere to optax style. We already have `adaptive` support for muon, so making this alias is unnecessary.

PR #1644, opened 2026-03-23
Add gradient and second-order gradient tests for regression losses

dylantirandaz:
## Summary
- Add first-order gradient tests for `squared_error`, `l2_loss`, `huber_loss`, and `log_cosh`
- Add second-order gradient (Hessian) test for `log_cosh` verifying `sech^2` behavior
- Add numerical stability test for `log_cosh` gradients with large inputs
- Resolves `TODO(b/188419459)`

## Test plan
- [ ] Run `python -m pytest optax/losses/_regression_test.py`
- [ ] Run `bash test.sh`


rdyro:
This PR in effect tests JAX since correct gradients are not the scope of optax, closing.

PR #1646, opened 2026-03-23
Remove deprecated second_order functions for 0.2.9

dylantirandaz:
## Summary
- Remove `optax.second_order.hvp`, `optax.second_order.hessian_diag`, and `optax.second_order.fisher_diag`
- These were deprecated in 0.2.7 with `version_removed='0.2.9'` — now that 0.2.9 development has started, complete the scheduled removal
- Remove corresponding docs section from `docs/api/utilities.rst`
- Add CHANGELOG entry under `[Unreleased]`

## Test plan
- [ ] Run `bash test.sh` — no tests should reference these functions
- [ ] Verify `import optax.second_order` still works (empty subpackage)


rdyro:
Please don't change the CHANGELOG, if we're removing the second_order module, the file should be removed as well.

PR #1650, opened 2026-03-30
Muon: fix excessive memory from Newton-Schulz unroll=True

bzantium:
## Summary
- Fix excessive memory usage in `orthogonalize_via_newton_schulz()` caused by `jax.lax.fori_loop` with `unroll=True`
- When unrolled, XLA materializes all intermediate tensors from all NS iterations simultaneously
- Add configurable `ns_unroll` parameter (default `False`) to `orthogonalize_via_newton_schulz()`, `scale_by_muon()`, and `muon()`

## Memory Analysis

**General formula** for per-parameter memory overhead with `unroll=True`:

```
overhead = ns_steps × (2 × min(m,n)² + m×n) × element_size
```

Each NS iteration creates intermediates `A = X @ X.T` (min(m,n)²), `B = c₁·A + c₂·A²` (min(m,n)²), and the result (m×n). With `unroll=True`, all `ns_steps` iterations' tensors coexist in memory.

| Weight Shape | dtype | Per-param overhead | 28 layers (typical LLM) |
|---|---|---|---|
| (2048, 6144) | bf16 | ~92 MB | ~2.6 GB |
| (4096, 11008) | bf16 | ~336 MB | ~9.4 GB |
| (8192, 22016) | bf16 | ~1.3 GB | ~37 GB |

With `ns_unroll=False` (new default), XLA reuses buffers across iterations.

## Benchmark Results

Weight shape (512, 1024), CPU:

| | `unroll=True` (old) | `unroll=False` (new) | Improvement |
|---|---|---|---|
| Step time | 255.5 ms | 157.2 ms | **38% faster** |
| Compiled HLO size | 57.4 KB | 27.8 KB | **52% smaller** |
| Numerical result | identical | identical | no difference |

Training a 2-layer MLP (32→64→1) on synthetic regression data for 500 steps:
- `ns_unroll=True` and `ns_unroll=False` produce **numerically identical** loss curves
- Both converge successfully — no side effects from the change

## Test plan
- [x] `test_ns_unroll_produces_identical_results` — parameterized across 3 shapes × 2 dtypes
- [x] `test_muon_ns_unroll_produces_identical_results` — parameterized across 3 shapes
- [x] `test_muon_unroll_equivalence` — training benchmark showing identical convergence
- [x] `test_muon_unroll_false_converges` — convergence verification with no NaN/Inf
- [x] All existing tests pass

### DIFF
--- a/optax/contrib/_muon.py
+++ b/optax/contrib/_muon.py
@@ -348,7 +348,7 @@ def orthogonalize_via_newton_schulz(
     if ns_coeffs_.ndim == 1:
       x = jax.lax.fori_loop(
           0, ns_steps, lambda i, x: _ns_iterator(i, x, ns_coeffs_), x,
-          unroll=True)  # Unroll to ensure efficient composition with jax.vmap.
+          unroll=False)
     else:
       def _scan_body(carry, coeffs_step):
         i, x = carry


rdyro:
Is it true that XLA does not reuse buffers in the unrolled case? Can you add a memory comparison analysis for an example? You can either get peak program memory from HLO stats or profile an example run.

If not unrolling is always better, perhaps we should just flip the hardcoded value instead of allowing the user to select this.

bzantium:
@rdyro Thanks for the review! Here are concrete memory measurements from a real TPU training run.

Setup: Qwen3-1.7B (2B params, 28 layers, hidden=2048, mlp=6144), v5e TPU (16GB HBM/chip), 256 chips, per_device_batch_size=1, max_target_length=4096, remat_policy=full, scan_layers=true, sa_block=512, dtype=bfloat16. Each run trained 10 steps with synthetic data using MaxText on JAX.

Compile-time memory (from XLA compiled.memory_analysis()):
```
┌───────────────────────────────────┬──────────────┬───────────┬───────────┐
│             Optimizer             │ Total Memory │ Temp Size │ Step Time │
├───────────────────────────────────┼──────────────┼───────────┼───────────┤
│ AdamW                             │ 3.3 GB       │ 3.2 GB    │ 0.54s     │
├───────────────────────────────────┼──────────────┼───────────┼───────────┤
│ Muon (unroll=True, optax default) │ 5.8 GB       │ 5.8 GB    │ 0.76s     │
├───────────────────────────────────┼──────────────┼───────────┼───────────┤
│ Muon (unroll=False)               │ 3.3 GB       │ 3.2 GB    │ 0.80s     │
└───────────────────────────────────┴──────────────┴───────────┴───────────┘
```
unroll=True increases compile-time HBM by 2.5 GB (+76%) over AdamW for this model. unroll=False brings it back down to match AdamW exactly, with a ~5% step time overhead (0.76s → 0.80s).

If you'd prefer just flipping the hardcoded value to unroll=False rather than exposing it as a user option, I'm happy to make that change.

rdyro:
Yeah, let's just do `unroll=False` please, no need for extra tests, thanks!

bzantium:
@rdyro Thank you for the feedback. I've updated the code accordingly.

bzantium:
@rdyro Just wanted to follow up in case this got buried. Would you be able to take another look when you have a moment? I appreciate your time!

Issue #1658, opened 2026-04-09
Add Self-Scaled Broyden (SSBroyden) and SSBFGS optimizers


armbrusl:
Hi! Id like to propose adding the SSBroyden and SSBFGS optimizers to `optax.contrib`.

These are full-matrix quasi-Newton methods from Urbán et al. (2025) that use a self-scaled inverse Hessian update. The paper demonstrates strong results on PINNs and similar small-to-medium scale problems.

I've ported the [PyTorch implementation from SciMBA](https://www.scimba.org/) to JAX/optax following the existing patterns (`scale_by_lbfgs` / `lbfgs`). The implementation includes:

- `ssbroyden()` and `ssbfgs()` as separate public optimizers
- A shared `scale_by_ss_quasi_newton()` core transformation
- Integration with `scale_by_zoom_linesearch` (same pattern as `optax.lbfgs`)
- Tests on quadratic and Rosenbrock functions

I have a working implementation ready. Would you be open to a PR?

**Reference:**
Urbán, J. F., Stefanou, P., & Pons, J. A. (2025). Unveiling the optimization process of physics informed neural networks: How accurate and competitive can PINNs be? *Journal of Computational Physics*, 523, 113656. [DOI](https://doi.org/10.1016/j.jcp.2024.113656)

satyam-mishra-dev:
Hi! I'd like to work on this issue. Could you assign it to me if it's still available? @rajasekharporeddy

PR #1659, opened 2026-04-10
Issue #1658 - Add SSBroyden and SSBFGS optimizers to optax.contrib 

armbrusl:
Port the Self-Scaled Broyden and Self-Scaled BFGS quasi-Newton optimizers from PyTorch (SciMBA) to JAX/optax. These maintain a dense inverse Hessian approximation with self-scaling updates, complementing the existing L-BFGS implementation for small to medium scale problems.

New public API:
- optax.contrib.ssbroyden() — SS-Broyden variant
- optax.contrib.ssbfgs() — SS-BFGS variant
- optax.contrib.scale_by_ss_quasi_newton() — shared core transform

Both integrate with scale_by_zoom_linesearch (same pattern as optax.lbfgs) and include tests on quadratic and Rosenbrock functions.

Reference: Urbán et al. (2025), Journal of Computational Physics, 523, 113656.


armbrusl:
I have added the CLA.

PR #1671, opened 2026-05-05
Repair use of deprecated parameter

NeilGirdhar:
Note: This function is untested on main.


NeilGirdhar:
Incidentally, would you reconsider #1? Maintaining downstream shims for all of these functions is becoming increasingly tedious, especially since there appears to be a fairly small upstream change that would eliminate a lot of duplicated workaround code.

At least in my case, I’ve ended up maintaining an entire layer of wrappers/shims around Optax to recover this functionality:
https://github.com/NeilGirdhar/tjax/tree/main/tjax/_src/gradient

From reading through the discussion, it seems I’m not the only one who has independently arrived at this kind of solution.

rdyro (commented):
Thanks! Could you add a small tests for this change?

rdyro:
> Incidentally, would you reconsider #1? Maintaining downstream shims for all of these functions is becoming increasingly tedious, especially since there appears to be a fairly small upstream change that would eliminate a lot of duplicated workaround code.
> 
> At least in my case, I’ve ended up maintaining an entire layer of wrappers/shims around Optax to recover this functionality: https://github.com/NeilGirdhar/tjax/tree/main/tjax/_src/gradient
> 
> From reading through the discussion, it seems I’m not the only one who has independently arrived at this kind of solution.

I'm working on a PR for this, but it's hard to make this change while passing all internal tests. This might take some time :(

NeilGirdhar:
> I'm working on a PR for this, but it's hard to make this change while passing all internal tests. This might take some time :(

Absolutely no problem!!  If this is on the horizon, and I'll one day be able to delete my shims, that makes me **extremely happy** :heart: 

By the way, as far as I can tell, my shims do pass your internal tests already--at least from the time I copied those tests.

> Thanks! Could you add a small tests for this change?

Happy to, and done!

optax/contrib/_privacy_test.py
```diff
@@ -104,6 +110,19 @@ def test_aggregated_updates_as_input_fails(self):
     with self.assertRaises(ValueError):
       dp_agg.update(mean_grads, state, self.params)
 
+  def test_dpsgd_accepts_key(self):
```
rdyro: could you parametrized on both seed `key: int = 0` and `key: jax.Array = jax.random.key(0)`

I think jax.random.key might need to called in the test scope rather than in the decorator, thanks!

NeilGirdhar: Done.

rdyro (approved):
Thanks!

rdyro (approved):
Thanks!

rdyro (approved):

PR #1688, opened 2026-06-05
fix: prevent PyPI/conda-forge dependency drift with automated release checks

### LINKED ISSUE #1594: Inconsistency in requirements (PyPI vs conda)
A version incompatibility issue came up while I was trying to release a new package (EzTaoX) on `conda-forge` that depends on optax (https://github.com/conda-forge/staged-recipes/pull/32095):
```
optax 0.2.6 has requirement jax>=0.5.3, but you have jax 0.4.31.
optax 0.2.6 has requirement jaxlib>=0.5.3, but you have jaxlib 0.4.31.
```

This appears to be caused by differing JAX dependency pins between the [conda-forge recipe](https://github.com/conda-forge/optax-feedstock/blob/9cd3c5fc0b84c214a5e9a5547aaf984ebd1d6453/recipe/meta.yaml) and the PyPI release. The latest version `v0.2.6` requires `jax>=0.5.3`. However, the conda recipe still pins jax according to `v0.2.5`, `jax>=0.4.27`.

Muneerali199:
## Description
The conda-forge optax feedstock was pinned to `jax>=0.4.27` while PyPI requires `jax>=0.5.3` (since v0.2.6), causing installation failures for conda users.

## Changes
- **pyproject.toml**: Added conda-forge to the sync checklist comment on jax pins
- **.github/workflows/pypi-publish.yml**: New release workflow step that fetches the conda-forge recipe and warns when jax dependency bounds diverge from PyPI
- **.github/workflows/tests.yml**: Updated sync comment to include conda-forge
- **RELEASE.md**: New document detailing the full release process with conda-forge sync instructions and checklist template

## How it works
On every new PyPI release, the publish action now:
1. Fetches `conda-forge/optax-feedstock/recipe/meta.yaml`
2. Parses `jax>=` from both `pyproject.toml` and the conda recipe
3. If they differ, prints a warning with exact pins and a link to open a feedstock PR

Fixes #1594


.github/workflows/pypi-publish.yml
```diff
@@ -28,6 +28,22 @@ jobs:
...
           echo "package ver. ($PACKAGE_VER) != release ver. ($RELEASE_VER)"; exit 1
         fi
+    - name: Check conda-forge dependency sync
+      run: |
+        echo "Checking conda-forge feedstock dependency bounds..."
+        CONDA_META=$(curl -s https://raw.githubusercontent.com/conda-forge/optax-feedstock/main/recipe/meta.yaml 2>/dev/null) || {
+          echo "warning: unable to fetch conda-forge recipe; skipping sync check"
+          exit 0
+        }
+        PYPI_JAX=$(grep -E '"jax>=' pyproject.toml | sed 's/.*"jax>=\([^"]*\)".*/\1/')
```
rdyro: this should be done with a toml parser, not sed

RELEASE.md
rdyro, line 1: please remove this file

.github/workflows/pypi-publish.yml
```diff
@@ -28,6 +28,22 @@ jobs:
...
+        CONDA_JAX=$(echo "$CONDA_META" | grep -E '\- jax >=' | sed 's/.*jax >=\([^"]*\)/\1/')
+        if [ -n "$PYPI_JAX" ] && [ -n "$CONDA_JAX" ] && [ "$PYPI_JAX" != "$CONDA_JAX" ]; then
```
rdyro: I'd recommend comparing versions with something like [this](https://github.com/jax-ml/jax/blob/e9fdb2865c89ddd31c0a05296e459dd55f8b9465/jax/version.py#L161)

.github/workflows/pypi-publish.yml
```diff
@@ -28,6 +28,22 @@ jobs:
         then
           echo "package ver. ($PACKAGE_VER) != release ver. ($RELEASE_VER)"; exit 1
         fi
+    - name: Check conda-forge dependency sync
+      run: |
+        echo "Checking conda-forge feedstock dependency bounds..."
+        CONDA_META=$(curl -s https://raw.githubusercontent.com/conda-forge/optax-feedstock/main/recipe/meta.yaml 2>/dev/null) || {
+          echo "warning: unable to fetch conda-forge recipe; skipping sync check"
+          exit 0
```
rdyro: this should fall through to one place where the warning gets printed, not two separate warnings

.github/workflows/pypi-publish.yml
```diff
@@ -28,6 +28,22 @@ jobs:
...
+        }
+        PYPI_JAX=$(grep -E '"jax>=' pyproject.toml | sed 's/.*"jax>=\([^"]*\)".*/\1/')
+        CONDA_JAX=$(echo "$CONDA_META" | grep -E '\- jax >=' | sed 's/.*jax >=\([^"]*\)/\1/')
+        if [ -n "$PYPI_JAX" ] && [ -n "$CONDA_JAX" ] && [ "$PYPI_JAX" != "$CONDA_JAX" ]; then
+          echo "warning: jax pin mismatch — PyPI: jax>=$PYPI_JAX, conda: jax>=$CONDA_JAX"
```
rdyro: maybe we can use something more visible, either `::warning` or `$GITHUB_STEP_SUMMARY`

.github/workflows/pypi-publish.yml
Muneerali199: Done — switched to tomllib/tomli with  for parsing.

RELEASE.md
Muneerali199: Done - RELEASE.md has been removed.

.github/workflows/pypi-publish.yml
Muneerali199: Done - using packaging.version.Version for semantic comparison.

.github/workflows/pypi-publish.yml
Muneerali199: Done - all warnings flow through a single WARN helper with sys.exit(0) as the single fallthrough path.

.github/workflows/pypi-publish.yml
Muneerali199: Done - switched to ::warning :: annotations for GitHub Actions visibility.

scripts/check_conda_sync.py
```diff
@@ -0,0 +1,68 @@
...
+            "https://raw.githubusercontent.com/"
+            "conda-forge/optax-feedstock/main/recipe/meta.yaml"
+        )
+        req = urllib.request.urlopen(url, timeout=10)
+        conda_meta = req.read().decode()
+    except Exception as e:
+        warn(f"unable to fetch conda-forge recipe ({e}); skipping sync check")
+        sys.exit(0)
+
+    m = re.search(r"^\s*-\s+jax\s*>=\s*([\w.]+)", conda_meta, re.MULTILINE)
```
rdyro: this is just regex in python, why not parse the yaml file with a yaml library?

scripts/check_conda_sync.py
```diff
@@ -0,0 +1,68 @@
...
+    if not m:
+        warn("could not find jax pin in conda-forge recipe")
+        sys.exit(0)
+    conda_jax = m.group(1)
+
+    # 3. Compare
+    if Version(jax_pin) != Version(conda_jax):
```
rdyro: this logic is wrong because it doesn't take into account whether it's a `<=` `==` or `>=`, right?

scripts/check_conda_sync.py
```diff
@@ -0,0 +1,68 @@
+"""Check that PyPI and conda-forge jax dependency pins are in sync."""
```
rdyro: please don't add a new top-level directory, this script could be under `.github`

scripts/check_conda_sync.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # 2. Fetch conda-forge recipe
+    try:
+        url = (
+            "https://raw.githubusercontent.com/"
+            "conda-forge/optax-feedstock/main/recipe/meta.yaml"
+        )
+        req = urllib.request.urlopen(url, timeout=10)
+        conda_meta = req.read().decode()
+    except Exception as e:
```
rdyro: can you use a narrow network exception?

scripts/check_conda_sync.py
```diff
@@ -0,0 +1,68 @@
...
+
+def main() -> None:
+    # 1. Parse pyproject.toml
+    with open("pyproject.toml", "rb") as f:
+        pyproject = tomllib.load(f)
+
+    jax_pin = None
+    for dep in pyproject.get("project", {}).get("dependencies", []):
+        m = re.match(r"jax\s*(>=|==|)\s*([\w.]+)", dep)
+        if m:
```
rdyro: you can use the walrus operator here

scripts/check_conda_sync.py
```diff
@@ -0,0 +1,68 @@
+"""Check that PyPI and conda-forge jax dependency pins are in sync."""
+
+import re
+import sys
+import urllib.request
+
+try:
```
rdyro: don't guard toml import, just ensure the python running this script is at least 3.11

scripts/check_conda_sync.py
```diff
@@ -0,0 +1,68 @@
...
+        req = urllib.request.urlopen(url, timeout=10)
+        conda_meta = req.read().decode()
+    except Exception as e:
+        warn(f"unable to fetch conda-forge recipe ({e}); skipping sync check")
+        sys.exit(0)
+
+    m = re.search(r"^\s*-\s+jax\s*>=\s*([\w.]+)", conda_meta, re.MULTILINE)
+    if not m:
+        warn("could not find jax pin in conda-forge recipe")
+        sys.exit(0)
```
rdyro: you're inside a function, you can just return

scripts/check_conda_sync.py
```diff
@@ -0,0 +1,68 @@
...
+    # 3. Compare
+    if Version(jax_pin) != Version(conda_jax):
+        warn(f"jax pin mismatch - PyPI: {jax_pin}, conda: {conda_jax}")
+        print(
+            "Please open a PR at "
+            "https://github.com/conda-forge/optax-feedstock "
+            f"to update recipe/meta.yaml: jax >={jax_pin}, "
+            f"jaxlib >={jax_pin}"
+        )
+        sys.exit(0)
```
rdyro: this line does nothing, remove it

scripts/check_conda_sync.py
Muneerali199: Done - using yaml.safe_load after stripping Jinja2 tags.

scripts/check_conda_sync.py
Muneerali199: Done - now extracts and compares both operator and version.

scripts/check_conda_sync.py
Muneerali199: Done - moved to .github/check_conda_sync.py.

scripts/check_conda_sync.py
Muneerali199: Done - using urllib.error.URLError.

scripts/check_conda_sync.py
Muneerali199: Done - used := for match pattern.

scripts/check_conda_sync.py
Muneerali199: Done - just import tomllib directly (Python 3.11+).

scripts/check_conda_sync.py
Muneerali199: Done - using return instead of sys.exit(0) throughout.

scripts/check_conda_sync.py
Muneerali199: Done - changed to return.

.github/check_conda_sync.py
```diff
@@ -0,0 +1,85 @@
...
+
+def extract_pin(dep: str) -> tuple[str, str] | None:
+    if m := re.match(r"jax\s*(>=|==|<=|!=|~=)\s*([\w.]+)", dep):
+        return m.group(1), m.group(2)
+    return None
+
+
+def main() -> None:
+    # 1. Parse pyproject.toml
+    with open(REPO_ROOT / "pyproject.toml", "rb") as f:
```
rdyro, line 32: why "rb" instead of "r"?

rdyro (approved):
Thanks!

.github/check_conda_sync.py
Muneerali199: actually It's required by tomllib.load() — it only accepts binary mode.

PR #1692, opened 2026-06-09
contrib: add SOAP optimizer (arXiv:2409.11321)

irhyl:
SOAP (Improving and Stabilizing Shampoo using Adam) combines Shampoo's Kronecker-factored preconditioning with Adam's per-coordinate adaptivity. For each 2-D weight matrix it maintains left/right Kronecker factors whose eigenbases define a rotation; Adam's moment buffers are tracked in that rotated space and the final update is unprojected back.  Parameters with fewer than two dimensions fall back to standard Adam.

New files:
optax/contrib/_soap.py: scale_by_soap and soap implementations
optax/contrib/_soap_test.py : 14 unit tests (shapes, orthogonality, convergence, JIT stability, Adam fallback, weight decay, mu_dtype, frequency control)

Modified files:
optax/contrib/__init__.py: export soap / scale_by_soap / ScaleBySOAPState
optax/contrib/_common_test.py: add soap to the shared optimizer test matrix and mark precondition_frequency as static

### DIFF
--- a/optax/contrib/__init__.py
+++ b/optax/contrib/__init__.py
@@ -70,6 +70,9 @@ from optax.contrib._schedule_free import schedule_free_adamw
 from optax.contrib._schedule_free import schedule_free_eval_params
 from optax.contrib._schedule_free import schedule_free_sgd
 from optax.contrib._schedule_free import ScheduleFreeState
+from optax.contrib._soap import scale_by_soap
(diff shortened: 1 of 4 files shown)


irhyl:
Hi all, just following up on this PR. I believe it's ready for review, but please let me know if there's anything else I should address. Thanks for your time!

PR #1698, opened 2026-06-11
Migrate from the old `flax.linen` to the latest `flax.nnx` module in `Simple NN with Flax` example.

rajasekharporeddy:


rdyro (approved):
Thanks!

Issue #357, opened 2022-06-09
Add SPSA optimization method 


ankit27kh:
The  Simultaneous Perturbation Stochastic Approximation (SPSA) optimisation method is a faster optimisation method.

>  If the number of terms being optimized is p, then the finite-difference method takes 2p measurements of the objective function at each iteration (to form one gradient approximation), while SPSA takes only two measurements

It is also naturally suited for noisy measurements. Thus, it will be useful when simulating noisy systems.

The theory (and implementation) for SPSA is:
* Described in James C. Spall, “[An Overview of the Simultaneous Perturbation Method for Efficient Optimization](https://www.jhuapl.edu/SPSA/PDF-SPSA/Spall_An_Overview.PDF)”, 1998;

Furthermore, it is implemented:
* In the [`noisyopt` package](https://noisyopt.readthedocs.io/en/latest/), specifically see [the source code here](https://github.com/andim/noisyopt/blob/master/noisyopt/main.py#L264);
* In Qiskit, [see the SPSA optimizer documentation](https://qiskit.org/documentation/stubs/qiskit.algorithms.optimizers.SPSA.html#qiskit.algorithms.optimizers.SPSA).

More information:
https://www.jhuapl.edu/SPSA/

ankit27kh:
@mtthss if no one is working on it, I would like to try. Can you provide me with some general points before I start, like which files to update, what to take care of etc., as I haven't contributed to optax before.

lockwo:
Is there any updates on this? I have previously worked with SPSA in TF (https://github.com/tensorflow/quantum/pull/653) and would be interested in working on this.

ankit27kh:
Hi @lockwo, are you still interested? If you can implement SPSA, it'll be of great help!

fabianp:
@ankit27kh : since there hasn't been activity for this in a year, I think it's safe for you to take over.

if you end up contributing this example, please do so to the contrib/ directory. Thanks!

carlosgmartin:
@fabianp I've created the following implementation of a pseudo-gradient estimator:

https://gist.github.com/carlosgmartin/0ee29182a17b35baf7d402ebdc797486

SPSA corresponds to the special case where `sampler=random.rademacher`.

I'd be happy to contribute this implementation to Optax. I also welcome any feedback. A good place to put it might be under [`optax.tree_utils`](https://optax.readthedocs.io/en/latest/api/utilities.html#tree).

carlosgmartin:
[...]

PR #1712, opened 2026-06-23
feat: Add SPSA optimization method (Issue #357)

Paramveersingh-S:
Addresses #357

## Description
This PR implements the Simultaneous Perturbation Stochastic Approximation (SPSA) gradient estimator to address the open feature request #357.

Rather than implementing it as a stateful `optax` optimizer, it is implemented as a composable gradient estimator (`optax.contrib.spsa_estimator`). This aligns best with JAX's functional paradigm, allowing users to pass the resulting `grad_fn` directly into any existing `optax` optimizer (SGD, Adam, etc.) and `optax.chain`. Standard polynomial schedules for learning rate and perturbation scaling are also provided.

## Verification
I have added rigorous unit tests in `tests/contrib/spsa_test.py` utilizing `chex.all_variants`:
1. **Unbiasedness Test**: Verified that over 10,000 samples, the expected SPSA gradient strictly matches the true gradient for a multivariable quadratic function.
2. **Optimizer Integration Test**: Verified seamless integration with `optax.sgd` minimizing a noisy objective over 50 steps.
3. Successfully tested compilation under `jax.jit` and `jax.vmap`.

> Note on Author Verification: Since SPSA is a classical algorithm by Spall (1998) and not a recent paper, I did not directly email the author. However, the mathematical unbiasedness tests confirm its correctness.


optax/contrib/_spsa.py
```diff
@@ -0,0 +1,117 @@
...
+
+import jax
+import optax.tree
+from optax._src import base
+
+
+def spsa_standard_schedule(
+    init_value: float,
+    decay_rate: float,
+    offset: float = 0.0,
```
servusdei2018: If a user instantiates this schedule with the defaults, the very first step (`count=0`) will result in a `ZeroDivisionError` (or yield `inf` in JAX). Change the default offset to something mathematically stable, or at least enforce that offset > 0 if count starts at 0.

optax/contrib/_spsa.py
```diff
@@ -0,0 +1,117 @@
...
+
+        y_plus = value_fn(params_plus, *args, **kwargs)
+        y_minus = value_fn(params_minus, *args, **kwargs)
+
+        # Note: Since delta_i is either 1 or -1, dividing by delta_i is
+        # equivalent
+        # to multiplying by delta_i. We multiply for numerical stability.
+        grad_estimate = jax.tree.map(
+            lambda d: (y_plus - y_minus) / (2.0 * c) * d, delta
+        )
```
servusdei2018: You are recalculating for every single leaf in the PyTree. `y_plus`, `y_minus`, and `c` are all scalars. Calculate this scalar coefficient once outside the tree map, then just apply the multiplication

servusdei2018: Do this instead:
```python
scalar_diff = (y_plus - y_minus) / (2.0 * c)
grad_estimate = jax.tree.map(lambda d: scalar_diff * d, delta)
```

optax/contrib/_spsa.py
```diff
@@ -0,0 +1,117 @@
...
+        params_minus = jax.tree.map(lambda p, d: p - c * d, params, delta)
+
+        y_plus = value_fn(params_plus, *args, **kwargs)
+        y_minus = value_fn(params_minus, *args, **kwargs)
+
+        # Note: Since delta_i is either 1 or -1, dividing by delta_i is
+        # equivalent
+        # to multiplying by delta_i. We multiply for numerical stability.
+        grad_estimate = jax.tree.map(
+            lambda d: (y_plus - y_minus) / (2.0 * c) * d, delta
```
servusdei2018: We need numerical safety here. If c decays to exactly 0 or gets sufficiently small this division will explode.

tests/contrib/spsa_test.py
```diff
@@ -0,0 +1,93 @@
...
+    def test_spsa_standard_schedule(self):
+        schedule = optax.contrib.spsa_standard_schedule(
+            init_value=1.0, decay_rate=0.5, offset=10.0
+        )
+
+        val_0 = schedule(0)
+        val_10 = schedule(10)
+
+        self.assertAlmostEqual(val_0, 1.0 / (10.0**0.5))
+        self.assertAlmostEqual(val_10, 1.0 / (20.0**0.5))
```
servusdei2018: Stick to `np.testing.assert_allclose`

Paramveersingh-S:
Thanks for the thorough review, @servusdei2018! I've pushed a new commit addressing all of your points:

1. **Schedule stability**: I've updated `spsa_standard_schedule` to default to `offset = 1.0` so that `count=0` is mathematically stable right out of the box and avoids yielding `inf`.
2. **Optimized gradient calculation**: The scalar division `(y_plus - y_minus) / (2.0 * safe_c)` is now hoisted and calculated just once before the `jax.tree.map`, applying only the multiplication across the PyTree leaves.
3. **Numerical safety**: I've protected the perturbation scale division using `jnp.maximum(c, jnp.finfo(jnp.result_type(c)).eps)` to explicitly avoid `ZeroDivisionError` explosions as `c` decays.
4. **Testing Standards**: I've updated the testing macros in `spsa_test.py` from `assertAlmostEqual` over to `np.testing.assert_allclose`.

All tests pass perfectly locally. Let me know if everything looks good on your end!

PR #1714, opened 2026-06-26
contrib: add sharding tests for contrib optimizers

irhyl:
## Summary

- Adds `optax/contrib/_sharding_test.py`, mirroring the pattern established in `optax/_src/sharding_test.py`
- Tests 14 contrib optimizers: `acprop`, `ademamix`, `adopt`, `cocob`, `dadapt_adamw`, `dog`, `dowg`, `galore`, `madgrad`, `muon`, `prodigy`, `schedule_free_adamw`, `schedule_free_sgd`, `simplified_ademamix`
- Three test cases per optimizer:
  - `test_init_with_abstract_input`:  verifies state can be created from `jax.ShapeDtypeStruct` (8 optimizers that support abstract init)
  - `test_state_sharding_type_init_match_update`: verifies sharding type is preserved across a gradient step (JAX ≥ 0.7.2, all 14 optimizers)
  - `test_state_sharding_type_preserved_with_jit`: verifies JIT does not alter sharding type (JAX ≥ 0.7.2, all 14 optimizers)

Optimizers that require concrete parameter values during `init` (e.g. `dadapt_adamw` which records an initial norm, `dog`/`dowg`/`prodigy` which track distance from initialization) are included in the sharding type tests but excluded from the abstract-init test, with a comment explaining the distinction.

### DIFF
--- a/optax/contrib/_sharding_test.py
+++ b/optax/contrib/_sharding_test.py
@@ -14,8 +14,6 @@
 # ==============================================================================
 """Sharding and dtype-stability tests for optax.contrib optimizers."""
 
-import os
-
 from absl.testing import absltest
 from absl.testing import parameterized
 import jax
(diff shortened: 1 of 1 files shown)


optax/contrib/_sharding_test.py
```diff
@@ -0,0 +1,116 @@
...
+    ),
+    'schedule_free_sgd': optax.contrib.schedule_free_sgd(
+        1e-2, warmup_steps=5000
+    ),
+}
+
+
+class ContribShardingTest(parameterized.TestCase):
+
+  @parameterized.named_parameters(ABSTRACT_INIT_OPTIMIZERS.items())
```
rdyro: Given a lot of optimizers here don't support an abstract init, I'd omit this test case. This'll simplify the optimizer list.

optax/contrib/_sharding_test.py
```diff
@@ -0,0 +1,116 @@
...
+from absl.testing import absltest
+from absl.testing import parameterized
+import jax
+import jax.numpy as jnp
+import optax
+from optax._src import test_utils
+from optax._src import utils
+
+
+os.environ['XLA_FLAGS'] = '--xla_force_host_platform_device_count=8'
```
rdyro: Can you set up a custom contextmanager with try/finally updating the jax.config.update("jax_cpu_device_num", ...) instead of using this env variable globally?

optax/contrib/_sharding_test.py
```diff
@@ -0,0 +1,116 @@
...
+class ContribShardingTest(parameterized.TestCase):
+
+  @parameterized.named_parameters(ABSTRACT_INIT_OPTIMIZERS.items())
+  def test_init_with_abstract_input(self, optimizer):
+    params = jax.ShapeDtypeStruct(shape=(2, 4, 8), dtype=jnp.float32)
+    state = optimizer.init(params)
+    self.assertIsNotNone(state)
+
+  @parameterized.named_parameters(SHARDING_OPTIMIZERS.items())
+  def test_state_sharding_type_init_match_update(self, optimizer):
```
rdyro: is the only difference between the two tests the application of `jax,jit`? maybe we should keep only one of the two versions or make this test parametric, there's a lot of duplication at them moment

rdyro (commented):
Thanks, this looks like a promising PR!

optax/contrib/_sharding_test.py
irhyl: Done, dropped the abstract-init test. The optimizer list is now the same set used in the sharding test.

optax/contrib/_sharding_test.py
irhyl: Yes, merged into a single test_state_sharding_type_stable. It now checks all three cases in one pass: eager init/update type stability, jit init matches eager init, and jit update matches eager update. Less repetition, same coverage.

optax/contrib/_sharding_test.py
irhyl: The actual config key is jax_num_cpu_devices (not jax_cpu_device_num), and it can only be set before the JAX backend initializes, a try/finally contextmanager would error on teardown with RuntimeError: jax_num_cpu_devices config should be updated before backends are initialized.

Replaced the os.environ global with jax.config.update('jax_num_cpu_devices', 8) at module level, before any JAX operations. This matches how optax/_src/sharding_test.py handles it (it uses os.environ['XLA_FLAGS'] at module level for the same reason), just using the JAX config API instead of env variables.

optax/contrib/_sharding_test.py
```diff
@@ -0,0 +1,93 @@
...
+
+      # Eager init and update should have matching sharding types.
+      state_eager = optimizer.init(params)
+      _, state_after_update = optimizer.update(params, state_eager, params)
+      test_utils.assert_trees_all_equal(
+          jax.tree.map(jax.typeof, state_eager),
+          jax.tree.map(jax.typeof, state_after_update),
+      )
+
+      # JIT-compiled init and update should match their eager counterparts.
```
rdyro: can we delete all the jit tests, the eager is sufficient

irhyl: Done, removed the jit checks. The test now only verifies that init and update preserve sharding types under eager execution.

rdyro (approved):
Thanks!

irhyl:
@rdyro thankyou for the review and approval. Let me know if there's anything else needed from my side before merging.

PR #1715, opened 2026-06-26
Fix incorrect (1 - b2) coefficient in scale_by_adan

codewithfourtix:
### Summary

`scale_by_adan` scales the gradient-difference term by `1 - b2` in two
places, but the algorithm documented in `optax.adan` scales it by
`(1 - beta_2)`, which equals `b2` in optax's parameterization. With the
default `b2=0.92` the optimizer applied a weight of `0.08` instead of
`0.92`, diverging from the documented equations from the second step onward.

### Details

optax parameterizes Adan's moment EMAs via `update_moment(..., decay=b2)`,
i.e. `v_t = b2 * v_{t-1} + (1 - b2) * (g_t - g_{t-1})`. Matching this to the
documented equation `v_t = (1 - beta_2) v_{t-1} + beta_2 (g_t - g_{t-1})`
forces `beta_2 = 1 - b2`. Therefore the documented coefficient `(1 - beta_2)`
that appears in:

- the squared accumulator `n_t = ... (g_t + (1 - beta_2)(g_t - g_{t-1}))^2`
- the update `u_t = m_t + (1 - beta_2) v_t`

equals **`b2`**, not `1 - b2`. This also matches Algorithm 1 of
Xie et al., 2022 (https://arxiv.org/abs/2208.06677), where the term is
`(1 - beta_2)`.

The first step is unaffected (the gradient difference is zero), which is why
the existing "loss decreases" tests did not catch this.

### Changes

- `scale_by_adan`: use `b2` instead of `1 - b2` for the gradient-difference
  term in both the `n` accumulator and the update direction.
- Add `test_adan_matches_documented_update`, which checks `scale_by_adan`
  against an independent reference implementation of the documented equations
  over several steps.

### Compatibility

This changes the numerical behavior of `optax.adan` to match its documented
algorithm and the reference paper. Existing code using `optax.adan` will now
follow the intended Adan update.

### DIFF
--- a/optax/_src/transform.py
+++ b/optax/_src/transform.py
@@ -665,7 +665,7 @@ def scale_by_adan(
     m = optax.tree.update_moment(g, state.m, b1, 1)
     v = optax.tree.update_moment(diff, state.v, b2, 1)
 
-    sq = optax.tree.add_scale(g, 1 - b2, diff)
+    sq = optax.tree.add_scale(g, b2, diff)
     n = optax.tree.update_moment_per_elem_norm(sq, state.n, b3, 2)
 
     t = numerics.safe_increment(state.t)
@@ -673,7 +673,7 @@ def scale_by_adan(
     v_hat = optax.tree.bias_correction(v, b2, t)
     n_hat = optax.tree.bias_correction(n, b3, t)
 
-    u = optax.tree.add_scale(m_hat, 1 - b2, v_hat)
+    u = optax.tree.add_scale(m_hat, b2, v_hat)
     denom = jax.tree.map(lambda n_hat: jnp.sqrt(n_hat + eps_root) + eps, n_hat)
     u = optax.tree.div(u, denom)
 
--- a/optax/_src/transform.py
+++ b/optax/_src/transform.py
@@ -621,9 +621,9 @@ class ScaleByAdanState(NamedTuple):
 
 
 def scale_by_adan(
-    b1: jax.typing.ArrayLike = 0.98,
-    b2: jax.typing.ArrayLike = 0.92,
-    b3: jax.typing.ArrayLike = 0.99,
+    b1: jax.typing.ArrayLike = 0.02,
+    b2: jax.typing.ArrayLike = 0.08,
+    b3: jax.typing.ArrayLike = 0.01,
     eps: jax.typing.ArrayLike = 1e-8,
     eps_root: jax.typing.ArrayLike = 0.0,
 ) -> base.GradientTransformation:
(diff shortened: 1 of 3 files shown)


optax/_src/transform.py
```diff
@@ -665,15 +665,15 @@ def update_fn(updates, state, params=None):
     m = optax.tree.update_moment(g, state.m, b1, 1)
     v = optax.tree.update_moment(diff, state.v, b2, 1)
 
-    sq = optax.tree.add_scale(g, 1 - b2, diff)
```
rdyro, line 668: I'd prefer optax stick closer to the math and we keep 1 - b2, but update the constructor defaults.

optax/_src/transform_test.py
```diff
@@ -73,6 +73,41 @@ def test_scalers(self, scaler_constr):
     test_utils.assert_tree_all_finite((params, updates, state))
     test_utils.assert_trees_all_equal_shapes(params, updates)
 
+  def test_adan_matches_documented_update(self):
+    # Reference implementation of the equations documented in `optax.adan`,
+    # where the gradient-difference term in both `n` and the update is scaled
+    # by (1 - beta_2), which equals `b2` in optax's (1 - beta) parameterization.
+    b1, b2, b3 = 0.98, 0.92, 0.99
```
rdyro: See comment above, use 0.02, etc.

rdyro (commented):
Thanks, you seem to have caught a genuine problem with the optax implementation!

rdyro:
You'll need to update the docstring in adan docstring in [alias.py](https://github.com/google-deepmind/optax/blob/cd65f389d5e6f98e50d515a10f6109e4a2d16e31/optax/_src/alias.py#L921) too:
```
Document: api/generated/optax.adan
----------------------------------
**********************************************************************
File "../optax/_src/alias.py", line ?, in default
Failed example:
    for _ in range(5):
     grad = jax.grad(f)(params)
     updates, opt_state = solver.update(grad, opt_state, params)
     print('Objective function: {:.2E}'.format(f(params)))
Expected:
    Objective function: 1.28E+01
    Objective function: 1.17E+01
    Objective function: 1.07E+01
    Objective function: 9.68E+00
    Objective function: 8.76E+00
Got:
    Objective function: 1.28E+01
    Objective function: 1.17E+01
    Objective function: 1.07E+01
    Objective function: 9.69E+00
    Objective function: 8.77E+00
```

codewithfourtix:
@rdyro  Done, kept the 1 - b2 form and switched the defaults to the paper's betas (b1=0.02, b2=0.08, b3=0.01), flipping the moment/bias-correction decays to 1 - b_i so the EMAs stay consistent. Behavior is unchanged, the parameterization now matches the paper. Updated the test to the new defaults and the docstring example. Let me know if you'd like anything else.

Issue #1132, opened 2024-11-09
Possible bug: `acprop` explodes during initial steps


carlosgmartin:
There appears to be a bug in [`acprop`](https://optax.readthedocs.io/en/latest/api/contrib.html#optax.contrib.acprop) that causes it to explode during its initial steps in some situations:

<details>

<summary>Code</summary>

```python3
import jax
import optax
from jax import lax, numpy as jnp
from matplotlib import pyplot as plt, rcParams


def rosenbrock_function(params, a=1, b=100):
    x, y = params
    return jnp.square(a - x) + b * jnp.square(y - jnp.square(x))


def main():
    lr = 1e-4
    iters = 1000

    params = jnp.zeros(2)

    fig, ax = plt.subplots(constrained_layout=True)

    for label, opt in [
        ("sgd", optax.sgd(lr)),
        ("adam", optax.adam(lr)),
        ("adabelief", optax.adabelief(lr)),
        ("acprop", optax.contrib.acprop(lr)),
    ]:

        def update(state, _):
            params, opt_state = state
            value, grads = jax.value_and_grad(rosenbrock_function)(params)
            updates, opt_state = opt.update(grads, opt_state, params)
            params = optax.apply_updates(params, updates)
            return (params, opt_state), value

        opt_state = opt.init(params)
        _, values = lax.scan(update, (params, opt_state), length=iters)
        ax.plot(values, label=label)

    ax.legend()
    ax.set(xlabel="iteration", ylabel="value", yscale="log")
    ax.grid(True, "both", alpha=0.1)
    rcParams["savefig.dpi"] = 300
    plt.show()


if __name__ == "__main__":
    main()
```

</details>

<img src="https://github.com/user-attachments/assets/4fcbd75d-d678-44ce-a2ea-3bfaa933c11a" width="400">

Zooming in:

<img src="https://github.com/user-attachments/assets/e6198840-f968-4de1-90f8-6752bd77e1aa" width="400">

I'm currently investigating this.

vroulet:
Hey @carlosgmartin,
Did you find the bug? We may report the bug to the original repo and author and add a note on the optimizer in the mean time (to be honest it's also fine if we don't add a note since many deep learning optimizers are quite frail anyway).
Once done we may close the issue.

carlosgmartin:
[...]

PR #1716, opened 2026-06-27
Fix ACProp second moment to use the new first moment

codewithfourtix:
### Summary

`scale_by_acprop` computes its second-moment residual against `state.mu`
(the previous-step first moment `m_{t-1}`) instead of the freshly-updated
`mu` (`m_t`), contradicting the documented update rule.

### Details

The `optax.contrib.acprop` docstring documents the second moment as:

    s_t = b2 * s_{t-1} + (1 - b2) * (g_t - m_t) ** 2 + eps_root

where `m_t` is the first moment **after** the step-t update. The code already
computes that new moment on the line above (`mu = update_moment(...)`), but the
residual is then computed as `g - state.mu`, i.e. `g_t - m_{t-1}`.

The new moment is the correct one on three independent grounds:
- the docstring's own equation uses `(g_t - m_t)`;
- the original PyTorch implementation the docstring follows
  (juntang-zhuang/ACProp-Optimizer) updates `exp_avg` first and then computes
  `grad - exp_avg`;
- optax's own AdaBelief sibling `scale_by_belief` uses the freshly-updated `mu`.

ACProp's asynchrony concerns which *second* moment feeds the denominator
(`s_{t-1}`), which the code already implements correctly; it does not affect
the residual's mean term.

Example: with `b1=0.9, b2=0.999, eps_root=0`, a first step of `g=1.0` should
give `s_1 = (1 - b2) * (g - m_1)**2 = 0.001 * 0.9**2 = 0.00081`, but the
current code yields `0.001` (using `g - m_0 = 1.0`).

### Changes

- `scale_by_acprop`: use the updated `mu` in the prediction-error residual.
- Add `_acprop_test.py` pinning the second moment to the documented
  `(g_t - m_t)` residual over several steps.

### DIFF
--- a/optax/contrib/_acprop.py
+++ b/optax/contrib/_acprop.py
@@ -62,7 +62,7 @@ def scale_by_acprop(
   def update_fn(updates, state, params=None):
     del params
     mu = optax.tree.update_moment(updates, state.mu, b1, 1)
-    prediction_error = jax.tree.map(lambda g, m: g - m, updates, state.mu)
+    prediction_error = jax.tree.map(lambda g, m: g - m, updates, mu)
     nu = optax.tree.update_moment_per_elem_norm(prediction_error, state.nu, b2,
                                                 2)
     nu = jax.tree.map(lambda v: v + eps_root, nu)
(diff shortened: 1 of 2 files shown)


carlosgmartin:
Potentially relevant:

https://github.com/google-deepmind/optax/issues/1132

https://github.com/aymuos15/optax/commit/449787b5c91cb50019674bca7d6a7dd7ddcc81ec

PR #1730, opened 2026-07-21
Treat integer hyperparameters as static in inject_hyperparams (fixes #412)

### LINKED ISSUE #412: Problems when jitting Adafactor with inject_hyperparams.
When wrapping `optax.adafactor` with `optax.inject_hyperparams` without specifying `static_args`

```
optax.inject_hyperparams(optax.adafactor)(learning_rate=0.1)
```

the `init` function of the resulting `GradientTransformation` cannot be jit compiled. The reason is that by default `inject_hyperparams` treats all arguments as dynamic and one of the argument has to be static to avoid a `TracerError`. A workaround is to specify the static argument:

```
optax.inject_hyperparams(optax.adafactor, static_args=("min_dim_size_to_factor",))(learning_rate=0.1)
```

However, this is not ideal since it requires the user to know which arguments should be static and which ones can be dynamic.

We should:

- Add a test to check whether any other optimizers are affected.
- Change the implementations so that all optimizers wrapped in inject_hyperparams can be jit compiled without any arguments being specified as static.

ArneshBanerjee:
## Summary

`optax.inject_hyperparams` converts every numeric argument into a traced array so it can be scheduled/overridden at runtime. Boolean arguments are already special-cased as *static*, because a traced boolean can't be used in Python control flow.

Integer arguments have the identical problem, but currently fall through to the "numeric" branch and get traced. Several optimizers use integer arguments for **structural** decisions:

- `min_dim_size_to_factor` in `adafactor` (`optax/_src/factorized.py:55` → `if shape[...] < min_dim_size_to_factor`)
- `memory_size` in `lbfgs` (`optax/_src/transform.py:1717` → `if memory_size < 1`)

As a result, jitting these fails with `TracerBoolConversionError` unless the user manually passes `static_args=(...)`:

```python
import jax, jax.numpy as jnp, optax
opt = optax.inject_hyperparams(optax.adafactor)(learning_rate=0.1)
jax.jit(opt.init)(jnp.ones((4, 4)))   # TracerBoolConversionError
```

This is exactly what #412 asks to fix ("all optimizers wrapped in `inject_hyperparams` can be jit compiled without any additional `static_args`").

## Fix

Since `bool` is a subclass of `int`, extending the existing static-value check from `bool` to `int` generalizes the current behavior and resolves both optimizers with no new special cases. Integer hyperparameters cannot be meaningfully scheduled under jit anyway (a schedule returns floats, and structural ints break control flow when traced).

## Verification

- `inject_hyperparams(optax.adafactor)` and `inject_hyperparams(optax.lbfgs)` now `jax.jit` out of the box; the previously-required `static_args` workaround is no longer needed.
- Added regression tests in `optax/schedules/_inject_test.py`.
- `pytest optax/schedules/ optax/_src/alias_test.py` → **634 passed, 70 skipped**, no regressions.

Fixes #412.


ArneshBanerjee:
Rebased on main; CLA is signed and CI is green.

This fixes #412: `inject_hyperparams` traced integer hyperparameters, which breaks inner factories that use them for structural control flow (`min_dim_size_to_factor` in adafactor, `memory_size` in lbfgs) with a `TracerBoolConversionError` under jit. Integers are now treated as static, matching how bools were already handled.

Would someone be able to review?

sylvesterkaczmarek (commented):
The new rule only recognizes Python `int`; zero-dimensional integer NumPy/JAX arrays still fall into `numeric_hps` and get traced. Thus a structural argument such as `memory_size=jnp.asarray(10)` can still hit the same `TracerBoolConversionError` this change is intended to prevent. Please classify integer-dtype scalar arrays as static too, or key the rule to structural parameters, and test NumPy/JAX integer scalars.

ArneshBanerjee:
I reproduced it: with `memory_size=jnp.asarray(10)` or `np.asarray(10)` the value still landed in `numeric_hps` and got traced, so it hit the same `TracerBoolConversionError`. (`np.int64(10)` happened to work already, but only by accident, since it is not an `ndarray` and fell through to the static branch.)

Pushed a fix:

1. The static rule now matches zero-dimensional NumPy and JAX values of integer or boolean dtype, not just Python `int`.

2. Classifying them as static is not sufficient on its own for `jax.Array`. An array closed over by a jitted function is staged out as a constant of the trace, so using it for structural control flow still raised the same error even once it was no longer injected. Static integer and boolean values are now converted to Python scalars so they are real compile-time constants. A traced value is passed through unchanged so the inner factory reports it rather than this raising something less clear.

I limited the rule to zero-dimensional values on purpose. A raw `jax.random.PRNGKey` is `uint32` with shape `(2,)`, so anything keyed on integer dtype alone would have made keys static too. There is a test covering both key types to pin that down.

Tests added over `int`, `bool`, `np.int64`, `np.asarray`, `jnp.asarray` and a bool array, checking the hyperparameter stays static and that the update jits. They fail on the previous commit for the three array cases.

sylvesterkaczmarek (commented):
Zero-dimensional NumPy/JAX integer and boolean values are now classified as static and converted to Python scalars before they can be closed over by jit as traced arrays. The new tests cover the structural integer cases that previously still raised TracerBoolConversionError. My finding is resolved.

PR #1742, opened 2026-07-28
Fix `AttributeError` in `Sharpness-Aware Minimization (SAM)` example notebook by using `_replace` for `NamedTuple`

rajasekharporeddy:
**Issue Description**:

Running the `examples/sam.ipynb` notebook currently fails with an 

```python
AttributeError: 'Store' object has no attribute 'replace'
```

during the training loops. This is because the `Store` class is defined as a `typing.NamedTuple`. In Python, the standard method to create a new instance of a `NamedTuple` with updated fields is `_replace()`, not `replace()`.

**Fix**:

This PR fixes the issue by updating `store.replace(...)` to `store._replace(...)` in both instances of the `make_step` function within the notebook.

### DIFF
--- a/examples/contrib/sam.ipynb
+++ b/examples/contrib/sam.ipynb
@@ -1,10 +1,10 @@
 {
   "cells": [
     {
+      "cell_type": "markdown",
       "metadata": {
         "id": "qEIr4kn4W6Zs"
       },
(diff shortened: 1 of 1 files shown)


rdyro (approved):
Thanks!

PR #1749, opened 2026-08-06
Skip invalid L-BFGS curvature updates

### LINKED ISSUE #1708: LBFGS - Add functionality to skip updates
Would it be possible to add the ability to skip updates to the lbfgs algorithm, in case a certain curvature condition does not hold? You can find the relevant information on page 537 of the book "Numerical Optimization" by Nocedal and Wright.

While it isn't great to skip updates, it prevents the algorithm from breaking when the Wolfe conditions are not satisfied (sometimes, one needs to use backtracking line search due to computational restrictions).

This would mean to add about 20 lines to the existing optax code - I'm happy to provide my current implementation.

Trighap52:
Fixes #1708.

## Summary
- Skip L-BFGS memory updates when the curvature pair has non-positive `s^T y`.
- Avoid writing negative inverse-curvature weights into `weights_memory`.
- Add regression coverage for the backtracking line-search reproducer from the issue.
- Document the skip behavior for `lbfgs` and `scale_by_lbfgs`.

## Why
When backtracking/Armijo line search is used, L-BFGS can observe a parameter/gradient pair that does not satisfy the positive-curvature condition. Storing that pair makes `weights_memory` negative, which can break the inverse Hessian approximation. This change leaves such slots as zero so the two-loop recursion ignores them.

## Tests
- `pre-commit run --files optax/_src/transform.py optax/_src/alias.py optax/_src/lbfgs_test.py`
- `ruff check .`
- `flake8 --select=E9,F63,F7,F82,E225,E251 --show-source --statistics`
- `pylint` module, test, and quote-consistency checks
- `pyrefly check optax/_src/transform.py optax/_src/alias.py`
- `pytest -n auto --pyargs optax`
- `make html -C docs`
- `make doctest -C docs`

### DIFF
--- a/optax/_src/alias.py
+++ b/optax/_src/alias.py
@@ -2703,7 +2703,9 @@ def lbfgs(
   preconditioning matrix subject to some secant condition, see references
   for more details. Computing :math:`P_k u_k` can be done by a sequence of
   vector operations using past differences of parameters and gradients stored in
-  a memory bufffer.
+  a memory buffer.
+  Pairs whose curvature :math:`\delta u_k^\top \delta w_k` is not positive are
+  skipped to preserve a positive-definite inverse Hessian approximation.
 
   The present function just outputs the LBFGS direction :math:`P_k u_k`.
   It can be chained with a linesearch ensuring sufficient decrease and low
--- a/optax/_src/lbfgs_test.py
+++ b/optax/_src/lbfgs_test.py
@@ -207,16 +207,28 @@ def _plain_lbfgs(
       else:
         identity_scale = jnp.vdot(dus[-1], dws[-1])
         identity_scale /= jnp.sum(dus[-1] ** 2)
+        identity_scale = jnp.where(
+            jnp.isnan(identity_scale), 1.0, identity_scale
+        )
     else:
       identity_scale = 1.0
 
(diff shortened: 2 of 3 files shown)


Trighap52:
I looked into the failing jobs.

`Pytest 3.10 on ubuntu-latest jax=newest` is related to this PR. It fails in:

```text
LBFGSTest::test_against_plain_implementation3
problem_name = 'himmelblau', scale_init_precond = False
ACTUAL:  [234.10304, 179.07068]
DESIRED: [234.02055, 179.15878]
```

I reproduced this locally with `jax==0.6.2`. The L-BFGS curvature guard itself is working; the mismatch appears to be a compiled `lax.while_loop` float32 numerical divergence in the existing plain-implementation comparison after the first tiny positive curvature pair.

Would you prefer that I open a small follow-up PR to relax the tolerance for this comparison while keeping the new regression test strict about preventing negative `weights_memory`?

The `Build and check types with pyrefly` failure appears unrelated to this PR. The reported errors are in untouched upstream files such as `optax/_src/linear_algebra.py`, `optax/_src/linesearch.py`, `optax/perturbations/_make_pert.py`, and `optax/projections/_projections.py`; none are in the files changed here.

PR #1751, opened 2026-08-13
Support negative axes in microbatching

anxkhn:
## Summary

Canonicalize negative batch axes before reshaping and slicing microbatches.

Negative `in_axes` values were used directly in shape slicing, producing invalid reshape targets. The fix resolves each axis against the relevant array rank in both `reshape_batch_axis` and `_take_fn`.

## Tests

- Added a regression test for `in_axes=-1`
- Verified the full microbatching test module

### DIFF
--- a/optax/_src/linear_algebra.py
+++ b/optax/_src/linear_algebra.py
@@ -276,6 +276,7 @@ def matrix_inverse_pth_root(
         _iter_condition, _iter_body, init_state
     )
     error = jnp.max(jnp.abs(mat_m - identity))
+    # pyrefly: ignore [missing-attribute]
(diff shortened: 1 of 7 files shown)


anxkhn:
hey @irhyl, could you please review this pr when you have a chance?

irhyl:
> hey @irhyl, could you please review this pr when you have a chance?

hi @anxkhn 
i am not the reviewer of this project, but i did take a look at the pr. i think ur at the right track with the negative axes.

the test is a bit minimal, though. it’d be good to cover a few more shapes, including in_axes=-2. if possible, also add a sharding test. otherwise, solid catch on the canonicalization.

PR #1755, opened 2026-08-16
Fix NaN in scale_by_adam on float16 zero-gradient steps

### LINKED ISSUE #1754: Adam-family optimizers (adam, amsgrad, adabelief, yogi) produce NaN on float16 zero-gradient steps
\`scale_by_adam\`, \`scale_by_amsgrad\`, \`scale_by_belief\`, and \`scale_by_yogi\` in \`optax/_src/transform.py\` compute their update as \`m / (sqrt(v + eps_root) + eps)\`. When params are \`float16\`, JAX's weak-type promotion casts the Python-float \`eps\` (default \`1e-8\`, or \`1e-16\` for belief) down to \`float16\` before the addition. float16's smallest representable subnormal is \`~6e-8\`, so \`eps\` rounds to exactly \`0.0\`.

On any step with an exactly-zero gradient (masked/padded tokens, a frozen-then-unfrozen layer, an unused embedding row, etc.), both moment estimates are \`0\`, so the denominator becomes \`sqrt(0) + 0 == 0\`, giving \`0 / 0 = NaN\`. Since NaN propagates through every later step, this silently poisons the rest of training.

\`float16\` is already an explicitly tested dtype elsewhere (\`alias_test.py::test_state_shape_dtype_shard_stability\`), but that test's gradient is never exactly zero, so it doesn't catch this. \`scale_by_yogi\`'s larger default \`eps=1e-3\` happens not to underflow, but is still fragile with a smaller \`eps\`.

kartik1pandey:
## Summary

Fixes NaN in `scale_by_adam` when used with `float16` params on a zero-gradient step.

### Changes

- `scale_by_adam`'s update now adds `eps`/`eps_root` in at least `float32` precision (`jnp.promote_types(v.dtype, jnp.float32)`) before casting the result back to the original dtype.
- Added `test_adam_no_nan_on_float16_zero_grad` regression test (float16 params, 3 steps of all-zero gradients, asserts every leaf stays finite).

### Why

`eps` (a plain Python float, default `1e-8`) is silently rounded to `0.0` when added to a `float16` array, since float16's smallest representable subnormal is `~6e-8`:

```python
>>> jnp.asarray(1e-8, dtype=jnp.float16)
Array(0., dtype=float16)
```

On a step with an exactly-zero gradient (masked tokens, a frozen-then-unfrozen layer, etc.), both moment estimates are `0`, so the `sqrt(v) + eps` safety denominator collapses to `0`, producing `0 / 0 = NaN`:

```python
params = {"w": jnp.array([1.0, -2.0, 0.5], dtype=jnp.float16)}
grads = {"w": jnp.zeros_like(params["w"])}
opt = optax.adam(1e-3)
state = opt.init(params)
updates, state = opt.update(grads, state, params)
print(updates["w"])  # [nan nan nan]
```

Since NaN propagates through every subsequent step via `apply_updates`, this silently poisons the rest of training. The fix is a no-op for `float32`/`float64` (verified bit-for-bit identical output before/after); only `float16`/`bfloat16` behavior changes.

Fixes #1754.

Scoped to `scale_by_adam` only — `scale_by_amsgrad`, `scale_by_belief`, and `scale_by_yogi` share the same pattern (tracked in the linked issue); happy to follow up separately.

### Testing

```
$ python -m pytest optax/_src/transform_test.py optax/_src/alias_test.py -q
560 passed, 62 skipped
```
`ruff check` is clean on both changed files.

### DIFF
--- a/optax/_src/numerics.py
+++ b/optax/_src/numerics.py
@@ -103,6 +103,48 @@ def safe_root_mean_squares(
   return jnp.where(rms <= min_rms, min_rms, jnp.sqrt(jnp.mean(abs_sq(x))))
 
 
+def add_eps_in_safe_dtype(
+    x: jax.typing.ArrayLike,
+    eps: jax.typing.ArrayLike,
+    min_dtype: jax.typing.DTypeLike = jnp.float32,
+) -> jax.Array:
+  """Returns ``x + eps``, computed in at least ``min_dtype`` precision.
+
+  ``eps`` is typically a small constant (e.g. the default ``1e-8`` used by
+  several optimizers) added to a value before taking its square root, to
+  keep a downstream division safe from a zero denominator. If ``x`` has a
+  low precision dtype (e.g. ``float16``, whose smallest representable
+  subnormal is ``~6e-8``), adding such an ``eps`` directly can silently
+  round to exactly ``x`` -- typically ``0.0`` when ``x`` is itself ``0``,
+  which defeats the purpose of the safety term and can turn it into a
+  ``0 / 0`` NaN. This function instead promotes ``x`` to at least
+  ``min_dtype`` before the addition, so ``eps`` is preserved.
+
+  This is a no-op change in behavior (bit-for-bit identical) when ``x`` is
+  already at least ``min_dtype`` precision (e.g. the default ``float32``
+  or ``float64``); only lower-precision dtypes (``float16``, ``bfloat16``)
+  are affected.
+
+  Callers that need the *result* of a computation built on top of this sum
+  back in ``x``'s original dtype (e.g. to keep an optimizer's returned
+  updates at the same precision as the incoming gradients) must cast back
+  explicitly -- this function only fixes the addition itself.
+
+  Args:
+    x: array the (possibly too-small-to-represent) ``eps`` is added to.
+    eps: term to add, typically a small non-negative constant.
+    min_dtype: the minimum floating point precision to compute the addition
+      in. Defaults to ``float32``.
+
+  Returns:
+    ``x.astype(safe_dtype) + eps``, where ``safe_dtype =
+    jnp.promote_types(x.dtype, min_dtype)``.
+  """
+  x = jnp.asarray(x)
+  safe_dtype = jnp.promote_types(x.dtype, min_dtype)
+  return x.astype(safe_dtype) + eps
+
+
 def safe_increment(count: jax.typing.ArrayLike) -> jax.Array:
   """Increments counter by one while avoiding overflow.
 
--- a/optax/_src/transform.py
+++ b/optax/_src/transform.py
@@ -136,11 +136,27 @@ def scale_by_rms(
     else:
       count_inc = jnp.asarray(0)
       nu_hat = nu
+    # `eps` is added in at least float32 precision (see
+    # `numerics.add_eps_in_safe_dtype`) so it does not silently underflow to
+    # zero under a low precision dtype like float16, which would otherwise
+    # turn `rsqrt(0)`/`1 / (0 + eps)` into `inf`, and then `inf * 0` (e.g. on
+    # a zero-gradient step) into a NaN update.
     if eps_in_sqrt:
-      scaling = jax.tree.map(lambda n: jax.lax.rsqrt(n + eps), nu_hat)
+      scaling = jax.tree.map(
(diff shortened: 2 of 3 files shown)


kartik1pandey:
@googlebot I signed it!

winklemad (commented):
The float32-promotion approach here looks right to me — computing the denominator in at least float32 keeps `eps` from rounding to `0.0` under float16 weak-type promotion, which is the actual cause.

One thing worth folding in before this merges: the same `... + eps` denominator (the `+ eps` *outside* the sqrt) is copy-pasted across several sibling optimizers in `transform.py`, and they underflow the same way, so a `scale_by_adam`-only fix leaves them NaN on the same input. I reproduced it against released optax 0.2.8:

```python
import jax.numpy as jnp, optax
p = jnp.zeros((3,), jnp.float16); g = jnp.zeros((3,), jnp.float16)
for name, tx in [
    ('adam',    optax.scale_by_adam()),
    ('amsgrad', optax.scale_by_amsgrad()),
    ('belief',  optax.scale_by_belief()),
    ('rms',     optax.scale_by_rms(eps_in_sqrt=False)),
    ('stddev',  optax.scale_by_stddev(eps_in_sqrt=False)),
]:
    st = tx.init(p); u, _ = tx.update(g, st, p)
    print(name, bool(jnp.isnan(u).any()))
# adam/amsgrad/belief/rms/stddev -> all True
```

The matching denominators are:
- `scale_by_amsgrad` — `m / (jnp.sqrt(v + eps_root) + eps)` (same line as adam's)
- `scale_by_belief` — `m / (jnp.sqrt(v) + eps)`
- `scale_by_rms` / `scale_by_stddev` — `1 / (jnp.sqrt(...) + eps)` on the `eps_in_sqrt=False` branch

Since it's the same fix each time, it might be cleanest to pull the safe-dtype denominator into a small shared helper and use it at each site, rather than patching adam alone — and then the test could parametrize over the affected optimizers. Happy to help with that if useful. Nice catch on the root cause either way.

kartik1pandey:
Pulled the safe-dtype denominator into a shared
helper, `numerics.add_eps_in_safe_dtype`, and applied it to `scale_by_amsgrad`,
`scale_by_belief`, and both branches of `scale_by_rms`/`scale_by_stddev`.

One addition beyond your repro: the `eps_in_sqrt=True` branch NaNs too
(`rsqrt(0) = inf`, then `inf * 0 = NaN`) fixed the same way, keeping the
`lax.rsqrt` fusion so float32/float64 stay bit-for-bit unchanged.
`scale_by_rms`/`scale_by_stddev` also needed an explicit downcast after the
multiply so the fix doesn't silently upcast float16 updates to float32.

Verified: repro now passes on all 7 cases, bit-for-bit identical
output on float32/float64 vs. pre-fix, 653 passed/62 skipped on
transform/alias/numerics tests, ruff clean. Test is now parametrized over
all 7 cases per your suggestion.

Note: `scale_by_yogi` doesn't actually need this its default `eps=1e-3`
is too large to underflow in float16. `scale_by_adamax` does NaN on the
same input but via a different code path (no shared `sqrt(...)+eps`
denominator)  flagging as a separate issue rather than folding in here.

winklemad:
Nice — this is a cleaner home for it than my per-call change. Folding it into `numerics.add_eps_in_safe_dtype` and covering `scale_by_amsgrad` / `scale_by_belief` / both branches of `scale_by_rms` / `scale_by_stddev` in one place is the right call, and good catch on the `eps_in_sqrt=True` branch too. LGTM from my side — happy to give it another look once CI is green.

kartik1pandey:
CI's green now all checks passing.
Ready whenever you get a chance to take another look. Thanks for the review!

kartik1pandey:
Rebased onto latest `main` to resolve the merge conflict (a duplicate pyrefly suppression that main had already picked up independently). All checks are now passing and the PR is mergeable — ready for review.

kartik1pandey:
@winklemad Update: while this PR was under review, PR #1764 ("Fix pyrefly type-check failures in the OSS CI") merged into `main` first, touching the same lines this branch had already patched — that's what caused the merge conflict.

I've rebased this branch onto latest `main` (commit d01841f) to resolve it — no functional changes to the Adam fix itself, just dropped the now-redundant pyrefly suppression. All CI checks are passing again (lint, pyrefly, doctests, full Pytest matrix) and the branch is conflict-free. Ready for review whenever you get a chance!

PR #1721, opened 2026-07-04
add missing docstring to get_spectral_radius_upper_bound in linear_algebra.py

RavSinghChandan:
`get_spectral_radius_upper_bound` in `optax/_src/linear_algebra.py` has only an inline comment explaining its purpose but no proper docstring, making it easy to miss when browsing generated API docs or using an IDE.

Added a Google-style docstring (matching the style used for `canonicalize_dtype`, `scale_gradient` and other helpers in this module) that explains:
- what the function computes (the minimum of the Frobenius and 1-norms as a spectral-radius upper bound)
- the input type expected
- what is returned

No logic changes.

PR #1757, opened 2026-08-16
Add missing pyrefly suppressions to fix main CI

### LINKED ISSUE #1756: main CI failing on pyrefly type-check (4 pre-existing errors, unrelated to any specific PR)
`pyrefly check` currently fails on `main` itself (confirmed across the last several CI runs, e.g. commit `2edbcbe2`), with 10 errors in 4 locations:

- `optax/_src/linear_algebra.py:279`
- `optax/_src/linesearch.py:429` (×7, same expression)
- `optax/perturbations/_make_pert.py:164`
- `optax/projections/_projections.py:157`

All four are the same class of issue: a value is provably safe at runtime (verified by the existing test suite passing) but its static type is broader than pyrefly can narrow — e.g. a `jax.lax.while_loop` carry variable typed as a union across all carry slots, or `jax.typing.ArrayLike` including `complex` in a comparison that's only ever done on real values. This looks like fallout from a recent JAX release changing its type stubs (see the maintainer commit `0f6d10b`, "add pyrefly suppressions for jax.jit annotation change", which patched a similar case elsewhere the same way), rather than a real bug in optax.

Fix: add narrow `# pyrefly: ignore[<code>]` suppressions at each site, matching the existing convention already used nearby in 3 of these same files. PR to follow.

kartik1pandey:
Fixes #1756.

## Summary

`pyrefly check` fails on `main` with 10 errors across 4 lines. Each is a pre-existing type-stub precision gap (not a real bug — confirmed by running the affected test suites), missed when this codebase migrated from `pytype` to `pyrefly` (#1674) or after a subsequent JAX typing change (cf. `0f6d10b`).

### Changes

- `linear_algebra.py`: suppress `missing-attribute` on `old_mat_h.dtype` — identical pattern to the line immediately below it, which already has this suppression.
- `linesearch.py`: suppress `unsupported-operation` on `decrease_error > atol` — `decrease_error` is typed as `ArrayLike` (includes `complex`), but is only ever assigned real values.
- `_make_pert.py`: suppress `bad-argument-type` on `baseline` passed to `vmap` — `baseline` can be `None`, but is only read inside an `if use_baseline:` guard, so the `None` is never dereferenced.
- `projections.py`: suppress `bad-argument-type` on `unravel_fn(new_values)` — `new_values` is always a `jax.Array` at runtime; the union with `ndarray` comes from `ArrayLike`'s breadth.

### Verification

- `pyrefly check --python-interpreter-path <venv>`: 0 errors (was 10; 482 suppressed vs. 472 before).
- `pytest optax/_src/linesearch_test.py optax/perturbations/_make_pert_test.py optax/projections/_projections_test.py -q`: 295 passed, 1 skipped.
- `optax/_src/linear_algebra_test.py::test_matrix_inverse_pth_root` reproduced standalone (its module needs `flax`, unavailable in my sandbox): all condition-number cases pass with identical error values to the documented thresholds.
- Checked for conflicts with other open PRs touching these files (#1721, #1662) — neither overlaps the changed lines.

### DIFF
--- a/optax/_src/linear_algebra.py
+++ b/optax/_src/linear_algebra.py
@@ -276,6 +276,7 @@ def matrix_inverse_pth_root(
         _iter_condition, _iter_body, init_state
     )
     error = jnp.max(jnp.abs(mat_m - identity))
+    # pyrefly: ignore[missing-attribute]
     is_converged = jnp.asarray(convergence, old_mat_h.dtype)  # pytype: disable=attribute-error  # lax-types # noqa: E501
     resultant_mat_h = is_converged * mat_h + (1 - is_converged) * old_mat_h
     # pyrefly: ignore [missing-attribute]
--- a/optax/_src/linesearch.py
+++ b/optax/_src/linesearch.py
@@ -426,6 +426,7 @@ def scale_by_backtracking_linesearch(
     if verbose:
       # We print information only if the linesearch failed.
       _cond_print(
+          # pyrefly: ignore[unsupported-operation]
           search_state.decrease_error > atol,
           "INFO: optax.scale_by_backtracking_linesearch:\n"
           "Backtracking linesearch failed to find a stepsize ensuring sufficent"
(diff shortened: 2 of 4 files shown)


kartik1pandey:
Closing this as superseded — the pyrefly failures this PR targets were independently fixed on `main` in d01841f ("Fix pyrefly type-check failures in the OSS CI"), which covers all 4 files touched here (and in two cases with a proper type fix rather than a suppression). No remaining diff to contribute.

PR #1761, opened 2026-08-24
Fix all-NaN output for complex inputs with 'schatten' preconditioning in Muon

shoemoney:
`_schatten_first_newton_schulz_iteration` computes the Gram matrix as `a = x @ x.T` (`optax/contrib/_muon.py:233` on main). For complex `x` this is not the Hermitian Gram matrix, so the modified first Newton-Schulz step diverges and `orthogonalize_via_newton_schulz(..., preconditioning='schatten')` returns an all-NaN array. A `muon(preconditioning='schatten')` step on a complex-valued model therefore NaN-poisons the parameters on the first update. Complex support is a tested contract of this code path: `test_newton_schulz` asserts complex Newton-Schulz produces a unitary matrix, but only exercised the default preconditioning.

The siblings in the same function family already take the conjugate: `_aol_first_newton_schulz_iteration` (`optax/contrib/_muon.py:217`) and `_base_newton_schulz_iteration` (`optax/contrib/_muon.py:247`) both use `a = x @ x.T.conj()`. Schatten, added alongside aol in #1602, is the only one that omitted it. `.conj()` is a no-op for real dtypes, so real-input behavior is unchanged.

Fix: use `x @ x.T.conj()` in the schatten first iteration. Test: `test_newton_schulz` is now parameterized over `'frobenius'`, `'aol'`, and `'schatten'`, so its existing real-orthogonality and complex-unitarity assertions cover all three modified first steps.

Verification (CPU, jax 0.11.1): before the fix, the schatten complex case fails with an all-NaN Gram matrix; after, `optax/contrib/_muon_test.py` passes (34 tests, 18 subtests), and the schatten complex output is unitary to ~1e-7 while the real schatten path is numerically unchanged (max deviation from identity ~1e-8). `'spectral'` is deliberately not added to the parameterization: it shares `_base_newton_schulz_iteration`, which already conjugates, so it has no instance of this bug.

### DIFF
--- a/optax/contrib/_muon.py
+++ b/optax/contrib/_muon.py
@@ -230,7 +230,7 @@ def _schatten_first_newton_schulz_iteration(
   """Schatten-4 Preconditioning with Newton-Schulz iteration."""
   # Implements the first Newton-Schulz step with Schatten-4 norm
   # preconditioning which allows for better orthogonalization performance.
-  a = x @ x.T
+  a = x @ x.T.conj()
   rescaling = jnp.clip(jnp.linalg.norm(a, ord='fro', axis=(-2, -1)), min=eps)
   s = jnp.expand_dims(jax.lax.rsqrt(rescaling), (0, -1))
   x, a = x * s, a * s ** 2
--- a/optax/contrib/_muon_test.py
+++ b/optax/contrib/_muon_test.py
@@ -274,7 +274,8 @@ class MuonTest(parameterized.TestCase):
     self.assertIsInstance(get_muon_mu(state)['w2']['a'], _masking.MaskedNode)
     self.assertIsInstance(get_muon_mu(state)['w2']['b'], _masking.MaskedNode)
 
-  def test_newton_schulz(self):
+  @parameterized.parameters('frobenius', 'aol', 'schatten')
+  def test_newton_schulz(self, preconditioning):
     """Test that Newton--Schulz orhogonalizes/unitiarizes correctly."""
     mat_real = jax.random.normal(jax.random.key(0), (4, 3), dtype=jnp.float32)
     mat_complex = jax.random.normal(jax.random.key(0), (4, 3),
@@ -289,6 +290,7 @@ class MuonTest(parameterized.TestCase):
     # For real matrices, Newton--Schulz should produce an orthonormal matrix
     mat_real_orth = _muon.orthogonalize_via_newton_schulz(
         mat_real, ns_coeffs, ns_steps=20, eps=1e-12,
+        preconditioning=preconditioning,
         dimension_numbers=_muon.MuonDimensionNumbers(0, 1),
     )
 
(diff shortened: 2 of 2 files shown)


shoemoney:
CI note: Build and check types with pyrefly failure is pre-existing on main and unrelated to this PR's changed files. The 10 pyrefly errors are in linear_algebra.py/linesearch.py/_make_pert.py/_projections.py with no overlap with _muon.py/_muon_test.py and reproduce identically on base commit 0391582 (main HEAD). No code fix required from this PR; rebase or label will clear it.

sylvesterkaczmarek (commented):
This looks right to me. The Schatten first step now builds the Hermitian Gram matrix like the AOL/base paths, while .conj() is a no-op for real inputs. Parameterizing the existing orthogonality/unitarity test also covers the missing complex Schatten case. I don't see a blocker here.

shoemoney:
Rebased onto current `main` in `3e4fd1e`. The conflict was in `test_newton_schulz`: `main` added `@jax.default_matmul_precision('highest')` while this branch parameterized the same test over the three preconditioning modes. Both are kept, stacked:

```python
@parameterized.parameters('frobenius', 'aol', 'schatten')
@jax.default_matmul_precision('highest')
def test_newton_schulz(self, preconditioning):
```

The precision pin matters more after the sweep than before it, since the complex `schatten` arm is the one that was producing all-NaN output and would be the most tolerance-sensitive at default precision.

The red check is not this change. `Pytest 3.12 on ubuntu-latest jax=nightly` failed in dependency install, before pytest ran at all:

```
× Failed to download `scipy==1.18.1`
├─▶ Failed to fetch:
│   `https://us-python.pkg.dev/ml-oss-artifacts-published/jax/scipy/scipy-1.18.1-cp312-cp312-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl`
╰─▶ HTTP status client error (404 Not Found)

hint: `scipy` (v1.18.1) was included because `jaxlib` (v0.11.2.dev20260911) depends on `scipy`
```

That is a 404 against the jax nightly artifact registry. Worth flagging beyond this PR: the matrix is fail-fast, so that one install failure cancelled `jax=newest` on 3.10, 3.12 and 3.13 and `jax=0.5.3` on 3.10. Every PR opened while that wheel is missing will look red for the same reason, with no test signal at all.

Local run in the meantime, jax 0.11.1 on macOS arm64:

```
optax/contrib/_muon_test.py  34 passed, 18 subtests passed
  test_newton_schulz sweep    3 passed, 9 subtests passed
```

sylvesterkaczmarek:
Rechecked the rebased head. The reviewed Muon fix is unchanged, and the current red nightly job fails during dependency install on the missing SciPy wheel before pytest runs. My previous no-blocker review still stands.

sylvesterkaczmarek:
Rechecked current d87ecc02. The complex Schatten path still uses the Hermitian Gram matrix fix I reviewed earlier, the regression coverage is intact, and the full current CI matrix including nightly JAX is green. I do not have repository permission to submit a formal approval here, but I have no blocker.

PR #1768, opened 2026-09-03
fix(transform): prevent NaN in scale_by_adamax on float16 zero-gradient steps

somuai:
## Summary

Fixes NaN in `scale_by_adamax` when used with `float16` parameters on zero-gradient steps (such as masked tokens, frozen layers, or unactivated embeddings).

### Problem Context

In `scale_by_adamax`, updates are computed by dividing the first moment by the infinity moment `nu`:
```python
updates = jax.tree.map(lambda m, v: m / v, mu_hat, nu)
```
Where `nu` is tracked by `optax.tree.update_infinity_moment` via `jnp.maximum(jnp.abs(g) + eps, decay * t)`.

When parameters and gradients are in `float16`, JAX casts the default `eps=1e-8` down to `float16`. Because the smallest positive subnormal in IEEE 754 `float16` is `~5.96e-8`, `1e-8` rounds down to `0.0`.

On any step with an all-zero or masked gradient:
- `jnp.abs(g) + eps` evaluates to `0.0 + 0.0 = 0.0`
- `decay * t` is `0.0` on the first step
- `nu` evaluates to `0.0`
- The update division yields `0.0 / 0.0 = NaN`

Once a NaN enters the parameter tree, it propagates across all subsequent steps via `apply_updates`, silently poisoning model training.

### Proposed Changes

1. In `optax/_src/transform.py::scale_by_adamax`, guard the denominator in at least `float32` precision (`safe_v = v.astype(jnp.promote_types(v.dtype, jnp.float32))`, `denom = jnp.maximum(safe_v, eps)`) before downcasting the update back to `m.dtype`.
2. Added regression test `test_adamax_no_nan_on_float16_zero_grad` in `optax/_src/transform_test.py` asserting that multiple zero-gradient steps on `float16` parameters stay finite and maintain the original dtype.
3. Verified bit-for-bit identical outputs on `float32` and `float64` compared to pre-fix behavior.

### Testing

```bash
$ pytest optax/_src/transform_test.py -k "adamax"
2 passed in 0.48s

$ ruff check optax/_src/transform.py optax/_src/transform_test.py
All checks passed!
```

### DIFF
--- a/optax/_src/transform.py
+++ b/optax/_src/transform.py
@@ -420,7 +420,20 @@ def scale_by_adamax(
     nu = optax.tree.update_infinity_moment(updates, state.nu, b2, eps)
     # Bias correction for mean. No bias correction needed for infinity moment.
     mu_hat = optax.tree.bias_correction(mu, b1, count_inc)
-    updates = jax.tree.map(lambda m, v: m / v, mu_hat, nu)
+
+    def _update(m, v):
+      if m is None:
+        return None
+      # `eps` is applied in at least float32 precision so it does not
+      # silently underflow to zero when `v` has a low precision dtype
+      # (e.g. float16, whose smallest subnormal is ~6e-8). Otherwise the
+      # denominator can become exactly 0, turning this into a 0/0 NaN
+      # whenever `m` and `v` are also 0 (e.g. on a zero-gradient step).
+      safe_v = v.astype(jnp.promote_types(v.dtype, jnp.float32))
+      denom = jnp.maximum(safe_v, eps)
+      return (m.astype(denom.dtype) / denom).astype(m.dtype)
+
+    updates = jax.tree.map(_update, mu_hat, nu, is_leaf=lambda x: x is None)
     return updates, ScaleByAdamState(count=count_inc, mu=mu, nu=nu)
 
   return base.GradientTransformation(init_fn, update_fn)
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
`scale_by_adamax` supports complex parameters, but `nu`/`denom` is real here. Casting `m` to `denom.dtype` therefore drops the imaginary component before division and changes complex Adamax updates. Could we keep `m` complex and only promote the denominator/result as needed, with a `complex64` regression?

somuai:
Good catch @sylvesterkaczmarek! Updated to use `m / denom` directly rather than casting `m` to `denom.dtype`.

Because `denom` is a real array evaluated in at least `float32`, `m / denom` divides complex numerators without stripping the imaginary part, while real `float16` inputs still benefit from JAX's promotion to safe `float32` division before being cast back down to `m.dtype`.

Added `test_adamax_complex_parameters` in `optax/_src/transform_test.py` asserting that `complex64` inputs retain non-zero imaginary components and preserve their original dtype. Pushed in f0100a7.

sylvesterkaczmarek (commented):
The Adamax update now divides the complex numerator by the promoted real denominator directly and only casts the result back afterward, so the imaginary component is preserved. The new complex64 regression pins that behavior while retaining the float16 zero-gradient fix. My concern is resolved.

somuai:
Thank you @sylvesterkaczmarek for the review and confirmation! All 15 CI checks are green and the complex64 regression tests are passing. Ready for DeepMind maintainer review and Copybara import when convenient.

somuai:
Gentle check-in on this PR — all 15 CI checks are passing green and all review items have been resolved. Friendly follow-up for @sylvesterkaczmarek whenever there is an opportunity to queue this for Copybara import. Thank you!

sylvesterkaczmarek (commented):
Rechecked current f0100a71. The float16 zero-gradient Adamax fix remains correct, the denominator is promoted without discarding complex numerator components, and the complex64 regression pins that behavior. All current CI checks are green. No remaining blocker from my review.

PR #1772, opened 2026-09-07
Fix negative CTC loss for confident alignments in float32

### LINKED ISSUE #1771: float32 CTC loss can round below zero for confident valid alignments
I had training runs regularly produce nans and then deteriorate late into training.
After a lot more time than I'd like to admit I've found the culprit. I observed mean CTC losses near `-1.9e-5` and `-3.4e-6`. 
In my multi task training setup a downstream adaptive loss balancer computed `log(loss + 1e-8)` and then generated NaNs from otherwise finite training data and gradients.
I didn't check whether this is specific to CTC loss or whether there are other loss functions that can return negative results.

## Reproduction

Observed with Optax 0.2.8 and JAX/jaxlib 0.11.1. No training or mixed precision is required. Run on CPU with x64 disabled:

```python
import jax
import jax.numpy as jnp
import optax

jax.config.update("jax_enable_x64", False)
loss = optax.ctc_loss(
    logits=jnp.array([[[0.0, 17.0], [0.0, 17.0]]], dtype=jnp.float32),
    logit_paddings=jnp.zeros((1, 2)),
    labels=jnp.array([[1]], dtype=jnp.int32),
    label_paddings=jnp.zeros((1, 1)),
    blank_id=0,
)
print(loss)
```

Observed loss: approximately `[-8.279875e-8]`.

There are three valid paths: `(1, 1)`, `(blank, 1)`, `(1, blank)`.
Writing `p` for the blank probability at each frame, their combined probability is `1 - p**2`; 
the true loss `-log1p(-p**2)` is approximately `1.713908e-15`.
The loss should therefore be nonnegative. 
With x64 enabled and float64 logits, the same example yields approximately `1.857496e-15`.

## Workaround

I now clamp each sequence's CTC loss to zero before averaging and guard the downstream balancer separately just in case. 

## Expected Behavior

I think Optax should either enforce CTC's nonnegative mathematical range or document that callers must handle small negative results.

rajasekharporeddy:
Fixes #1771

### Summary
In `float32` arithmetic, `optax.ctc_loss` and `optax.ctc_loss_with_forward_probs` can round below zero ($< 0$) for highly confident alignments due to precision limits in `log_softmax` and `logaddexp`. This PR enforces the mathematical bounds by clamping forward log-probabilities from above at `0.0` and per-sequence loss from below at `0.0`.

### Changes
1. In `optax.losses.ctc_loss_with_forward_probs`:
   - Clamped `logalpha_phi` and `logalpha_emit` from above at `0.0` using `jnp.minimum(..., 0.0)`.
   - Clamped `per_seq_loss` from below at `0.0` using `jnp.maximum(per_seq_loss, 0.0)`.
2. In `optax/losses/_classification_test.py`:
   - Added unit test `test_confident_alignment_nonnegative` in `CTCTest` to verify that both `ctc_loss` and `ctc_loss_with_forward_probs` maintain non-negative loss and non-positive log-probabilities under confident alignments.


rajasekharporeddy:
Hi @rdyro, Please take a look at this PR.

PR #1780, opened 2026-09-11
Fix AdamW example in getting started notebook

rishabhsai:
The "Popular optimizers" cell in docs/getting_started.ipynb builds its AdamW example with `optax.scale_and_decay`, which does not exist. On current main, `python -c "import optax; optax.scale_and_decay"` raises AttributeError, and the notebook only passes because the helper is defined but never called. The cell now chains `scale_by_adam`, `add_decayed_weights(weight_decay)`, and `scale_by_learning_rate(learning_rate)`, the same chain `optax.adamw` uses in optax/_src/alias.py. I checked it by running every code cell of the notebook in a fresh venv, calling the helper's init, and confirming its update on a small pytree matches `optax.adamw` with `jnp.allclose`.


selamw1:
Good catch!, `optax.scale_and_decay `does not exist, and replacing it with `scale_by_adam`, `add_decayed_weights`, and `scale_by_learning_rate` correctly aligns the example with the implementation in `optax.adamw`. 
Thanks for the fix!

selamw1 (commented):
Good catch! optax.`scale_and_decay` does not exist, and replacing it with `scale_by_adam`, `add_decayed_weights`, and `scale_by_learning_rate` correctly aligns the example with the implementation in `optax.adamw`.
Thanks for the fix! LGTM.

rishabhsai:
@rdyro selamw1 has already checked this one. Could you take a look when you get a chance? It replaces `optax.scale_and_decay`, which doesn't exist, with the chain `optax.adamw` actually uses. #1781 is a small docstring fix for the onecycle and warmup cosine schedules.

PR #1788, opened 2026-09-18
Fix sam with sync_period=1 undoing the previous step in transparent mode

Nicholas022400701:
### Summary

`optax.contrib.sam` accepts `sync_period=1`, which means the outer optimizer runs on every step and no adversarial step is taken. In transparent mode this configuration does not reduce to the base optimizer. Every step after the first undoes the previous step, so training stalls one step away from the initial parameters.

### Details

`transparent_update_fn` computes the outer update as

```python
opt_updates, opt_state = optimizer.update(updates, state.opt_state, state.cache)
opt_updates = jax.tree.map(lambda c, p, u: c - p + u, state.cache, params, opt_updates)
...
cache = pick_one(first_step, params, state.cache)
```

`state.cache` holds the parameters saved at the first step of the previous period. When `sync_period` is 1 every step is both the first and the last step of its period, so at step `t` the outer update is computed from the parameters of step `t - 1`. With SGD as the outer optimizer the parameters become `x_t = x_0 + u_{t-1}`. The first step is correct and every later step throws the previous one away.

The fix passes the current parameters to the outer optimizer when `sync_period == 1`, so the wrapper reduces to the base optimizer, which is also what `opaque_mode=True` already does for this configuration. The `sync_period >= 2` path is unchanged.

### Reproduction

SGD with `learning_rate=0.1` on the parabola from `_sam_test.py`, starting at `[-1, 10, 1]` and targeting `[1, -1, 1]`, compared after two steps with the same SGD used directly:

```
ACTUAL:  array([-0.68,  8.24,  1.  ], dtype=float32)   sam(sgd, ..., sync_period=1)
DESIRED: array([-0.28,  6.04,  1.  ], dtype=float32)   sgd
```

The actual value equals `x_0 - 0.2 * 0.8 * (x_0 - x_star)`, that is the second update applied to the initial parameters instead of to the first-step parameters.

### Changes

- `_sam.py`: when `sync_period == 1` use the current parameters instead of `state.cache` for the outer optimizer call and the `c - p + u` correction. Nothing else changes.
- `_sam_test.py`: add `test_sync_period_one_matches_base_optimizer`, which runs five steps of `sam(sgd, sync_period=1)` in both modes and checks each step against plain SGD. On main the transparent case fails at the second step with the numbers above and the opaque case passes.
- `_sam_test.py`: add `test_transparent_mode_matches_opaque_mode`, which runs `adamw` as the outer optimizer and `adam` as the adversarial optimizer for `sync_period` 1 to 3 with and without `reset_state` and checks the transparent mode against the opaque mode after every period. On main the two `sync_period=1` cases fail and the `sync_period` 2 and 3 cases pass. With the fix all 12 tests in the file pass.

Run on CPU with Python 3.13.14, jax 0.11.2 and chex 0.1.92. `pylint`, `ruff` and `flake8` are clean on both files.

AI disclosure: I used an AI coding agent to help write this patch, the tests and this description. I have read the change and the tests myself and I will answer review comments personally.


sylvesterkaczmarek (commented):
Moving cache before optimizer.update() also changes sync_period > 1 on every first_step: the outer optimizer now receives current params instead of the previous cache, and its returned opt_state is kept even though that step's opt_updates are discarded. A param-dependent/stateful outer transform can therefore change outside the sync_period=1 case. Could this be scoped to period 1 or covered with a param-dependent outer optimizer?

Nicholas022400701:
Done in 04e08be and 5fd9276. The outer optimizer now gets the current parameters only when `sync_period == 1` and the `sync_period > 1` path is unchanged from main.

For the record, on a first step with `sync_period > 1` the existing `pick_one(last_step, ...)` lines already dropped both the `opt_updates` and the `opt_state` of that call, and with the previous revision adamw trajectories for `sync_period` 2 and 3 were bit identical to main. The scoped version does not need that argument though.

I also added `test_transparent_mode_matches_opaque_mode`. It runs `adamw` as the outer optimizer and `adam` as the adversarial optimizer for `sync_period` 1 to 3 with and without `reset_state` and checks the transparent mode against the opaque mode after every period. On main the two `sync_period=1` cases fail and the other four pass.

sylvesterkaczmarek:
Rechecked 5fd9276b. The behavior change is now scoped to sync_period == 1, so the sync_period > 1 transparent path continues to feed the outer optimizer the cached parameters exactly as before. The new adamw/adam regression also exercises a parameter-dependent, stateful outer optimizer across periods 1-3 and both reset-state modes. My concern is resolved.

Issue #1784, opened 2026-09-15
RMS transforms produce NaN on float16 zero-gradient steps


vishesh9131:
### Description

`scale_by_rms` and `scale_by_stddev` produce `NaN` updates on the first zero-gradient step when the updates and state use `float16`. This also affects `optax.rmsprop`.

The default `eps=1e-8` is evaluated in float16 during the denominator calculation and underflows to zero. The result is an infinite scale multiplied by a zero gradient.

### Reproduction

```python
import jax.numpy as jnp
import optax

for factory in (optax.scale_by_rms, optax.scale_by_stddev):
  tx = factory()
  grads = jnp.zeros((4,), dtype=jnp.float16)
  updates, _ = tx.update(grads, tx.init(grads))
  print(factory.__name__, updates)
```

Current output:

```
scale_by_rms [nan nan nan nan]
scale_by_stddev [nan nan nan nan]
```

The issue occurs with both values of `eps_in_sqrt` and `bias_correction`. It also occurs when an ArrayLike epsilon has already underflowed, for example `jnp.asarray(1e-8, dtype=jnp.float16)`.

### Expected behavior

A zero gradient should produce a finite, zero update while preserving the float16 update dtype.

### Proposed fix

Compute the variance and denominator in at least float32, floor a zero denominator to the smallest positive normal value of that computation dtype, and cast the final update back to the gradient dtype. Add JIT-compiled regression coverage for both RMS transforms, both epsilon placements, bias correction on/off, and scalar/ArrayLike epsilon values.

PR #1790, opened 2026-09-18
Fix float16 zero-gradient NaNs in AdaDelta, Adan, and NovoGrad

### LINKED ISSUE #1789: AdaDelta, Adan, and NovoGrad produce NaN on float16 zero-gradient steps
## Bug

`optax.scale_by_adadelta`, `optax.scale_by_adan`, and `optax.scale_by_novograd` still compute their RMS denominator in the update dtype. With default `eps` (`1e-6` / `1e-8`) that underflows in float16, so a zero-gradient step is `0/0 = NaN` and poisons the rest of training.

Adam-family (`#1754`) and RMS (`#1784`) already have open fixes. These three remaining adaptive scalers have the same hole.

## Reproduction

```python
import jax.numpy as jnp
import optax

params = jnp.zeros(4, dtype=jnp.float16)
grads = jnp.zeros(4, dtype=jnp.float16)
for factory in (optax.scale_by_adadelta, optax.scale_by_adan, optax.scale_by_novograd):
    tx = factory()
    updates, _ = tx.update(grads, tx.init(params), params)
    print(factory.__name__, updates)
```

Expected: finite zeros. Actual: NaNs.

I have a fix with tests and will send a PR.

Anu27n:
## Summary

Adam-family (`#1754`) and RMS (`#1784`) already have open float16 0/0 fixes. **AdaDelta, Adan, and NovoGrad still add `eps` in the update dtype**, so default `eps` underflows in float16 and a zero gradient becomes `NaN`.

This computes the denominator in at least float32 and floors a zero denom, then casts the update back.

Fixes #1789

## Test plan

- [x] `optax/_src/transform_test.py::test_zero_gradient_float16_is_finite` for AdaDelta, Adan, and NovoGrad
- [ ] CI transform tests


Anu27n:
Thanks for reproducing this — your gist matches what I see.

AdaDelta’s default `eps=1e-6` is still representable in float16 (as a subnormal), so `sqrt(0 + eps)` stays nonzero and a zero gradient stays `0`. Adan and NovoGrad default to `eps=1e-8`, which underflows to `0` in float16, so the RMS denom becomes `0/0` and the update is `NaN`. That’s why only those two blow up at the library defaults.

The AdaDelta change in the PR is the same float32 denom used for Adam/RMS: it keeps AdaDelta finite if a caller passes a smaller `eps` (for example `1e-8`) in float16. Happy to drop AdaDelta from this PR if you’d rather keep the scope to the two default-eps NaNs.

